PR #2388, opened 2015-05-26
[v6r13] dirac-dms-get-file broken

atsareg:
FIX: getFile() allows to pass a list of LFNs

### DIFF
--- a/Interfaces/API/Dirac.py
+++ b/Interfaces/API/Dirac.py
@@ -83,7 +83,7 @@ class Dirac( API ):
   def __checkFileArgument( self, fnList, prefix = None, single = False ):
     if prefix is None:
       prefix = 'LFN'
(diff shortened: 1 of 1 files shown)


chaen:
Review OK

phicharp:
I have been checking that script and my test was successful (with a single file)...  I was trying to replicate the former behavior and probably at the time I thought it was not allowed to use the API getFile() with a list of LFNs. It is true however that now the script and the API were incompatible (as probably I also added this possibility in the script or removed a loop that was therein. 
Apologies for this inconsistency that you simple change fixes indeed!

phicharp:
Andrei: probably a mistake but the PR is for Integration, not v6r13 ;-)

atsareg:
Wrong target branch

PR #2390, opened 2015-05-27
[v6r13] Fixes to DMSHelpers

phicharp:
The module Core.Utilities.SiteSEMapping was updated to make use of the DMSHelpers module. This should ease the porting of all components to using this single source of information.
DataManager was updated to use directly DMSHelpers

### DIFF
--- a/Resources/Storage/StorageFactory.py
+++ b/Resources/Storage/StorageFactory.py
@@ -264,7 +264,7 @@ class StorageFactory:
     # This is a temporary for backward compatibility
     if "ProtocolName" in protocolDict and not protocolDict['PluginName']:
       protocolDict['PluginName'] = protocolDict['ProtocolName']
-    protocolDict.pop( 'ProtocolName' )
+    protocolDict.pop( 'ProtocolName', None )
 
     # Evaluate the base path taking into account possible VO specific setting
     if self.vo:
--- a/Resources/Storage/StorageFactory.py
+++ b/Resources/Storage/StorageFactory.py
@@ -35,6 +35,12 @@ class StorageFactory:
         self.vo = result['Value']
       else:
         RuntimeError( "Can not get the current VO context" )
+    self.remotePlugins = []
+    self.localPlugins = []
+    self.name = ''
+    self.options = {}
+    self.protocolDetails = []
+    self.storages = []
 
   ###########################################################################################
   #
@@ -60,7 +66,7 @@ class StorageFactory:
       pluginName = parameterDict['PluginName']
     # Temporary fix for backward compatibility
     elif parameterDict.has_key( 'ProtocolName' ):
-      pluginName = parameterDict['ProtocolName']  
+      pluginName = parameterDict['ProtocolName']
     else:
       errStr = "StorageFactory.getStorage: PluginName must be supplied"
       gLogger.error( errStr )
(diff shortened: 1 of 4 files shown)


Resources/Storage/StorageFactory.py
```diff
@@ -35,6 +35,12 @@ def __init__( self, useProxy = False, vo = None ):
         self.vo = result['Value']
       else:
         RuntimeError( "Can not get the current VO context" )
+    self.remotePlugins = []
```
chaen: These are already initialized in getStorages. Can it be initialized in one place only ? The __init__ is probably the best one

Resources/Storage/StorageFactory.py
```diff
@@ -93,7 +101,6 @@ def getStorages( self, storageName, pluginList = [] ):
     # In case the storage is made from a base SE, get this information
     res = self._getConfigStorageName( storageName, 'BaseSE' )
     if not res['OK']:
-      self.valid = False
```
chaen: why is that removed ? Is this attribute even used anywhere ?

Resources/Storage/StorageFactory.py
```diff
@@ -253,13 +260,13 @@ def _getConfigStorageProtocolDetails( self, storageName, protocolSection ):
...
-        
+
     # This is a temporary for backward compatibility
-    if "ProtocolName" in protocolDict and not protocolDict['PluginName']:  
+    if "ProtocolName" in protocolDict and not protocolDict['PluginName']:
       protocolDict['PluginName'] = protocolDict['ProtocolName']
-    protocolDict.pop( 'ProtocolName' )  
-        
-    # Evaluate the base path taking into account possible VO specific setting 
+    protocolDict.pop( 'ProtocolName' )
```
chaen: Either this has to go inside the "if" statement, or in a try catch, or pop('ProtocolName', None)

Resources/Storage/StorageFactory.py
phicharp: No, this is why I removed it ;-)

Resources/Storage/StorageFactory.py
phicharp: Yes I know... I had no idea whether it was required to re-initialize them in getStorages. As it is it doesn't harm and makes the number of pylint warnings decrease... Someone who has time/expertise should check whether they can be removed from getStorages ;-)

Resources/Storage/StorageFactory.py
phicharp: There is no change here on my side... Just Eclipse beautifying (although I have no clue why the comment disappeared ;-). I will check the logic in here...

phicharp: In my Eclipse the comment is not removed, just a few spaces removed in the 2 preceding lines... strange :(

phicharp: B.t.w. it would be good if everybody could avoid adding blank spaces here and there... overtime Eclipse is removing them and creates an artificial change which makes reviewing difficult... at least one good point with Eclipse ;-)

phicharp: Oups, sorry... just checked... Indeed the comment was there twice and I removed it once, so all is OK

Resources/Storage/StorageFactory.py
chaen: From a quick look, it looks like they don't need to be reinitialized.

phicharp: OK, I'll remove them after a second check...

phicharp: Removed

chaen:
review OK

PR #2391, opened 2015-05-29
Allow enabling job repository by cmd line switch

sposs:
The default is False, i.e. the same a before.

Follows discussion...


sposs:
Too fast Andrei, where is the review?

atsareg:
review OK

PR #2392, opened 2015-05-29
Allow setting the repo file name

sposs:
Use the same switch but use the value as file name. Means it cannot use the default file name ($HOME/.dirac.repo.cfg, will create another fix for that).

### DIFF
--- a/Interfaces/scripts/dirac-wms-job-submit.py
+++ b/Interfaces/scripts/dirac-wms-job-submit.py
@@ -30,10 +30,13 @@ if len( args ) < 1:
 from DIRAC.Interfaces.API.Dirac                              import Dirac
 unprocessed_switches = Script.getUnprocessedSwitches()
 use_repo = False
+repo_name = ""
 for sw, value in unprocessed_switches:
   if sw.lower() in ["r", "usejobrepo"]:
     use_repo = True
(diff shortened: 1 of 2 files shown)


sposs:
Don't merge yet, I need to do something else.

sposs:
Ok, now it's OK

Interfaces/scripts/dirac-wms-job-submit.py
```diff
@@ -30,10 +30,13 @@
 from DIRAC.Interfaces.API.Dirac                              import Dirac
 unprocessed_switches = Script.getUnprocessedSwitches()
 use_repo = False
+repo_name = ""
 for sw, value in unprocessed_switches:
   if sw.lower() in ["r", "usejobrepo"]:
     use_repo = True
-dirac = Dirac(use_repo)
+    repo_name = value
+    repo_name = repo_name.replace(".cfg", ".repo")
```
chaen: ".repo" ? the other change you did are with ".rep". Is it normal ?

sposs: It doesn't really matter. The only relevant thing was to change from .cfg to anything else. This particular line above is just to make sure no one uses .cfg as extension.

PR #2394, opened 2015-05-29
[v6r11] Add first implementation for GlobusComputingElement

andresailer:
Implemented and tested functions:
jobSubmit
getJobOutput

Implemented functions:
getCEStatus (empty implementation)
killJob (not sure where this is called from)

### DIFF
--- a/Resources/Computing/GlobusComputingElement.py
+++ b/Resources/Computing/GlobusComputingElement.py
@@ -49,12 +49,6 @@ class GlobusComputingElement( ComputingElement ):
     self.proxyRenewal = 0
 
   #############################################################################
-  def _addCEConfigDefaults( self ):
-    """Method to make sure all necessary Configuration Parameters are defined
-    """
-    # First assure that any global parameters are loaded
-    ComputingElement._addCEConfigDefaults( self )
-
   def __writeRSL( self, executableFile ):
     """ Create the RSL for submission
 
--- a/Resources/Computing/GlobusComputingElement.py
+++ b/Resources/Computing/GlobusComputingElement.py
@@ -210,9 +210,8 @@ class GlobusComputingElement( ComputingElement ):
 
     return 'Unknown'
 
-  def getJobOutput( self, jobID, localDir = None ):
-    """ Get the specified job standard output and error files. If the localDir is provided,
-        the output is returned as file in this directory. Otherwise, the output is returned 
+  def getJobOutput( self, jobID, _localDir = None ):
+    """ Get the specified job standard output and error files. The output is returned
         as strings. 
     """
 
--- a/Resources/Computing/GlobusComputingElement.py
+++ b/Resources/Computing/GlobusComputingElement.py
@@ -1,0 +1,283 @@
+########################################################################
+# File :   GlobusComputingElement.py
+# Author : A.S.
+########################################################################
+
+""" Globus Computing Element
+
+   Allows direct submission to Globus Computing Elements with a SiteDirector Agent
+
+"""
+
+__RCSID__ = "$Id$"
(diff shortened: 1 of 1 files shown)


Resources/Computing/GlobusComputingElement.py
```diff
@@ -0,0 +1,283 @@
...
+from DIRAC.Resources.Computing.ComputingElement          import ComputingElement
+from DIRAC.Core.Utilities.Grid                           import executeGridCommand
+from DIRAC                                               import S_OK, S_ERROR
+
+from DIRAC.ConfigurationSystem.Client.Helpers.Registry   import getGroupOption
+from DIRAC.FrameworkSystem.Client.ProxyManagerClient     import gProxyManager
+from DIRAC.WorkloadManagementSystem.DB.PilotAgentsDB     import PilotAgentsDB
+
+
+import os, tempfile
```
chaen: The PEP8 specifications that we try to follow more and more recommend to do the import on different lines
https://www.python.org/dev/peps/pep-0008/#imports

Resources/Computing/GlobusComputingElement.py
```diff
@@ -0,0 +1,283 @@
...
+
+class GlobusComputingElement( ComputingElement ):
+  """Globus computing element class
+  implementing the functions jobSubmit, getJobOutput """
+
+  #############################################################################
+  def __init__( self, ceUniqueID ):
+    """ Standard constructor.
+    """
+    ComputingElement.__init__( self, ceUniqueID )
```
chaen: You should rather use the 'super' method

Resources/Computing/GlobusComputingElement.py
```diff
@@ -0,0 +1,283 @@
...
+    self.submittedJobs = 0
+    self.mandatoryParameters = MANDATORY_PARAMETERS
+    self.pilotProxy = ''
+    self.queue = ''
+    self.outputURL = 'gsiftp://localhost'
+    self.gridEnv = ''
+    self.proxyRenewal = 0
+
+  #############################################################################
+  def _addCEConfigDefaults( self ):
```
chaen: if you are not doing anything more that the parent class, you don't need to re-implement it

Resources/Computing/GlobusComputingElement.py
andresailer: In this case pylint complains about "using super in an old-style class", because "ComputingElement" does not inherit from "object".
Is it safe to use super here then?

andresailer: To answer my own question: It is not possible to use super in this case

chaen: why not ? It should not be a problem unless we have multiple inheritance

andresailer: Because old and new style classes are different, and old-style classes don't have super
```
class ClassN(object):
    def __init__(self):
        self.var = "N1"

class ClassNI(ClassN):
    def __init__(self):
        super(ClassNI, self).__init__()

class ClassO:
    def __init__(self):
        self.var = "N1"

class ClassOI(ClassO):
    def __init__(self):
        super(ClassOI, self).__init__()

if __name__ == "__main__":
    n = ClassNI()
    o = ClassOI()
```
Leads to
```
  File "OldvsNew.py", line 19, in <module>
    o = ClassOI()
  File "OldvsNew.py", line 15, in __init__
    super(ClassOI, self).__init__()
TypeError: must be type, not classobj
```
See for example: http://stackoverflow.com/questions/9698614/super-raises-typeerror-must-be-type-not-classobj-for-new-style-class

PR #2395, opened 2015-06-02
[v6r13] fix get current url

chaen:
This is preventing staging from FTS. We should get rid of this method anyway.

### DIFF
--- a/Resources/Storage/StorageBase.py
+++ b/Resources/Storage/StorageBase.py
@@ -229,7 +229,7 @@ class StorageBase( object ):
     if not result['OK']:
       return result
     cwdUrl = result['Value']
-    fullUrl = '%s/%s' % ( cwdUrl, fileName )
+    fullUrl = '%s%s' % ( cwdUrl, fileName )
     return S_OK( fullUrl )
     
   def getName( self ):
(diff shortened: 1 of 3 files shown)


phicharp:
Review OK

phicharp:
Review OK for the new changes. This PR is extremely urgent

Resources/Storage/StorageBase.py
```diff
@@ -229,7 +229,7 @@ def getCurrentURL( self, fileName ):
     if not result['OK']:
       return result
     cwdUrl = result['Value']
-    fullUrl = '%s/%s' % ( cwdUrl, fileName )
+    fullUrl = '%s%s' % ( cwdUrl, fileName )
```
atsareg: I think this fix is not correct. I guess this is done to avoid double slash in the fullUrl since in the only place it is used the lfn ( starting with '/' ) is passed.  In the FTSClient, where this is used, we should rather use constructURLFromLFN( lfn ) method of a storage element. getCurrentURL() was meant rather to be used to "navigate" the SE namespace.

chaen: Yes, totally agree, and this is the plan. However:
* changing the logic in FTS is much heavier
* the fact remains: getCurrentURL is wrong as such, there should not be any double slash. In fact, I believe we should remove this, since it is not used anywhere. But not now, now we need to release :-)

atsareg:
ok

PR #2396, opened 2015-06-03
[v6r13] Fix to DownloadInputData

phicharp:
FIX: could not find cached files (missing [lfn])
CHANGE: remove caching of StorageElements as it is already a cache!
FIX: localFile was not defined properly

It was impossible to download a file if not from a local SE.
In addition if there were no local SEs we were instantiating a weird SE called None
Changed method names as they were totally misleading (PFN don't exist any longer)

### DIFF
--- a/WorkloadManagementSystem/Client/DownloadInputData.py
+++ b/WorkloadManagementSystem/Client/DownloadInputData.py
@@ -20,6 +20,10 @@ import os, tempfile, random
 
 COMPONENT_NAME = 'DownloadInputData'
 
+def _isCached( lfn, seName ):
+  result = StorageElement( seName ).getFileMetadata( lfn )
+  return result['OK'] and result['Value']['Successful'].get( lfn, {} ).get( 'Cached', False )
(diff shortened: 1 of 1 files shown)


phicharp:
We need this PR to be merged for RAW data mesh processing

WorkloadManagementSystem/Client/DownloadInputData.py
```diff
@@ -72,8 +73,8 @@ def execute( self, dataToResolve = None ):
     # determine Disk and Tape SEs
     diskSEs = set()
     tapeSEs = set()
-    for localSE in localSEList:
-      seStatus = self.__storageElement( localSE ).getStatus()['Value']
+    for localSE in [se for se in localSEList if se]:
+      seStatus = StorageElement( localSE ).getStatus()['Value']
```
atsareg: This is a good occasion to start checking the output result status before asking for Value

atsareg:
Review OK

PR #2397, opened 2015-06-04
!!![v6r13p1] new voms policy

chaen:
Andrei, please don't merge this one yet. I would like you to merge it after we finally released v6r13, with a pre-tag, so that we can run the full certification procedure again, and then put it in a patch version. This is quite a big change.

This adds a much finer grain granularity to permission management.
* We don't only ask "can I write", but "can I add replica", "can I delete file", etc
* A new SecurityManager is added: PolicyBasedSecurityManager. This allows to define in the CS a python file which will be called to evaluate the permission. The CS attribute is SecurityPolicy
* One of such a policy is already defined: VOMSPolicy. This implements some sort of group of group, so that a file can belong to several group at the time. The groups are linked by their VOMS role. Also, this implements a 3 level posix system, except that it is not traversal. 
* The FileCatalog exposes the method 'hasAccess'
* The DataManager now uses this hasAccess method rather than evaluating the write permission itself.
* Obviously, the LFC plugin and the other Security managers have been adapted to it.
* Also, when no permission was granted, the call to the Managers were still attempted,  which resulted in S_ERROR("empty input"), not very helpful. I fixed that for all calls.


atsareg:
This is a new security plugin. If it is not used, nobody notices or suffers - is it correct ? If it is, it is safe to add it.

atsareg:
Well, actually it has repercussions everywhere. v6r14 ?

chaen:
Nop, too far. 
We discussed it in the BILD meeting where you were there, and we agreed to put it in a patch. 
But I understand that it is delicate, this is why I ask you to do anotehr pre release after the real release, so we can turn all the certification again, make sure that it does not change the existing. At least, all the unit and integration tests I ran were behaving properly

phicharp:
Numbers being only numbers ;-), if we want to stick to the definition that a patch release only contains minor changes / bug fixes, I think we should be flexible enough to be able to change the numbering and what was assumed to be called v6r14 me moved to v6r15 and this release (containing only this and possibly a few patches of v6r13) be called v6r14.
Fixing long ahead what is going to go into the next minor version should not prevent us to apply urgent large changes that require certification. 
With this scheme we can go on patching v6r13 while v6r14 (the new one) is certified.
Does this make sense?

chaen:
Good idea. Make it immediately v6r14, so v6r13 will have a lifetime of 1 week :-)

atsareg:
Well, I actually mean straight after the v6r13 is released. If this will be v6r14, even better. Otherwise an early v6r13 patch relying on certification tests

andresailer:
It would be great to not have commit messages like "fix" , "tmp commit" or similar.
One can use git rebase --interactive <commitSha>^ to merge (or split) commits, and git commit --amend where appropriate
And instead of merging the release branch into the topic branch one can rebase the topic branch.

atsareg:
Is this change backward compatible for those installations that do not use the new SecurityManager ?

atsareg:
Merged into the integration branch

chaen:
It is meant to be at least. 
Out of curiosity, how did you merge it into integration, since the PR was against v6r13 ?

atsareg:
I have merged it by hand in my local git repository from your remote branch

PR #2404, opened 2015-06-05
[v6r13] Fix one use case in DMSHelpers

phicharp:
After a change to be able to define an SEGroup as accessible to a site, it was no longer possible to define it as a site (which is the main use case)
Sites should be excluded and then added

### DIFF
--- a/DataManagementSystem/Client/DataManager.py
+++ b/DataManagementSystem/Client/DataManager.py
@@ -1304,7 +1304,7 @@ class DataManager( object ):
       replicaTuples = [replicaTuple]
     for replicaTuple in replicaTuples:
       if not isinstance( replicaTuple, tuple ):
-        errStr = "registerFile: Supplied file info must be tuple of list of tuples."
+        errStr = "registerFile: Supplied file info must be tuple or list of tuples."
         self.log.debug( errStr )
         return S_ERROR( errStr )
     if not replicaTuples:
--- a/DataManagementSystem/Client/DataManager.py
+++ b/DataManagementSystem/Client/DataManager.py
@@ -909,7 +909,7 @@ class DataManager( object ):
     else:
       fileTuples = [fileTuple]
     for fileTuple in fileTuples:
-      if not isinstance( fileTyple, tuple ):
+      if not isinstance( fileTuple, tuple ):
         errStr = "registerFile: Supplied file info must be tuple of list of tuples."
         self.log.debug( errStr )
         return S_ERROR( errStr )
@@ -955,7 +955,7 @@ class DataManager( object ):
     else:
       replicaTuples = [replicaTuple]
     for replicaTuple in replicaTuples:
-      if not isinstance( fileTyple, tuple ):
+      if not isinstance( replicaTuple, tuple ):
         errStr = "registerFile: Supplied file info must be tuple of list of tuples."
         self.log.debug( errStr )
         return S_ERROR( errStr )
@@ -1303,7 +1303,7 @@ class DataManager( object ):
     else:
       replicaTuples = [replicaTuple]
     for replicaTuple in replicaTuples:
-      if not isinstance( fileTyple, tuple ):
+      if not isinstance( replicaTuple, tuple ):
         errStr = "registerFile: Supplied file info must be tuple of list of tuples."
         self.log.debug( errStr )
         return S_ERROR( errStr )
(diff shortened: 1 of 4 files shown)


phicharp:
Added to the same PR, just for convenience:

* Return a success dictionary when no LFNs are passed to the DataManager:
This was causing an error to be returned in some cases (e.g. removeFile as fc.exists() returns an error. It should also be fixed, to be decided if in the interface or in the implementation

* Reset the attempts counter when resetting a request Waiting as now the error string no longer is "Max attempts reached"

DataManagementSystem/Client/DataManager.py
```diff
@@ -947,14 +950,17 @@ def registerReplica( self, replicaTuple, catalog = '' ):
...
+    if isinstance( replicaTuple, list ) or isinstance( replicaTuple, dict ) or isinstance( replicaTuple, set ):
+      replicaTuples = list( replicaTuple )
     else:
-      errStr = "registerReplica: Supplied file info must be tuple of list of tuples."
-      self.log.debug( errStr )
-      return S_ERROR( errStr )
+      replicaTuples = [replicaTuple]
+    for replicaTuple in replicaTuples:
+      if not isinstance( replicaTuple, tuple ):
+        errStr = "registerFile: Supplied file info must be tuple of list of tuples."
```
andresailer: I think this should be "tuple OR list of tuples"

DataManagementSystem/Client/DataManager.py
```diff
@@ -947,14 +950,17 @@ def registerReplica( self, replicaTuple, catalog = '' ):
 
         'replicaTuple' is a tuple or list of tuples of the form (lfn,pfn,se)
     """
-    if isinstance( replicaTuple, list ):
-      replicaTuples = replicaTuple
-    elif isinstance( replicaTuple, tuple ):
-      replicaTuples = [ replicaTuple ]
+    if isinstance( replicaTuple, list ) or isinstance( replicaTuple, dict ) or isinstance( replicaTuple, set ):
```
andresailer: you can do isinstance( fileTuple, (list, dict, set) )
By the way, what about tuple, that is what if fileTuple is a tuple of tuple(s))

DataManagementSystem/Client/DataManager.py
phicharp: Sorry, typo

DataManagementSystem/Client/DataManager.py
phicharp: Right, will fix it

phicharp:
Latest commit corresponds to remark of Andre in 
https://github.com/DIRACGrid/DIRAC/issues/2405

atsareg:
Review OK

DataManagementSystem/Client/DataManager.py
```diff
@@ -904,14 +904,17 @@ def registerFile( self, fileTuple, catalog = '' ):
...
-      self.log.debug( errStr )
-      return S_ERROR( errStr )
+      fileTuples = [fileTuple]
+    for fileTuple in fileTuples:
+      if not isinstance( fileTuple, tuple ):
+        errStr = "registerFile: Supplied file info must be tuple or list of tuples."
+        self.log.debug( errStr )
+        return S_ERROR( errStr )
+    if not fileTuples:
+      return S_OK( {'Successful':[], 'Failed':{}} )
```
chaen: Shouldn't successful also be a dictionary ?

phicharp: Well, this depends... Usually the test done after that is like:

if lfn in result['Value']['Successful']

of course if it is 

for lfn, se in result['Value']['Successful'].items():

there may be a problem... Probably {} is safer in this respect, right as most (all?) constructs possible when expecting a list would work with a dict?

atsareg: Yes, it is mandatory that in this
      structure both Successful and Failed are dicts.
      As a matter of definition.
      
        Cheers,
        Andrei
      
      On 09/06/15 12:58, Philippe Charpentier wrote:
    
    
      In DataManagementSystem/Client/DataManager.py:
      >      else:
> -      errStr = "registerFile: Supplied file info must be tuple of list of tuples."
> -      self.log.debug( errStr )
> -      return S_ERROR( errStr )
> +      fileTuples = [fileTuple]
> +    for fileTuple in fileTuples:
> +      if not isinstance( fileTuple, tuple ):
> +        errStr = "registerFile: Supplied file info must be tuple or list of tuples."
> +        self.log.debug( errStr )
> +        return S_ERROR( errStr )
> +    if not fileTuples:
> +      return S_OK( {'Successful':[], 'Failed':{}} )

      Well, this depends... Usually the test done after that is like:
      if lfn in result['Value']['Successful']
      of course if it is 
      for lfn, se in result['Value']['Successful'].items():
      there may be a problem... Probably {} is safer in this respect,
        right as most (all?) constructs possible when expecting a list
        would work with a dict?
      —
          it on GitHub.

PR #2412, opened 2015-06-09
[v6r13] fix reset/cancel request and SRM2 stage request

chaen:
* Adapt the reset and cancel RMS script to also accept request name, providing it is unique
* The error code for SRM_REQUEST_QUEUED changed in a more recent version of srm-ifce, so now we handle both case
* There were some error in the difficult merging when converting the RMS to SQLAlchemy and json, mostly in tests. Corrected that


phicharp:
Review OK

PR #2418, opened 2015-06-10
[v6r13] DMS fixes 

phicharp:
This PR fixes he following:

* dirac-dms-get-file: broken as local site usually has no SE (this is an important fix)
* dirac-dms-user-lfns: sort result

### DIFF
--- a/DataManagementSystem/Utilities/DMSHelpers.py
+++ b/DataManagementSystem/Utilities/DMSHelpers.py
@@ -156,7 +156,7 @@ class DMSHelpers( object ):
   def isSEFailover( self, storageElement ):
     seList = resolveSEGroup( self.__opsHelper.getValue( 'DataManagement/SEsUsedForFailover', [] ) )
     # FIXME: remove string test at some point
-    return storageElement in resolveSEGroup( seList ) or ( isinstance( storageElement, basestring ) and 'FAILOVER' in storageElement.upper() )
+    return storageElement in resolveSEGroup( seList ) or ( not seList and isinstance( storageElement, basestring ) and 'FAILOVER' in storageElement.upper() )
 
   def isSEForJobs( self, storageElement, checkSE = True ):
     if checkSE:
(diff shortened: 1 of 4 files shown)


chaen:
review OK

phicharp:
Added 2 fixes, please re-review ;-)

RequestManagementSystem/Client/Request.py
```diff
@@ -85,6 +86,8 @@ def __init__( self, fromDict = None ):
     self.RequestName = None
     self.OwnerGroup = None
     self.SourceComponent = None
+    
+    self.dmsHelper = DMSHelpers())
```
andresailer: To many closing parenthesis

DataManagementSystem/Utilities/DMSHelpers.py
```diff
@@ -155,7 +155,8 @@ def getStorageElements( self ):
 
   def isSEFailover( self, storageElement ):
     seList = resolveSEGroup( self.__opsHelper.getValue( 'DataManagement/SEsUsedForFailover', [] ) )
-    return storageElement in resolveSEGroup( seList )
+    # FIXME: remove string test at some point
+    return storageElement in resolveSEGroup( seList ) or ( not seList and isinstance( storageElement, basestring ) and 'FAILOVER' in storageElement.upper() )
```
chaen: This means that if we define the SEsUsedForFailover attribute, they all MUST be there, otherwise even if called something FAILOVER, it won't pass the test. Correct ?

RequestManagementSystem/Client/Request.py
chaen: Ugly coupling, but we have no choice :-( 
But I repeat again: soon (very soon), this optimization based on failover will have to go to the failoverModule, and not here. And then the coupling will be gone :-)

chaen: Well spotted Andre !! :-)

atsareg:
Please, conclude on the review

RequestManagementSystem/Client/Request.py
```diff
@@ -469,7 +472,7 @@ def optimize( self ):
           fileSetB = set( list( f.LFN for f in op2 ) )
           if fileSetA == fileSetB:
             # Source is useless if failover
-            if 'FAILOVER' in op1.SourceSE:
+            if self.dmsHelper.isSEFailover( op1.sourceSE ):
```
chaen: Damn !!! SourceSE, not sourceSE.... this one went unnoticed.

PR #2419, opened 2015-06-11
[v6r13] FIX exception in the RMS with some RegisterFile Operation

chaen:
Some argument could be None, so split wouldn't work

### DIFF
--- a/RequestManagementSystem/Client/Operation.py
+++ b/RequestManagementSystem/Client/Operation.py
@@ -201,12 +201,12 @@ class Operation( object ):
   @property
   def sourceSEList( self ):
     """ helper property returning source SEs as a list"""
-    return self.SourceSE.split( "," )
+    return self.SourceSE.split( "," ) if self.SourceSE else ['']
 
   @property
   def targetSEList( self ):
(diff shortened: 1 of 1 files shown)


phicharp:
Review OK

RequestManagementSystem/Client/Operation.py
```diff
@@ -201,12 +201,12 @@ def __len__( self ):
   @property
   def sourceSEList( self ):
     """ helper property returning source SEs as a list"""
-    return self.SourceSE.split( "," )
+    return self.SourceSE.split( "," ) if self.SourceSE else ['']
```
andresailer: Are you intentionally putting an empty string in the list?
[''] does not evaluate to false.

chaen: I know, this was the old behavior. There are places where we do Operation.targetSEList[0], which certainly not the right thing to do, but this requires much more digging that can't be done now. However this fix is necessary for the RMS to execute some operation, it is hotfixed

andresailer: Ok, I understand.

PR #2424, opened 2015-06-12
[v6r13] DMS fixes 

phicharp:
* Added printout in FTSAgent and FTSJob to allow a better debugging of failing transfers
* Temporary fix for allowing getTransportURL even when the SE is banned for read. This requires a check in SRM2 plugin before calling the service


chaen:
Review OK, but we shouldn't forget to discuss this changes a bit more in depth :-)

chaen:
Ah wait, you should remove getTransportURL from the readMethod then

phicharp:
I'll do it but it is irrelevant as the check for okMethod is done before! ;-)

phicharp:
These changes are included in the next PR, Close

PR #2429, opened 2015-06-16
[v6r13] URGENT fspecify the registration protocol when doing dm.putAndRegister

chaen:
Forgot to specify the protocol, so we were asking for a turl, which srm is not always able to give right away after transferring. 
We need a quick patch with this please


phicharp:
Review OK!
This is an urgent fix as SRM is unable to return a tURL so fast after file transfer

PR #2430, opened 2015-06-16
[v6r14] Minor bug fix to include ARC CEs

rajanandakumar:
Minor bug fix to include ARC among the supported CE types. This could probably go in starting v6r12, but I leave it to the experts to determine this.


rajanandakumar:
Actually, this patch can go in earliest in v6r13 - it will fail in v6r12 as WMSUtilities is not instrumented to support it.

PR #2433, opened 2015-06-16
[v6r13]: Minor Fixes for Request

andresailer:
Pylint error: Error is the member variable _Error does not exist
Disabled error for two false positives.
~~None of this seems to have an impact, but it is weird and probably was not done on purpose.~~
Edit: Removed "_Status is the member variable, Status is the member function"
"_LastUpdate is the member variable, LastUpdate is the member function"

### DIFF
--- a/RequestManagementSystem/Client/Request.py
+++ b/RequestManagementSystem/Client/Request.py
@@ -167,7 +167,7 @@ class Request( object ):
       # # All operations Done -> Done
       elif opStatus == "Done" and self.__waiting == None:
         rStatus = "Done"
-        self._Error = ''
+        self.Error = ''
     self.Status = rStatus
 
   def getWaiting( self ):
--- a/RequestManagementSystem/Client/Request.py
+++ b/RequestManagementSystem/Client/Request.py
@@ -342,7 +342,7 @@ class Request( object ):
 
 
   @hybrid_property
-  def LastUpdate( self ):
+  def LastUpdate( self ): # pylint: disable=E0202
     """ last update getter """
     return self._LastUpdate
 
@@ -356,7 +356,7 @@ class Request( object ):
     self._LastUpdate = value
 
   @hybrid_property
-  def Status( self ):
+  def Status( self ): # pylint: disable=E0202
     """ status getter """
     self._notify()
     return self._Status


RequestManagementSystem/Client/Request.py
```diff
@@ -167,8 +167,8 @@ def _notify( self ):
       # # All operations Done -> Done
       elif opStatus == "Done" and self.__waiting == None:
         rStatus = "Done"
-        self._Error = ''
-    self.Status = rStatus
+        self.Error = ''
+    self._Status = rStatus
```
chaen: While the two other changes are correct, this one is not. You want to go through the setter to update the time and things like this

andresailer: so you mean
```
self.Status( rStatus )
```

chaen: no
self.Status = rStatus

Have a look at what are "properties" in Python, this is it :-)

andresailer: Removed this change from the commit

RequestManagementSystem/Client/Request.py
```diff
@@ -370,7 +370,7 @@ def Status( self, value ):
     # If the status moved to Failed or Done, update the lastUpdate time
     if value in ( 'Done', 'Failed' ):
       if value != self._Status:
-        self.LastUpdate = datetime.datetime.utcnow().replace( microsecond = 0 )
```
chaen: Now, regarding that one, it is more of a best practice... if the internal argument is protected by property, it is because we are "doing something on it". So it is probably best to let it go through the property as well.

andresailer: What do you think about adding
```
# pylint: disable=E0202 
```
to the respective lines to hide the pylint error message?

chaen: But is this general or only on a per File basis or what ?

andresailer: Like this
```
def LastUpdate( self ): # pylint: disable=E0202
```
It is just the line

chaen: Ah cool ! Go for it :-)

andresailer: Ok, done :)

chaen: However, it was a acknowledged bug that was fixed, so it should not happen anymore.... http://www.logilab.org/revision/90049

andresailer: I saw that too, but you are using @hybrid_property, and not setter/getter/deleter or property.
Maybe I should make a pull request for pylint...

chaen: Oh yes right !! Well, anyway, I will get ride of this hybrid_property very very soon. I don't want to expose SQLAlchemy on the client side if possible.

PR #2434, opened 2015-06-16
[v6r13] Minor Fix for StorageElement

andresailer:
This was importing the Module and later trying to call it (line 541), so this needs to import the Class
(also found by pylint).


chaen:
Review OK

PR #2435, opened 2015-06-16
[v6r13] FailoverRequest: no more setters

andresailer:
These setters are no longer available (they are part of the old RequestManagementSystem), must directly assign to variables.


chaen:
review OK

PR #2440, opened 2015-06-17
[v6r13] Make sure we use source replicas with correct checksum

phicharp:
Some replicas don't have a valid checksum registered, which makes  transfers fail.
We check the SE checksum against either the operation or the FC checksum before considering a replica.

Use this opportunity for removing the useless StorageElement caching as StorageElement is already a cache!

### DIFF
--- a/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
+++ b/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
@@ -57,12 +57,13 @@ def filterReplicas( opFile, logger = None, dataManager = None ):
 
   replicas = replicas["Successful"].get( opFile.LFN, {} )
 
-  # FC checksum is more likely better than opFile.Checksum
-  fcMetadata = FileCatalog().getFileMetadata( opFile.LFN )
-  fcChecksum = fcMetadata.get( 'Value', {} ).get( 'Successful', {} ).get( opFile.LFN, {} ).get( 'Checksum', '' )
-  # Replace opFile.Checksum if it doesn't match a valid FC checksum
-  if fcChecksum and ( not opFile.Checksum or not compareAdler( fcChecksum, opFile.Checksum ) ):
-    opFile.Checksum = fcChecksum
+  if not opFile.Checksum:
+    # Set Checksum to FC checksum if not set in the request
+    fcMetadata = FileCatalog().getFileMetadata( opFile.LFN )
+    fcChecksum = fcMetadata.get( 'Value', {} ).get( 'Successful', {} ).get( opFile.LFN, {} ).get( 'Checksum', '' )
+    # Replace opFile.Checksum if it doesn't match a valid FC checksum
+    if fcChecksum:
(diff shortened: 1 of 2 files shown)


DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -59,10 +57,16 @@ def filterReplicas( opFile, logger = None, dataManager = None, seCache = None ):
 
   replicas = replicas["Successful"].get( opFile.LFN, {} )
 
+  # FC checksum is more likely better than opFile.Checksum
+  fcMetadata = FileCatalog().getFileMetadata( opFile.LFN )
```
chaen: opFile.Checksum is set from the FC no ?

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -59,10 +57,16 @@ def filterReplicas( opFile, logger = None, dataManager = None, seCache = None ):
...
+  fcChecksum = fcMetadata.get( 'Value', {} ).get( 'Successful', {} ).get( opFile.LFN, {} ).get( 'Checksum', '' )
+  # Replace opFile.Checksum if it doesn't match a valid FC checksum
+  if fcChecksum and ( not opFile.Checksum or not compareAdler( fcChecksum, opFile.Checksum ) ):
+    opFile.Checksum = fcChecksum
```
chaen: Moreover, should we really just silently replace it ? It would be a very abnormal situation no ?

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
phicharp: In almost all cases it is not set... I think it was there just in case
Transformations don't set it, this is why we have this problem at RAL

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
phicharp: I can add a printout, but is it worth? I am pretty sure that if set, it is OK... The main use case for this test is when it is not set. Then we set it from the FC

phicharp: Of course we need to do that only once.. In that sense we could assume that when it is there it is OK and only get it form the FC when absent...

chaen: This seems more sensitive to me

phicharp: I guess you mean sensible ;-)!

chaen: yes :D

chaen:
review ok

phicharp:
This PR will be superseded with TS fixes...

PR #2442, opened 2015-06-18
[v6r12] DirectoryUsage: avoid defining table in two places

andresailer:
Just create the new table like the old table.


chaen:
Neat :-) But v6r12, really ?

andresailer:
Thanks :)
I guess here I don't really care where this goes in.

PR #2447, opened 2015-06-22
[v6r13] Return S_ERROR if no attributes found

closier:


Core/Security/VOMSService.py
```diff
@@ -96,6 +96,12 @@ def attGetUserNickname( self, dn, ca ):
       result = self.__soapClients[ 'Attributes' ].service.listUserAttributes( user )
     except Exception as e:
       return S_ERROR( "Error in function getUserNickname: %s" % str( e ) )
-    if 'listUserAttributesReturn' in dir( result ):
-      return S_OK( result.listUserAttributesReturn[0].value )
-    return S_OK( result[0].value )
+
+    if result != None:
```
chaen: The correct way of comparing is 'result is not None'

Core/Security/VOMSService.py
```diff
@@ -96,6 +96,12 @@ def attGetUserNickname( self, dn, ca ):
...
+      if 'listUserAttributesReturn' in dir( result ):
+        return S_OK( result.listUserAttributesReturn[0].value )
+
+      return S_OK( result[0].value )
+    else:
+      return S_ERROR( result )
```
chaen: so this return S_ERROR(None)

chaen:
Review OK

PR #2450, opened 2015-06-23
[v6r13] Fix Request.optimize()

chaen:
Since the relationship between entities is now handled by SQLAlchemy, there are "automatic actions" taken on the servers that are not visible on the client side (and hence escaped the tests we had). Basically, when reassigning a File to another Operation, SQLAlchemy removes the link from its old parent to the File, while without it (i.e. on the client side), we must do it by hand. This is done only at one place (Request.optimize), so I corrected it here, and commented.


phicharp:
Review OK

PR #2452, opened 2015-06-23
[v6r13] Few DMS fixes for Alias and BaseSE

phicharp:
* Take into account Alias and BaseSE in site-SE relation
* Remove unnecessary access to RSS from DataManager and use SE.getStatus()
* Add checksumType to RMS files when adding checksum value

### DIFF
--- a/DataManagementSystem/Client/DataManager.py
+++ b/DataManagementSystem/Client/DataManager.py
@@ -680,7 +680,7 @@ class DataManager( object ):
     # Check whether the destination storage element is banned
     log.verbose( "Determining whether %s ( destination ) is Write-banned." % destSEName )
 
-    if not self.__SEActive( destSEName ).get( 'Value', {} ).get( 'Write' ):
+    if not destStorageElement.getStatus().get( 'Value', {} ).get( 'Write', False ):
       infoStr = "Supplied destination Storage Element is not currently allowed for Write."
       log.debug( infoStr, destSEName )
       return S_ERROR( infoStr )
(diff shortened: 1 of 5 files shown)


DataManagementSystem/Client/DataManager.py
```diff
@@ -1543,40 +1543,20 @@ def checkActiveReplicas( self, replicaDict ):
...
-      return S_ERROR( 'SE not known' )
-
-    seStatus = { 'Read' : True, 'Write' : True }
-    if res['Value'][resolvedName].get( 'ReadAccess', 'Active' ) not in ( 'Active', 'Degraded' ):
-      seStatus[ 'Read' ] = False
-    if res['Value'][resolvedName].get( 'WriteAccess', 'Active' ) not in ( 'Active', 'Degraded' ):
-      seStatus[ 'Write' ] = False
-
-    return S_OK( seStatus )
+  def __SEActive( self, se, access = 'Read' ):
```
chaen: Is it a good idea to have a default value for the access type ? Whether an SE is active always depends of what you want to do with it no ?

DataManagementSystem/Utilities/DMSHelpers.py
```diff
@@ -102,10 +116,10 @@ def getSiteSEMapping( self ):
       for site in sites:
         candidateSEs = gConfig.getValue( '/Resources/Sites/%s/%s/SE' % ( grid, site ), [] )
         if candidateSEs:
+          for se in list( candidateSEs ):
```
chaen: Maybe something like : 
candidateSEs.extend([ eqSe for se in candidateSEs for eqSe in equivalentSEs.get(se,[])])
is less tricky, because one day, someone "smart" could think that list(canidateSEs) is pointless because it is already a list. 
OR a comment ;)

phicharp: Thanks, done!

chaen:
Review OK

PR #2453, opened 2015-06-25
Implemented XRSLExtraString

rajanandakumar:
This update has the additional configurable XRSLExtra string as requested by Belle. There is a function added "__getXRSLExtraString" which also contains some documentation on what the function is supposed to do. This code "should just


rajanandakumar:
Completing the previous comment : This code "should just work".

Cheers,
Raja.

PR #2460, opened 2015-06-29
[v6r13] Use SHA1 signature encryption everywhere

atsareg:
CHANGE: X509Chain - use SHA1 signature encryption when generating proxy by Proxy Manager


chaen:
are you sure that this works everywhere ?

atsareg:
It certainly better than the mixture of md5 and sha1 that we had before

chaen:
Yes but I am sorry but doing this in a patch release, without discussing it, without explaining it, without letting any one know (13 minutes between the commit and the merge).... I totally agree that having this mixture is certainly not ideal, but at least we knew it was working

atsareg:
It was not working for at least one crucial case, that's why was put in rather urgently. I have tested in several cases in the France-Grilles installation without seeing problems.

PR #2465, opened 2015-07-01
[v6r12] Modified SiteDirector.py to change the way to count existing pilot jobs for SSH sites

yujikato:
The change in L591 is made basically to include the status "Waiting", whose meaning is the same as "Scheduled" but used for SSH sites. In the current code, when only "Waiting" job exists (no "Running" job) in a SSH site,  accoring to  L251 of Resources/Computing/ComputingElement.py, it is recognized no job exists at this site and SiteDirector submits jobs until running jobs appear. Attached file shows the number of waiting jobs in our site where maximum number of waiting job is set to be several 10. We have tested this modification in r12, which is used at Belle II now.

The change in L508-L509 is already discussed in PR No2456 but we want to include it in r12, too.

diracjob-week


atsareg:
Review OK

PR #2466, opened 2015-07-01
!!! [v6r15] New error handling

chaen:
This PR contains what is necessary for the new error handling described in RFC https://github.com/DIRACGrid/DIRAC/wiki/error_system

It also applies it in part of the DMS. 

A new DError object has to be returned in place of S_ERROR, or used instead of a string in the Failed dictionnary. It is totally backward compatible. It contains an error code and a technical error message. 

The habit should be now taken to compare the errors using cmpError function instead of more or less clever string parsing, and in case of error, always return the original error.


Core/Utilities/DIRACError.py
```diff
@@ -0,0 +1,207 @@
+""" :mod: DIRACError
```
phicharp: I would suggest to call it DErrno rather than DIRACError as it is confusing to have DError and DIRACError

Core/Utilities/DIRACError.py
```diff
@@ -0,0 +1,207 @@
...
+      :param candidate : error code to compare with
+
+      :return True or False
+
+      If a DError instance is passed, we compare the code with DError.errno
+      If it is a Integer, we do a direct comparison
+      If it is a String, we use compatErrorString and strerror to check the error string
+  """
+
+  if type( inErr ) == str :  # old style
```
phicharp: use is instance() as below

DataManagementSystem/Client/DataManager.py
```diff
@@ -350,6 +350,7 @@ def __getFile( self, lfn, replicas, metadata, destinationDir ):
     if not res['OK']:
       return res
     errTuple = ( "No SE", "found" )
+    errToReturn = S_ERROR()
```
phicharp: This seems useless ;-)

chaen:
Good points. Corrected

atsareg:
This branch has conflicts, please resolve. The update PR should be better done against the rel-v6r15 branch

chaen:
Superseded by https://github.com/DIRACGrid/DIRAC/pull/2551

PR #2468, opened 2015-07-02
[v6r13] Check the SE status when doing remove file

chaen:
Currently, when a site/SE is in downtime, the RMS keep trying again and again. Now, not anymore


phicharp:
Review OK

atsareg:
Should it be complemented by setting a notBefore stamp for let say an hour in the future to avoid even picking up those requests ?

chaen:
It is done in the checkSEsRSS method :-)

PR #2470, opened 2015-07-02
!!! [v6r12] Add dealing with status 22 to SRM2Storag.exists

andresailer:
I've had the problem for quite some time, that my FailoverRequests failed because the file already existed on the cern Castor, it had size 0 and I had to remove it by hand.
Now I figured out that the reason is that the SRM2Storage.exists function encountered status 22
(added some debug printout of urldict etc. during investigation)
```
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: SRM2Storage.exists: Checking the existance of 1 path(s) 
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: SRM2Storage.exists urlDict:  {'status': 0, 'estimated_wait_time': 0, 'stat': [16893, 0L, 0, 1, 2, 2, 0L, 0L, 1435832987L, 1435832987L], 'SRMReqID': None, 'locality': 'ONLINE_AND_NEARLINE', 'explanation': None, 'turl': None, 'surl': '/castor/cern.ch/grid/ilc/prod/clic/365gev/yyveyx_o/ILD/SIM/00005250/000'}  
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: __putFile: Checking for existance of file True  
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: SRM2Storage.exists: Checking the existance of 1 path(s) 
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: SRM2Storage.exists urlDict:  {'status': 22, 'estimated_wait_time': 0, 'SRMReqID': None, 'explanation': '[SE][Ls][SRM_FILE_BUSY] <none>', 'ErrorMessage': '[SE][Ls][SRM_FILE_BUSY] <none>', 'turl': None, 'surl': '/castor/cern.ch/grid/ilc/prod/clic/365gev/yyveyx_o/ILD/SIM/00005250/000/yyveyx_o_sim_5250_587.slcio'}  
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: SRM2Storage.exists pathSURL: {'OK': True, 'Value': 'srm://srm-public.cern.ch:8443/srm/managerv2?SFN=/castor/cern.ch/grid/ilc/prod/clic/365gev/yyveyx_o/ILD/SIM/00005250/000/yyveyx_o_sim_5250_587.slcio'}  
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage  ERROR: SRM2Storage.exists: Failed to get path metadata. srm://srm-public.cern.ch:8443/srm/managerv2?SFN=/castor/cern.ch/grid/ilc/prod/clic/365gev/yyveyx_o/ILD/SIM/00005250/000/yyveyx_o_sim_5250_587.slcio: [SE][Ls][SRM_FILE_BUSY] <none>
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/SRM2Storage   INFO: __putFile: Failed to find pre-existance of destination file: SRM2Storage.exists: Failed to get path metadata. [SE][Ls][SRM_FILE_BUSY] <none> 
2015-07-02 11:29:52 UTC RequestManagement/RequestExecutingAgent/00005250_00000587_job_12539118/0/ReplicateAndRegister  ERROR: DataManager error: DataManager.replicateAndRegister: Completely failed to replicate file. 
```
which caused the __putFile to fail in the end.

### DIFF
--- a/Resources/Storage/SRM2Storage.py
+++ b/Resources/Storage/SRM2Storage.py
@@ -80,6 +80,8 @@ class SRM2Storage( StorageBase ):
     self.gfalLongTimeOut = gConfig.getValue( "/Resources/StorageElements/GFAL_LongTimeout", 1200 )
     # # gfal retry on errno.ECONN
     self.gfalRetry = gConfig.getValue( "/Resources/StorageElements/GFAL_Retry", 3 )
+    # # should busy files be considered to exist
+    self.busyFilesExist = gConfig.getValue( "/Resources/StorageElements/SRMBusyFilesExist", False )
 
     # # set checksum type, by default this is 0 (GFAL_CKSM_NONE)
     self.checksumType = gConfig.getValue( "/Resources/StorageElements/ChecksumType", None )
@@ -763,7 +765,7 @@ class SRM2Storage( StorageBase ):
         if urlDict['status'] == 0:
           self.log.debug( "SRM2Storage.exists: Path exists: %s" % pathSURL )
           successful[pathSURL] = True
-        elif urlDict['status'] == 22:
+        elif urlDict['status'] == 22 and self.busyFilesExist:
           self.log.debug( "SRM2Storage.exists: Path exists, file busy (e.g., stage-out): %s" % pathSURL )
           successful[pathSURL] = True
         elif urlDict['status'] == 2:
--- a/Resources/Storage/SRM2Storage.py
+++ b/Resources/Storage/SRM2Storage.py
@@ -763,6 +763,9 @@ class SRM2Storage( StorageBase ):
         if urlDict['status'] == 0:
           self.log.debug( "SRM2Storage.exists: Path exists: %s" % pathSURL )
           successful[pathSURL] = True
+        elif urlDict['status'] == 22:
+          self.log.debug( "SRM2Storage.exists: Path exists, file busy (e.g., stage-out): %s" % pathSURL )
+          successful[pathSURL] = True
         elif urlDict['status'] == 2:
           self.log.debug( "SRM2Storage.exists: Path does not exist: %s" % pathSURL )
           successful[pathSURL] = False


chaen:
I have been thinking about this fix for a while, but can't decide whether it is good or not. We have similar problem recently when transferring from the pit to castor. The problem is that there is a race condition here, that we have observed, and you too now. Basically what happens is that you start a process in the RMS to copy the file, it times out and you kill it. But in fact, I am pretty sure it is not properly killed. Then later on you want to execute it again. As it is now, the Operation will fail because it can't determine the existence of the file. But if you say "true" for status 22, then you will remove the file entry in the srm database, and copy your file, and the RMS will be happy. BUT, the first process, might take long to finish, and when this happens, when gfal will try to do the PUT_DONE, realize that it is not possible because it was already done by newer process, and say "ah, transfer went wrong, I clean what I did", and he will remove your file. As a consequence, this file will be considered successfully transferred by the RMS, be in your catalog, but not on CASTOR.

andresailer:
And here I was thinking I was the only one with this issue... How are you dealing with this problem?

I think I only ever get this problem, when there is a castor issue and the filesize of the file remnant has 0 size. The file remnant is created from the upload by jobs, so I am pretty confident there is no other task trying to copy the file. When I see there are completed jobs I would wait some time and clean the file remnants and reset the failover requests by manually running a script.

chaen:
we observe this only when transferring  the raw files from the pit, so when we put castor under heavy load. Currently, I deal with it manually as you do, but we ask the CASTOR guys to tune their system, so that this does not happen anymore.

atsareg:
I think we should try to think in terms of how to minimize the damage of this SRM2 error. The case that Chris is describing can happen but is very rare if ever. The André's case looks to be more likely to me. So, fixing it this way is not solving all the cases but minimizes the problems of today.

chaen:
Yep, but it creates the pit hole of loosing files without realizing at all, which is worse in my opinion

andresailer:
To make everyone happy one can add a flag "ignoreStatus22".

chaen:
Sounds like a good idea, the default being the old behavior

atsareg:
@Chris: you want to make your ( LHCb's ) life easier :) . However the more frequent is the other case and what is more important for less expert users ( this is not about you André ) in other communities. So,
it is better added to the LHCb Castor SE config in one place rather than putting in many SRM SEs for the others.

andresailer:
Thanks, Andrei :)
But I agree with Chris that the default behaviour shouldn't change (except for bug fixes), and definitely not in patch releases. I recently made some bad experiences with somebody removing default shifter values.

But then changes always break things for somebody  https://xkcd.com/1172/  :)

Something like this? (I don't quickly see how to get parameters from specific CEs into this place...)
```
diff --git a/Resources/Storage/SRM2Storage.py b/Resources/Storage/SRM2Storage.py
index 517929e..00c15c9 100755
--- a/Resources/Storage/SRM2Storage.py
+++ b/Resources/Storage/SRM2Storage.py
@@ -80,6 +80,8 @@ class SRM2Storage( StorageBase ):
     self.gfalLongTimeOut = gConfig.getValue( "/Resources/StorageElements/GFAL_LongTimeout", 1200 )
     # # gfal retry on errno.ECONN
     self.gfalRetry = gConfig.getValue( "/Resources/StorageElements/GFAL_Retry", 3 )
+    # # should busy files be considered to exist
+    self.busyFilesExist = gConfig.getValue( "/Resources/StorageElements/SRMBusyFilesExist", False )
 
     # # set checksum type, by default this is 0 (GFAL_CKSM_NONE)
     self.checksumType = gConfig.getValue( "/Resources/StorageElements/ChecksumType", None )
@@ -495,7 +497,7 @@ class SRM2Storage( StorageBase ):
         elif urlDict['status'] == 1:
           self.log.debug( "prestageFile: File found to be already staged.", pathSURL )
           successful[pathSURL] = urlDict['SRMReqID']
-        elif urlDict['status'] == 22:
+        elif urlDict['status'] == 22 and self.busyFilesExist:
           self.log.debug( "prestageFile: Stage request for file %s queued.", pathSURL )
           successful[pathSURL] = urlDict['SRMReqID']
         elif urlDict['status'] == 2:
```
Then the default value be changed in v6r14 or whenever.

chaen:
Fine with me. Andrei, true, I want to make my life easier :D However, Andre is correct that one should not change the behavior in a patch (even though this is between feature and patch). But more importantly, the situation as it is now involves manual intervention (so, not really simplifying my life :D) to keep going. But with this fix, we risk to irremediably lose data, which I am worried about

andresailer:
Can someone give me some hints how to make this flag SE parameter, or is a global flag good enough?

chaen:
What do you need exactly ? You can check for example the getStorageElementOptions method in the CS helpers which seems to be doing what you want. But I would say that a global SE flag is better. Less operational hassle when debugging

andresailer:
Thanks, I'll look at the getStorageElementsOptions, unless @atsareg is also happy with the global flag.

atsareg:
I agree that the global flag is more appropriate here

andresailer:
This "!!!" means conflicts, right? Anything I need to do?

atsareg:
!!! Means that the PR can not be merged for whatever reason. I marked it considering that the PR discussion did not converge, but it actually did. So, I merged it but did not change the title

PR #2481, opened 2015-07-06
[v6r12] FileManagerBase fix in addReplicas

andresailer:
_findFiles can return S_ERROR, need to check.

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
@@ -793,6 +793,8 @@ class FileManagerBase( object ):
     connection = self._getConnection( connection )
     successful = {}
     res = self._findFiles( lfns.keys(), ['DirID', 'FileID', 'Size'], connection = connection )
+    if not res['OK']:
+      return res
     failed = res['Value']['Failed']
     for lfn in failed.keys():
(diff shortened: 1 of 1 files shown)


DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
```diff
@@ -793,6 +793,8 @@ def _addReplicas( self, lfns, connection = False ):
     connection = self._getConnection( connection )
     successful = {}
     res = self._findFiles( lfns.keys(), ['DirID', 'FileID', 'Size'], connection = connection )
+    if not res['OK']:
+      return S_ERROR( "Failed to run _findFiles" )
```
chaen: The fix is a good idea, but please return the original error, otherwise we have no way to track down what happens :-)

andresailer: Do you mean just, ```return res``` or something more?

chaen: return res is perfect :-)

andresailer: Changed to ```return res```

chaen:
Review OK

PR #2486, opened 2015-07-07
[v6r13] SandboxStoreClient is vo aware in the web portal context

atsareg:
FIX: SandboxStoreClient - get the vo info from the delegatedGroup argument to use for the StorageElement instantiation
FIX: BatchSystems - always return Message in case of errors ( cherry-picked from integration )


chaen:
Review OK for me

PR #2487, opened 2015-07-07
!!! [v6r15] Implement recursive chown, chgrp and chdir for LHCb managers

chaen:
Propagates the recursive flag till the end, and use it in the SP managers


atsareg:
This PR is to be applied after the #2554 is merged in. In principle, should work for the LHCb managers out of the box with may be some minor modifications.

atsareg:
I guess this PR is to be reimplemented

PR #2491, opened 2015-07-09
[v6r13] URGENT stupid bugs that surprisingly were not found during tests

chaen:
The ReqClient fixes is to avoid the dirac-rms-show-request to crash in some cases
The two others are stupid mistakes...


atsareg:
I guess those changes are tested by hot fixes in LHCb ?

chaen:
Yes. The others were as well, but they were working with the previous patch interestingly :-)

PR #2494, opened 2015-07-10
[v6r12] SRM2Storage, busyfile new return code?; GraphData protect against missing key

andresailer:
a) I am not sure, but now the Castor-SRM return status 16 for SRM_FILE_BUSY. I don't know if this is a new return code for the same issue as before, or a different Issue.

b) I was trying to plot WMSHistory NumberOfJobs grouped by Status, limited to two groups , limited to the last two days and got an exception for a missing key in the parsed_errors

### DIFF
--- a/Core/Utilities/Graphs/GraphData.py
+++ b/Core/Utilities/Graphs/GraphData.py
@@ -400,9 +400,9 @@ class PlotData:
     else:
       self.keys = self.sortKeys()
 
-    self.values = [ self.parsed_data[k] for k in self.keys ]
-    self.errors = [ self.parsed_errors[k] for k in self.keys ]
-    values_to_sum = [ self.parsed_data[k] for k in self.keys if k != '' ]
+    self.values = [ self.parsed_data.get(k, 0.0) for k in self.keys ]
+    self.errors = [ self.parsed_errors.get(k, 0.0) for k in self.keys ]
+    values_to_sum = [ self.parsed_data.get(k, 0.0) for k in self.keys if k != '' ]
 
     self.real_values = []
     for k in self.keys:
--- a/Resources/Storage/SRM2Storage.py
+++ b/Resources/Storage/SRM2Storage.py
@@ -766,7 +766,7 @@ class SRM2Storage( StorageBase ):
         if urlDict['status'] == 0:
           self.log.debug( "SRM2Storage.exists: Path exists: %s" % pathSURL )
           successful[pathSURL] = True
-        elif urlDict['status'] == 22 and self.busyFilesExist:
+        elif urlDict['status'] in ( 16, 22 ) and self.busyFilesExist:
           self.log.debug( "SRM2Storage.exists: Path exists, file busy (e.g., stage-out): %s" % pathSURL )
           successful[pathSURL] = True
         elif urlDict['status'] == 2:


chaen:
Hum, this is strange... did you look at the castor specification to see if they don't correspond to different situation ?

andresailer:
No idea where I would find that information. Not even sure if this is due to the new dirac version, new externals, lcg bundles, or the castor update for castor public that was done one Wednesday...

chaen:
There was a change in the srm-ifce version that changed an 11 to a 22 . it maybe changed that as well... I'll check with Alejandro if that is really the case. Because if it is CASTOR specific, and then the other storage have another error code, then we are doomed :-)

chaen:
but wait, which new Dirac version ? I see you commit against v6r12, so you are not speaking about v6r13 ? the one I refer to was only starting at some point in v6r13 if I recall correctly

andresailer:
I changed from v6r11 to v6r22 on Wednesday, also changed LCGBundles. Probably also different server externals.

chaen:
well I am afraid that without more detailed versions, I can't say very much. But it would still be very interesting to know where would the difference be

Core/Utilities/Graphs/GraphData.py
```diff
@@ -401,7 +401,7 @@ def initialize( self ):
       self.keys = self.sortKeys()
 
     self.values = [ self.parsed_data[k] for k in self.keys ]
```
zmathe: Could you please also add a protection to this line as well?
self.values = [ self.parsed_data[k] if k in self.parsed_data else for k in self.keys ]

Core/Utilities/Graphs/GraphData.py
```diff
@@ -401,7 +401,7 @@ def initialize( self ):
...
-    self.errors = [ self.parsed_errors[k] for k in self.keys ]
+    self.errors = [ self.parsed_errors[k] if k in self.parsed_errors else 0.0 for k in self.keys ]
     values_to_sum = [ self.parsed_data[k] for k in self.keys if k != '' ]
```
zmathe: and also to this line:
values_to_sum = [ self.parsed_data[k] if k in self.parsed_data else 0.0 for k in self.keys if k != '' ]

Core/Utilities/Graphs/GraphData.py
zmathe: ops sorry: self.values = [ self.parsed_data[k] if k in self.parsed_data else 0.0 for k in self.keys ]

andresailer:
Added the key checks

Core/Utilities/Graphs/GraphData.py
```diff
@@ -400,9 +400,9 @@ def initialize( self ):
     else:
       self.keys = self.sortKeys()
 
-    self.values = [ self.parsed_data[k] for k in self.keys ]
-    self.errors = [ self.parsed_errors[k] for k in self.keys ]
-    values_to_sum = [ self.parsed_data[k] for k in self.keys if k != '' ]
+    self.values = [ self.parsed_data[k] if k in self.parsed_data else 0.0 for k in self.keys ]
```
chaen: self.parsed_data.get(k, 0.0) is maybe clearer isn't it ?

andresailer: Now that you mention it :)

atsareg:
Any more details about the SRM return codes ?

andresailer:
Not from my side. I can drop this part if people prefer.

atsareg:
I will merge this now. @chaen: please try to check if this is Castor specific with Alejandro or whoever appropriate. Would be nice to avoid <if backend == "Castor"...> but may be it is necessary

chaen:
I repeat "I am afraid that without more detailed versions, I can't say very much".
But by just looking at the release notes of SRM-IFCE : https://its.cern.ch/jira/browse/DMC-637

andresailer:
I think I had moved to the lcgBundles 2015-07-06, after which I noticed that the return code was different. However on the same day there was an upgrade of castorpublic, I have no idea which part is responsible for the error code. By now I moved back to 2015-01-09 on my servers.

PR #2499, opened 2015-07-13
[v6r13] Fixing relative paths for put and get operations

fibbo:
Relative paths didn't work because the 'file:' protocol specifier (that gfal2 needs) was set as such that it interpreted all paths as absolute, resulting in an error for relative paths.


andresailer:
You can use 
```git rebase --interactive HEAD~~~~``` (I think four '~', maybe five) to really get rid of the commit from this branch instead of adding another commit  to undo the first one. (Be careful to not lose the commit for ever) and then ```git push --force``` to this branch.

Or make a new branch (from origin/rel-v6r13) and cherry pick the commits you want to have, and make a new PR.

fibbo:
closing this one and creating a new PR on a clean branch as suggested.

PR #2503, opened 2015-07-16
[v6r12] Fixes for the pilot on osg and globus sites

andresailer:
Solves   #2462 
With these changes I can run on sites that define OSG_WN_TMP

And with the Job Contact string the connection from Pilot to Job is done now also for jobs running on GlobusCEs 

EDIT: for v6r11 of course this would be somewhere completely different

### DIFF
--- a/WorkloadManagementSystem/PilotAgent/pilotCommands.py
+++ b/WorkloadManagementSystem/PilotAgent/pilotCommands.py
@@ -486,6 +486,11 @@ class ConfigureSite( CommandBase ):
     if os.environ.has_key( 'OSG_WN_TMP' ):
       self.pp.flavour = 'OSG'
 
+    #GLOBUS Computing Elements
+    if 'GLOBUS_GRAM_JOB_CONTACT' in os.environ:
+      self.pp.flavour = 'GLOBUS'
+      pilotRef = os.environ['GLOBUS_GRAM_JOB_CONTACT']
+
     self.__doOSG()
 
     # Direct SSH tunnel submission
--- a/WorkloadManagementSystem/PilotAgent/pilotCommands.py
+++ b/WorkloadManagementSystem/PilotAgent/pilotCommands.py
@@ -491,8 +491,6 @@ class ConfigureSite( CommandBase ):
       self.pp.flavour = 'GLOBUS'
       pilotRef = os.environ['GLOBUS_GRAM_JOB_CONTACT']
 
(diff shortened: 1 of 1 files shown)


andresailer:
Don't merge this. Somehow it seems that creating a subfolder on the OSG sites is necessary and the GLOBUS flavour needs to go. Will push the fix tomorrow.

andresailer:
Lost one of my dirac servers today, will get back to this next week.

andresailer:
The actual problem was that "vo" was not defined at the end of this doOSG function. 
Still not sure if any of that is still needed...

WorkloadManagementSystem/PilotAgent/pilotCommands.py
```diff
@@ -559,10 +564,11 @@ def __doOSG ( self ):
     """ Treat the OSG case """
 
     osgDir = ''
+    vo = self.pp.releaseProject.replace( 'DIRAC', '' ).upper()
```
atsareg: This is certainly not a correct way to evaluate the VO. The best way to do that is to get it based on the pilot proxy DIRAC group

andresailer: Before fixing this, can we not just drop the whole doOSG function?

atsareg: This function essentially creates extra directory for the pilot, apparently there was a good reason for that although this looks strange as this should be handled by the batch system. I do not see a good reason for defining VO_%s_SW_DIR variable as there are no uses of it unless there will be a pilot command for checking the app software. As for the VO, I think pilot can get an option /DIRAC/VirtualOrganization since it is not mixing different VOs. However, I do not know if vo.upper() to be used as it is done on doOSG() will be correct for all the VOs. Please,decide what do you actually need for the OSG case based on your experience.

andresailer: I just removed the whole doOSG function. My pilots worked without creating another subfolder and all the workingdirectories are created by the batch system as you say.

andresailer:
Removed the doOSG funciton, it is not needed.

atsareg:
As far as the code is concerned, it looks ok. I rely on your experience with the OSG for the logic. Thanks

andresailer:
Thanks, it definitely works better than before without this function :)

PR #2506, opened 2015-07-17
[v6r13] fixing relative path issues

fibbo:
Relative paths didn't work because the 'file:' protocol specifier (that gfal2 needs) was set as such that it interpreted all paths as absolute, resulting in an error for relative paths.

This is the same PR as before but on a clean branch.


chaen:
Review OK

PR #2507, opened 2015-07-20
[v6r14] Updates in the DIRAC installation tools

atsareg:
NEW: dirac-install - makes us of the DIRAC tar files in CVMFS if available
NEW: dirac-install-client - a guiding script to install the DIRAC client from A to Z
BUGFIX: CSGlobals - bug in the getCSExtensions()

### DIFF
--- a/ConfigurationSystem/Client/Helpers/CSGlobals.py
+++ b/ConfigurationSystem/Client/Helpers/CSGlobals.py
@@ -40,8 +40,8 @@ class Extensions( object ):
       exts = gConfig.getValue( '/DIRAC/Extensions', [] )
       for iP in range( len( exts ) ):
         ext = exts[ iP ]
-        if ext.find( "DIRAC" ) == len( ext ) - 5:
-          ext = ext[:5]
+        if ext.endswith( "DIRAC" ):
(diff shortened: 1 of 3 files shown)


chaen:
I thought the feature list was closed for v6r14 ?

atsareg:
This is just an addition which does not change the current functionality. And a bug fix

sposs:
Hi,
I'm rather against any new feature even if minor... After all, we need to
follow our own rules, otherwise it will become a mess...

Bug fixes are of course needed, but I think they should have their own
commit.
Cheers,
SP

chaen:
I support Stephane. By the way, in the BILD of 2 weeks ago, you said you would make the release of v6r14. Where are we with this ?

atsareg:
I appreciate you following the rules but let's introduce another one as well. New additions which are not interfering with the already fixed functionality and will pass unnoticed by anybody who is not aware of them can be added at any moment to allow trial in a production environment.

chaen:
I still don't think it is a good idea. Thanks to the git branch mechanism, you can very well make a "pre" like release that you would deploy in your environment.

PR #2512, opened 2015-07-24
[v6r14] Added show doc command

Sbalbp:
As requested in #1013, a new command has been added to get information about services and agents' documentation.


atsareg:
This does not take into account docs of the modules imported from extensions

PR #2517, opened 2015-07-29
[v6r13] SEActive: Add mandatory vo parameter

andresailer:
I think this is missing here.


phicharp:
Review OK

atsareg:
In which context this was a problem ? Just for my understanding please.

andresailer:
InputData resolution does not work without adding the vo here

PR #2522, opened 2015-08-04
[v6r14] Updates for ARC/CREAM CE code

atsareg:
Several updates to remove obsoleted stuff and to use ARC/CREAM ComputingElements consistently with any other CEs.


andresailer:
Could the __prepareProxy be moved to ComputingElement? There is a bit of code in the GlobusComputingElement getPilotOutput function that also gets a proxy. But actually I think with the changes in the WMSAdministator the proxy is set up when the WMSAdministrator calls the CE functions.

atsareg:
I think this is a good idea

PR #2527, opened 2015-08-06
[v6r13] Fix to RMS

phicharp:
This fixes a bug that was there forever when using the file owner's proxy for removing a file or a replica.

It also increases the timeout for listing a directory of user as some uses have 10,000's directories in their top directory...

### DIFF
--- a/DataManagementSystem/Agent/RequestOperations/RemoveReplica.py
+++ b/DataManagementSystem/Agent/RequestOperations/RemoveReplica.py
@@ -159,11 +159,11 @@ class RemoveReplica( DMSRequestOperationsBase ):
         # #  you're a data manager - save current proxy and get a new one for LFN and retry
         saveProxy = os.environ["X509_USER_PROXY"]
         try:
-          proxyFile = self.getProxyForLFN( opFile.LFN )
-          if not proxyFile["OK"]:
-            opFile.Error = proxyFile["Message"]
+          fileProxy = self.getProxyForLFN( opFile.LFN )
+          if not fileProxy["OK"]:
+            opFile.Error = fileProxy["Message"]
           else:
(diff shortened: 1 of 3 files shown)


chaen:
Review OK

atsareg:
The requests must be executed with credentials of the owner of a request. As it is now, anybody can make a request to remove anybody's else files. I suggest that this review of the RMS operations to be the main new feature of the v6r15

chaen:
It is normally executed with the credentials of the owner. This is just a because of the permission mess that we always had that in case the request is executed by the DataManager, we allow it to take yet another proxy to try to remove the files.
Of course, anyone can create a request to delete anyone's else file, but it won't work better than if you try with a script.

atsareg:
We have to abandon the logic of shifter in the RMS. Requests have their owner, full stop. The owner of the request must be verified at the time of the request submission.

phicharp:
Yes of course, but this will only be possible when we have introduced “groups of groups” otherwise there is no way we can manage data in a proper way since we don’t know which owner requests from TS must have… So it is not only the RMS but also the TS that must be changed!

> Le 7 août 2015 à 13:49, Andrei Tsaregorodtsev <notifications@github.com> a écrit :
> 
> We have to abandon the logic of shifter in the RMS. Requests have their owner, full stop. The owner of the request must be verified at the time of the request submission.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/DIRACGrid/DIRAC/pull/2527#issuecomment-128681811>.
> 

-- 
Dr Philippe Charpentier
Office 2-R-007, Tel : +41 22 767 4244 ,  GSM :  +41 75 411 0167
mailto:Philippe.Charpentier@cern.ch  http://cern.ch/Philippe.Charpentier

PR #2537, opened 2015-08-13
[v6r14] Fixes in SE machinery

atsareg:


Resources/Storage/StorageFactory.py
```diff
@@ -89,7 +89,7 @@ def getStorages( self, storageName, pluginList = None ):
     if pluginList is None:
       pluginList = []
     if not self.vo:
-      return S_ERROR( 'Mandatory vo parameter is not defined' )
+      gLogger.warn( 'No VO information available' )
```
chaen: why isn't it mandatory anymore ?

atsareg: It is used if it is defined ( for some possible VO specific parameters ). If not defined, nothing specific. For example, like that it allows usage by dirac_admin group which is not hooked up on any VO. We have to think more about VO-group-user relations still. However, for the moment this is more harm than help if the VO parameters is mandatory

chaen: I see... For us it does not change anything anyway, however you are right, one should think more in depth about the multi VO paradigm. But maybe, one should do it before doing small patches here and there ?

chaen:
Review OK

Resources/Storage/StorageFactory.py
andresailer: Why is this still mandatory in v6r13? Changing this also allows us to download sandboxes as diracAdmin.

atsareg:
Indeed, this PR should got o v6r13 patch. Will do.

atsareg:
Redirected to rel-v6r13

PR #2545, opened 2015-08-19
[v6r12] Add XROOT to registration protocols

andresailer:
This is needed to run replicateAndRegister to SEs where only the XROOT protocol is available


chaen:
review ok

PR #2547, opened 2015-08-19
[v6r12] ARCComputingElement: check against missing ceParameters

andresailer:
Check if WorkingDirectory is set
Create WorkingDirectory if it does not exist

This also happens now that the PilotStatusAgent is calling killJob


atsareg:
Note that ARC CE is completely new in v6r14

andresailer:
I know that the ARCCE is re-written, but I don't know when I will update, so right now I have all these hotfixes

PR #2550, opened 2015-08-20
[v6r13] Fix in DM and add MJF as test

phicharp:
* Fixes a problem when a file is set Problematic in the FC but indeed doesn't exist at all
* Adds reading CPU power from MJF for comparison with our script


atsareg:
Is it indeed to v6r13 ? Thn make it to rel-v6r13, please

phicharp:
Grrr, stupid me! github is nice but should allow to change the branch in a PR... I close and redo it!

PR #2553, opened 2015-08-21
[v6r14] Plugin for file protocol

chaen:
Implement a plugin to use directly the file protocol. Tests included

### DIFF
--- a/Resources/Storage/FileStorage.py
+++ b/Resources/Storage/FileStorage.py
@@ -1,5 +1,5 @@
 """
- This is the XROOTD StorageClass
+ This is the File StorageClass
  """
 
 __RCSID__ = "$Id$"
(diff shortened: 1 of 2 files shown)


Resources/Storage/FileStorage.py
```diff
@@ -0,0 +1,629 @@
...
+  def __stat( path ):
+    """  Issue a stat call and format it the dirac way, and add the checksum
+
+      :param self: self reference
+      :param path: path on the storage
+      :returns Successful S_OK(metadataDict) or S_ERROR
+    """
+    try:
+      statInfo = os.stat( path )
+      metadataDict = {}
```
phicharp: Probably it would be worth returning all information usually returned by SRM (Unavailable, Cached etc...) as I am not sure we are protected everywhere with .get()

chaen: yeah... will fix, and just wrote on my list to clean these metadata thing everywhere

phicharp:
Final review: OK

PR #2554, opened 2015-08-21
[v6r15] Recursive operations in the FileCatalog

atsareg:
The main point is that recursive operations to change path owner, group and mode are finally available. Some reorganization of the code is made to make the file and directory set and change operations 
symmetrical. 

Some mixed in changes in the doc strings to allow doc generation by sphinx.


DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
```diff
@@ -336,148 +337,152 @@ def getDirectoryParameters( self, path ):
...
+                                          self.setDirectoryMode,
+                                          self.db.fileManager.setFileMode,
+                                          recursive = recursive )
 
-    result = checkArgumentFormat( paths )
-    if not result['OK']:
-      return result
-    arguments = result['Value']
+#####################################################################
+  def changeDirectoryParameter( self, paths,
```
chaen: This method is absolutely not compatible with the changes I have done. This implements the recursive behavior at a higher level than the method actually changing the parameter (setDirectoryParameter), while I was propagating the recursive info till the very end for performance reasons

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
```diff
@@ -580,14 +585,17 @@ def getFileIDsInDirectory( self, dirID, credDict, startItem = 1, maxItems = 25 )
     return result
 
 
-  def getFilesInDirectory( self, dirID, credDict ):
+  def getFilesInDirectory( self, dirID, credDict, requestString = False ):
```
chaen: Why is this one needed if there is already one in the FileManagerBase ?

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
```diff
@@ -1257,14 +1207,14 @@ def _checkLFNPFNConvention( self, lfn, pfn, se ):
     result = self._getStorageElement( se )
```
chaen: I still believe that this method should not be here, and even more that we should not instanciate an SE in a catalog

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
atsareg: Yes, I know. And I think this is correct as it keeps independent Directory and File Manager implementations. On the other hand the real update query is a single one, so the performance should not be affected.

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
atsareg: This one is different. It is a simple one to just get the list of files explicitly or as a query without taking into account visibility logic, etc. But I agree that this should be rather moved to the File Manager implementation. All files must have a reference dirID but can be stored differently in different implementations. So, this would me more generic.

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
atsareg: Agreed. It can be just dropped for the moment ( _checkLFNPFNConvention is not actually used ). The way how the SE information is used in the DFC is one more thing to be sorted out. As we discussed, this should be rather moved to the client.

DataManagementSystem/DB/FileCatalogDB.py
```diff
@@ -262,62 +262,118 @@ def hasAccess( self, opType, paths, credDict ):
...
+      return result
+    for p in result['Value']['Successful']:
+      if result['Value']['Successful'][p]:
+        dirList.append( p )
+    fileList = []
+    if len( dirList ) < len( paths ):
+      result = self.isFile( paths, credDict )
+      if not result['OK']:
+        return result
+      fileList = result['Value']['Successful'].keys()
```
chaen: This is not good, since in Successful you have True or False

DataManagementSystem/DB/FileCatalogDB.py
```diff
@@ -262,62 +262,118 @@ def hasAccess( self, opType, paths, credDict ):
...
+        :param dict credDict: dictionary of the caller credentials
+        :param boolean recursive: flag to apply the operation recursively
+    """
+    res = self._checkPathPermissions( 'changePathGroup', paths, credDict )
     if not res['OK']:
       return res
-    failed.update( res['Value']['Failed'] )
-    successful = res['Value']['Successful']
+    failed = res['Value']['Failed']
+    result = self.__changePathFunction( res['Value']['Successful'], credDict,
```
chaen: You must test here and in the others that res['Value']['Successful'] is not empty to avoid the "ERROR:empty input[]" problem that I fixed in a previous PR

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
```diff
@@ -336,148 +337,156 @@ def getDirectoryParameters( self, path ):
 #####################################################################
   def _setDirectoryParameter( self, path, pname, pvalue ):
```
chaen: In this method, you are using the actual table name explicitly, so I think you should not implement it here, but in the DirectoryLevelTree

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
```diff
@@ -336,148 +337,156 @@ def getDirectoryParameters( self, path ):
...
-      else:
-        successful[path] = True
+        continue
+      if recursive:
+        result = self.__getDirID( path )
+        if not result['OK']:
+          failed[path] = result['Message']
+          continue
+        dirID = result['Value']
+        result = self.getSubdirectoriesByID( dirID, requestString = True, includeParent = True )
```
chaen: if you include the parent, why do you need to call directoryFunction in line 453 ?

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
```diff
@@ -259,7 +246,7 @@ def _addFiles( self, lfns, credDict, connection = False ):
     extraLfns = {}
     for lfn in lfns:
       masterLfns[lfn] = dict( lfns[lfn] )
-      if 'SE' in lfns[lfn] and type( lfns[lfn]['SE'] ) == ListType:
+      if 'SE' in lfns[lfn] and isinstance( lfns[lfn]['SE'], list ):
```
chaen: if isinstance(lfns[lfn].get('SE'), list):
?

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
```diff
@@ -1157,6 +1111,16 @@ def _getIntStatus(self,statusID,connection=False):
...
 
+  def getFileIDsInDirectory( self, dirID, requestString = False ):
+    """ Get a list of IDs for all the files stored in given directories or their
+        subdirectories
+    :param mixt dirID: single directory ID or a list of directory IDs
+    :param boolean requestString: if True return result as a SQL SELECT string
+    :return: list of file IDs or SELECT string
+    """
+
+    return self._getDirectoryFileIDs( dirID, requestString )
```
chaen: Since requestString is a named parameter, it should be called as such

DataManagementSystem/DB/FileCatalogComponents/WithFkAndPs/DirectoryClosure.py
```diff
@@ -522,33 +522,6 @@ def _setDirectoryParameter( self, path, pname, pvalue ):
     else:
       return DirectoryTreeBase._setDirectoryParameter( self, path, pname, pvalue )
 
-
-
-  def setDirectoryGroup( self, path, gname ):
```
chaen: Please leave this file untouched, I will need these methods (with an underscore in front...)

DataManagementSystem/DB/FileCatalogDB.py
```diff
@@ -262,62 +262,126 @@ def hasAccess( self, opType, paths, credDict ):
...
-    if not res['OK']:
-      return res
-    failed.update( res['Value']['Failed'] )
-    successful = res['Value']['Successful']
+    dirArgs = {}
+    fileArgs = {}
+
+    for path in paths:
+      if ( not path in dirList ) and ( not path in fileList ):
+        failed[path] = 'Path not found'
```
chaen: A standard error message like "No such file or directory", or even better, a DError object would be perfect here :-)

chaen:
Please let me know when the changes are done, so I will review again, and once this is done and merged, I will put my changes on top

DataManagementSystem/DB/FileCatalogComponents/DirectoryTreeBase.py
atsareg: In line 453 the call is just to change the directory attribute which is OK if no recursive flag. Here the result is used also for contained files including the parent directory.

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
atsareg: Yes, why not.

DataManagementSystem/DB/FileCatalogComponents/WithFkAndPs/DirectoryClosure.py
atsareg: Those methods are untouched ( only _ prepended ) but moved to DirectoryTreeBase which DirectoryClosure inherits. If you need your own implementation, add it here

PR #2555, opened 2015-08-24
[v6r15] new dfc client rebase

chaen:
Implements all the methods of the FileCatalog on the client side, with doc and default args

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/WithFkAndPs/DirectoryClosure.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/WithFkAndPs/DirectoryClosure.py
@@ -5,6 +5,11 @@ from DIRAC.Core.Utilities.List import intListToString, stringListToString
 
 """ DIRAC FileCatalog component representing a directory tree with
     a closure table
+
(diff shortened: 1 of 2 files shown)


phicharp:
I have a general concern: some of these methods are not specified in the FileCatalog abstraction... This is not really nice as in principle one should not have to know what is the actual implementation. Code should not depend on extensions, right?
However the PR looks OK, but this is really a problem with file catalogs...

atsareg:
Do we really need rpc and url arguments in each method and not just in the constructor ? What is the use case ? Some catalogs will not be able to implement the same arguments.

chaen:
Philippe, we already have this problem with the metadata method... it' true that we should have a nicer solution. I will think of it.

Andrei, that's the only way to specify a longer timeout for a single call.

atsareg:
Timeout is fine. Do you need rpc and url arguments to specify the timeout ?

chaen:
sorry, I answered too fast. No, it's probably good enough in the constructor. I will correct

chaen:
modified according to your remark Andrei

PR #2558, opened 2015-08-26
[v6r13] Forbid pilot to execute in /tmp

atsareg:
Force executing pilot in the current working directory if not specified otherwise


phicharp:
OK

PR #2559, opened 2015-08-27
[v6r15] Move physical accounting the StorageElement

chaen:
This will allow to track all physical operations, and with the proper protocol. It also removes the "getFile" from the DataManager accounting


Resources/Storage/StorageElement.py
```diff
@@ -732,6 +753,95 @@ def __getattr__( self, name ):
...
+    oDataOperation.setValueByKey( 'Protocol', storageParameters.get( 'Protocol', 'unknown' ) )
+  
+    if not callRes['OK']:
+      # Everything failed
+      oDataOperation.setValueByKey( 'TransferTotal', len( lfns ) )
+      oDataOperation.setValueByKey( 'FinalStatus', 'Failed' )
+    else:
+      totalSize = 0
+
+      oDataOperation.setValueByKey( 'TransferTotal', len( callRes['Value']['Successful'] ) )
```
phicharp: Must be damned sure that all SE implementations will return the correct structure!

phicharp:
Review OK

chaen:
Made changes to take the review into account

PR #2560, opened 2015-08-27
[v6r14] Fix in the getDigest

chaen:
Fix itt from RequestDB, and slightly modify the informaion returned by Request.getDigest


phicharp:
Review OK

PR #2563, opened 2015-08-28
[v6r14] Bug fixes of various kinds

rajanandakumar:
Hopefully this fixes different problems -

1. Stupid bug (used jobID where it should have been job)
2. Job update does *not* want the jobManagement stuff defined. Else it crashes. However, the job retrieval explicitly needs it defined. So, a kind of hack has been put in. Could this be because of the way the ARC library has been installed? It does not happen in my vanilla development machine at RAL, where the ARC libraries are installed by the sys admin here, in standard paths.
3. The getCEStatus function returned the results for all vos in the given CE. So, we were submitting a lot fewer jobs than expected. This has been fixed to use the DIRAC bdii function, but may need to be better protected at a later date.

Hopefully this PR has been done against the correct version now.

Cheers,
Raja.


rajanandakumar:
Let me request a pause on this PR. Some of the modifications, especially to the getARCJob method may not be  needed, if the ARC libraries are correctly installed, from talking to the ARC developers.

rajanandakumar:
Different pr made for really solving the issues (hopefully)

PR #2574, opened 2015-09-03
Need to support file names using unicode

sposs:
This is a problem for me as my file names comes from a DB returning unicode. Also more generic.


chaen:
Review OK. 
I think that for this particular case, it would be better to make one PR dedicated to replace the '== types.StringType' with isinstance(toto, basestring)

fstagni:
v6r11...? :(

sposs:
Yes, I know, Andrei said he would integrate it in a more recent version.
I'm waiting for him to close this.

atsareg:
fixed in v6r14p3

PR #2579, opened 2015-09-04
[v6r14] JobWrapper default for outputData + other fixes

atsareg:
FIX: JobWrapper - avoid defaulting outputData to ['']
BUGFIX: dirac-populate-component-db - DIRACexit instead of return 
FIX: Matcher - use VO specific Operations helper


phicharp:
Review OK

PR #2584, opened 2015-09-08
Syntax is basically open() not os.open()

Andrew-McNab-UK:
Fixed syntax, using open()

This bug breaks dirac-wms-job-submit command unless you put the whole JDL on the command line!

### DIFF
--- a/Interfaces/API/Dirac.py
+++ b/Interfaces/API/Dirac.py
@@ -306,8 +306,8 @@ class Dirac( API ):
     if isinstance( job, basestring ):
       if os.path.exists( job ):
         self.log.verbose( 'Found job JDL file %s' % ( job ) )
-        fd = os.open( job, 'r' )
-        jdlAsString = fd.read()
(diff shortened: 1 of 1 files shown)


Interfaces/API/Dirac.py
```diff
@@ -306,8 +306,8 @@ def submitJob( self, job, mode = 'wms' ):
     if isinstance( job, basestring ):
       if os.path.exists( job ):
         self.log.verbose( 'Found job JDL file %s' % ( job ) )
-        fd = os.open( job, 'r' )
-        jdlAsString = fd.read()
+        f = open( job, 'r' )
```
chaen: pylint will complain if you name your variable 'f'. Make it at least 2 letters please

chaen:
Also, against which branch do you want to do it ? Integration is now for v6r16 :-)

atsareg:
Replaced by #2587

PR #2586, opened 2015-09-09

atsareg:


Resources/Catalog/FileCatalog.py
```diff
@@ -34,6 +34,15 @@ class FileCatalog( object ):
 
   write_methods += write_meta_methods
 
+  lfn_methods = ['exists', 'isLink', 'readLink', 'isFile', 'getFileMetadata', 'getReplicas',
```
chaen: lfn_methods = write_methodes + ro_methods before adding the meta ?

DataManagementSystem/DB/FileCatalogComponents/Utilities.py
```diff
@@ -60,9 +57,52 @@ def checkArgumentDict( path ):
...
+    return S_OK( ( urls, urlMap ) )
+  else:
+    return S_OK( urls )
+
+def checkCatalogArguments( f ):
+  """ Decorator to check arguments of FileCatalog calls in the clients
+  """
+  def processWithCheckingArguments(*args, **kwargs):
+
+    checkFlag = kwargs.pop( 'NoLFNChecking', True )
```
chaen: so if 'NoLFNChecking == True', we do the check ? Shouldn't it be the other way around ?

Resources/Catalog/FileCatalog.py
```diff
@@ -94,25 +103,33 @@ def w_execute( self, *parms, **kws ):
...
+      kws['NoLFNChecking'] = True
+      allLfns = fileInfo.keys()
+      parms1 = parms[1:]
+
     for catalogName, oCatalog, master in self.writeCatalogs:
 
-      # Skip if metadata related method on pure File Catalog
-      if self.call in FileCatalog.write_meta_methods and not catalogName in self.metaCatalogs:
+      # Skip if the method is not implemented in this catalog
+      if not oCatalog.hasCatalogMethod( self.call ):
```
chaen: Since we agreed that all clients will implement all methods, we can just use hasattr

Resources/Catalog/FileCatalogClient.py
```diff
@@ -10,6 +10,31 @@
 from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Base.Client import Client
 from DIRAC.ConfigurationSystem.Client.Helpers.Registry import getVOMSAttributeForGroup, getDNForUsername
+from DIRAC.DataManagementSystem.DB.FileCatalogComponents.Utilities import checkCatalogArguments
+
+# List of common File Catalog methods implemented by this client
+CATALOG_METHODS = [ 'exists', 'addFile', 'removeFile', 'isFile', 'getFileMetadata', 'getFileSize',
```
chaen: All the methods are implemented, so we can use hasattr

chaen:
Okay, when we agreed that catalogs would advertise their capabilities, I did not understood that. Maybe we should have discussed it more. I believe it is a very bad idea to get ride of the FileCatalogBase. In fact, I wanted to do the opposite, and enforce that every replica catalog inherits from this. Otherwise, this means that we have to know a priori in the code what catalog is going to be instantiated when doing FileCatalog(). The same way than StorageElement is generic, I think FileCatalog should follow the same. 
Now, what makes this extremely complicated is all these meta methods. I would be in favor of stripping down the FileCatalog to what it was, i.e. a caller for Replica Catalogs, and then "MetaCatalog" class or whatever you want to call it, for all the meta methods, that currently only the DFC has. We would have FileCatalogBase, MetaCatalogBase, the DFC client can inherit from both. 
I really believe we have to standardize the interface and make a clear distinction between both, otherwise it forces us to know a priori what catalog is going to be instantiated.

Resources/Catalog/FileCatalog.py
atsareg: Not necessarily, I would like to keep those independent. lfn_methods are those that have LFN standard form dictionary as its first argument. I do not want to distinguish meta and other methods any more.

DataManagementSystem/DB/FileCatalogComponents/Utilities.py
atsareg: You are right, corrected

Resources/Catalog/FileCatalog.py
atsareg: Well, this is not the case right now since the FIleCatalogClient is still inheriting Client which can pass unfound methods to the service. That's why I wanted an explicit list. Also this is a kind of documentation of the client mandate.

atsareg:
Well, the FileCatalogueBase is just documenting the catalog interface. It is useless otherwise. In Python we do not have a notion of the interface to be implemented as part of the language. If we keep it and all the clients will inherit it, all the clients will "implement" all the methods without any advantage.

Resources/Catalog/FileCatalog.py
chaen: In fact, you should be able now to remove that inheritance.

And the documentation should be in each method no ? We start to duplicate so many time these informations (in the FileCatalog read or write method list, in the FileCatalog.lfn_methods, in the FileCatalogClient  CATALOG_METHODS, in the SecurityManager...)

atsareg: OK, I agree to remove  CATALOG_METHODS and use hasattr instead. We still need to distinguish read/write methods and methods using LFNs as their first argument. The latter can be done, in principle if we use a keyword, e.g. LFNs, to pass this argument and then call all the methods with **kwargs only. This will be certainly cleaner but will need fixes in very many places where these methods are called

chaen: It will not be needed if we separate FileCatalog and MetaCatalog :-)

atsareg: No, some FileCatalog methods do not have LFNs argument neither, e.g. getUsers(), etc

chaen: Hum, but these fall in a special "management" category no ? I mean, ultimately they have nothing to do with file management

atsareg: Then we will need yet another FileCatalogManagementClient which will not simplify things

chaen: I don't think that for these we need a common interface, since they are management operation, very rare, and specific to a given catalog. I would just not interface them with the FileCatalog. When doing a management operation on a given catalog, we should use that client directly

atsareg: Well, addUser() is a quite common operation, I think, which you do want to apply on all the catalogs at the same time.

PR #2587, opened 2015-09-09
[v6r14] JobWrapper default for outputData + other fixes

atsareg:
FIX: JobWrapper - avoid defaulting outputData to ['']
BUGFIX: dirac-populate-component-db - DIRACexit instead of return
FIX: Matcher - use VO specific Operations helper
BUGFIX: Dirac - use open and not os.open

This PR replaces #2584

### DIFF
--- a/WorkloadManagementSystem/JobWrapper/JobWrapper.py
+++ b/WorkloadManagementSystem/JobWrapper/JobWrapper.py
@@ -697,8 +697,7 @@ class JobWrapper( object ):
       outputSandbox = [ outputSandbox ]
     if outputSandbox:
       self.log.verbose( 'OutputSandbox files are: %s' % ', '.join( outputSandbox ) )
-    outputData = []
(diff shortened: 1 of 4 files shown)


WorkloadManagementSystem/JobWrapper/JobWrapper.py
```diff
@@ -697,8 +697,8 @@ def processJobOutputs( self, arguments ):
       outputSandbox = [ outputSandbox ]
     if outputSandbox:
       self.log.verbose( 'OutputSandbox files are: %s' % ', '.join( outputSandbox ) )
-    outputData = self.jobArgs.get( 'OutputData', '' )
-    if isinstance( outputData, basestring ):
+    outputData = self.jobArgs.get( 'OutputData', [] )
+    if outputData and isinstance( outputData, basestring ):
```
chaen: Why do you test 'if outputData' ? I don't think it is necessary, is it ?

PR #2598, opened 2015-09-15
[v6r14] conditional lock in the DictCache for the __del__

chaen:
The property does not seem very __del__ friendly. 
An exception is *sometimes* thrown, however it is ignored and transformed into warning. No consequences, just printout.
This should fix it

### DIFF
--- a/Core/Utilities/DictCache.py
+++ b/Core/Utilities/DictCache.py
@@ -158,11 +158,14 @@ class DictCache( object ):
     finally:
       self.lock.release()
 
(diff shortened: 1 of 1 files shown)


atsareg:
Well, actually without this fix, the files cached were not cleaned at the DictCache object destructing. Now this should be OK

chaen:
You can imagine that they were in my tests, otherwise I would not have done the PR. This __del__ method is really a poison :-(

PR #2599, opened 2015-09-15
[v6r12] Add agent to restart the RequestExecutingAgent when it gets stuck

andresailer:
Just a little agent that restarts the RequestExecutingAgent(s) if the log file is too old.
Workaround for #1901

### DIFF
--- a/RequestManagementSystem/Agent/RestartReqExeAgent.py
+++ b/RequestManagementSystem/Agent/RestartReqExeAgent.py
@@ -1,0 +1,111 @@
+""" :mod: RestartReqExeAgent
+    =====================
+
+    .. module: RestartReqExeAgent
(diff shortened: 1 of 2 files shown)


chaen:
Hi,

wouldn't you rather consider upgrading to (at least) v6r13 ? We (LHCb) didn't see it happening anymore.

atsareg:
This is rather an overkill. I guess the next step will be another agent which restarts the small agent which restarts the RequestExecutingAgent. First, consider the update where quite some effort was put into the reliability of the RequestExecutingAgent. Second, if you still need some temporary solution, make this agent part of the ILCDIRAC extension.

andresailer:
I can't upgrade to v6r13 as long as the InputDataResolution doesn't work for my UserJobs.

Ok, will then move this to ILCDIRAC

PR #2600, opened 2015-09-15
[v6r12]  Make systematic use of getPath to avoid '//' in path names

calancha:
Some functions receiving a path as argument were not normalizing
this path, which may cause some functions fail.
Normalizing all paths using the member function
getPath makes code more readable.

For instance following function call fail because path ends with '/':

```
meta set /ilc/user/c/calanchac/stdhep/E500/ Energy 500

Error: Server error while serving setMetadata: '/ilc/user/c/calanchac/stdhep/E500/'
```

### DIFF
--- a/DataManagementSystem/Client/FileCatalogClientCLI.py
+++ b/DataManagementSystem/Client/FileCatalogClientCLI.py
@@ -2074,7 +2074,7 @@ File Catalog Client $Revision: 1.17 $Date:
 
       listToPrint = None
       if dirsOnly:
-        listToPrint = set( self.getPath(fullpath) for fullpath in result['Value'] )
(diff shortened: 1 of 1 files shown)


DataManagementSystem/Client/FileCatalogClientCLI.py
```diff
@@ -2076,7 +2074,7 @@ def do_find(self,args):
 
       listToPrint = None
       if dirsOnly:
-        listToPrint = set( "/".join(fullpath.split("/")[:-1]) for fullpath in result['Value'] )
+        listToPrint = set( self.getPath(fullpath) for fullpath in result['Value'] )
```
andresailer: This is not doing the same thing. the join/split removes the last part of the LFN to leave just the folder.
If you want to change this line replace things with os.path.dirname

PR #2601, opened 2015-09-15
[v6r15] Remove import * from the general __init__.py

atsareg:


chaen:
Using some crafted greps, I could see that in DIRAC, these are the direct imports from DIRAC:
DErrno
DError
dexit
DIRACexit
DIRACExit
exit
gConfig
gLogger
gMonitor
List
rootPath
S_ERROR
siteName
S_OK

The DError and DErrno I will get ride myself. The rest is okay ? 

Btw, maybe a future step would be to review the need of custom modules such as List, or Os, or Time ?

atsareg:
For the first glance:
      
      On 15/09/15 16:38, chaen wrote:
    
    
      Using some crafted greps, I could see that in DIRAC, these are
        the direct imports from DIRAC:
        DErrno
        DError
        dexit    OK
        DIRACexit OK
        DIRACExit OK
        exit OK
        gConfig OK
        gLogger OK
        gMonitor  must be fixed
        List  must be fixed
        rootPath OK
        S_ERROR OK
        siteName OK
        S_OK OK
      
      The DError and DErrno I will get ride myself. The rest is okay
        ? 
      Btw, maybe a future step would be to review the need of custom
        modules such as List, or Os, or Time ?
      —
      
    
    Thanks for this grep result. Of course reviewing is always welcome,
    those modules are dating many years
    already.
    
      Cheers,
      Andrei
    
      Reply
        to this email directly or view
          it on GitHub.

chaen:
Here is the command btw:
grep -R 'from DIRAC import' | awk -F 'from DIRAC import' {'print $2'} | tr ' ' '\n' | sed 's/ +//g' | sed 's/,//g' | sort -u

Once you are happy with the left list, I guess review OK

PR #2603, opened 2015-09-16
!!! [v6r12] Set/Remove >1 metadata field each time

calancha:
Set/Remove metadata fields for a large bunch of files/directories using the dirac file catalog is cumbersome: only one metadata field can be set/remove each time.
This commit allow set/remove unlimited number of metadata fields each time.


atsareg:
This PR has conflicts with your previous PR, please fix

PR #2609, opened 2015-09-17
[v6r14] Bugfixes for GFAL2 plugins

fibbo:
### GFAL2_StorageBase.py
* Changed listDirectory so it works with the extra parameter internalCall
* Fixed protocol for getFile and putFile (from file: to file://)

### GFAL2_SRM2Storage.py
* Setting default protocol to gsiftp because with the default protocols from the CS there are authentication errors
* Added more comments
* Use of super() calls
* Stop question user.replicas while querying metadata because certain endpoints timeout if the file does not exist

### GFAL2_XROOTStorage.py
* Remove checksum check for getFile with XROOT because of permission problems with xroot
* Use of super() calls


Resources/Storage/GFAL2_SRM2Storage.py
```diff
@@ -38,7 +38,13 @@ def __init__( self, storageName, parameters ):
     self.gfal2.set_opt_integer( "SRM PLUGIN", "OPERATION_TIMEOUT", self.gfal2Timeout )
```
chaen: Rather that rewriting the same, just call __setSRMOptionstoDefault

Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -434,14 +434,17 @@ def __getSingleFile( self, src_url, dest_file ):
     params.overwrite = True  # old gfal removed old file first, gfal2 can just overwrite it with this flag set to True
     params.src_spacetoken = self.spaceToken
 
-    params.checksum_check = True if self.checksumType else False
+    if disableChecksum:
```
chaen: params.checksum_check = bool(self.checksumType and not disableChecksum)
(bool casting is maybe not needed)

Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -710,6 +713,9 @@ def __getSingleMetadata( self, path ):
           metadataDict['Unavailable'] = 1
         else:
           metadataDict['Unavailable'] = 0
+        if attributeDict['user.status'] == 'ONLINE_AND_NEARLINE':
```
chaen: First of all this whole part can be very much factorized. Something like
user_status = attributeDict.get('user.status', '')
metadataDict['Cached'] = int('ONLINE' in user_status)
metadataDict['Migrated'] = int('NEARLINE' in user_status)
metadataDict['Lost'] = int(user_status == 'LOST')
metadataDict['Unavailable'] = int(user_status == 'UNAVAILABLE')

Second, all these status are returned by SRM. Does that also work for the other protocol ? The 4 entries in the dict must be present, but if these status (nearline, and so on) are srm specific, then they should be in the srm plugin, not the base class)

chaen:
Comments and explanations about the optional parameters you added are missing

Resources/Storage/GFAL2_StorageBase.py
fibbo: https://github.com/DIRACGrid/DIRAC/commit/9f7f49bdd7d357f72fea966802b2410e8ef418df#diff-b5b53b92cedb0af1928db29712fc47df

In this PR XROOT has this metadata added 'for compatibility with SRM Storage' (migrated is not there though). So for that reason I would keep it in there, unless the XROOT has it for another reason and we don't need it anymore.

chaen: Yes but what you have wont work, since for xroot it should return 'Cached' to 1, and 0 for the rest. Unless 'ONLINE' is also a status returned by gfal2 when querying xroot, which was my other question ?

fibbo: Ah you are right, there is no online status for xroot. I will move that part to the SRM module and extend the XROOT module with the hardcoded version?

chaen: yes, but until I have time to unify, please hardcode in xroot. Or just hardcode it in the base, and have it overwriten by srm

Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -289,7 +289,7 @@ def __putSingleFile( self, src_file, dest_url, sourceSize ):
           errStr = "GFAL2_StorageBase.__putFile: The local source file does not exist or is a directory"
           self.log.error( errStr, src_file )
           return S_ERROR( errStr )
-        src_url = 'file:%s' % src_file
+        src_url = 'file://%s' % os.path.abspath( src_file )
```
chaen: Btw, I just realized that depending where we get the source from, it might already contains 'file://' at the beginning (for example, if it is SRM giving it to us). Please do the test (everywhere obviously) before adding it

Resources/Storage/GFAL2_SRM2Storage.py
```diff
@@ -47,10 +47,13 @@ def __setSRMOptionsToDefault( self ):
...
     self.gfal2.set_opt_string( "SRM PLUGIN", "SPACETOKENDESC", self.spaceToken )
-    self.gfal2.set_opt_string_list( "SRM PLUGIN", "TURL_PROTOCOLS", self.defaultLocalProtocols )
+    self.gfal2.set_opt_integer( "SRM PLUGIN", "REQUEST_LIFETIME", self.gfal2requestLifetime )
+    # Setting the TURL protocol to gsiftp because with other protocols we have authorisation problems
+#    self.gfal2.set_opt_string_list( "SRM PLUGIN", "TURL_PROTOCOLS", self.defaultLocalProtocols )
+    self.gfal2.set_opt_string_list( "SRM PLUGIN", "TURL_PROTOCOLS", ['gsiftp'] )
 
 
-  def _getExtendedAttributes( self, path, protocols = False ):
+  def _getExtendedAttributes( self, path, protocols = False, attributes = None ):
```
chaen: The doc for 'attributes' is missing

Resources/Storage/GFAL2_SRM2Storage.py
```diff
@@ -62,11 +65,24 @@ def _getExtendedAttributes( self, path, protocols = False ):
     '''
     if protocols:
       self.gfal2.set_opt_string_list( "SRM PLUGIN", "TURL_PROTOCOLS", protocols )
-    res = GFAL2_StorageBase._getExtendedAttributes( self, path )
+    res = super( GFAL2_SRM2Storage, self )._getExtendedAttributes( path, attributes )
```
chaen: attributes is a named argument, so name it when you call the method

Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -692,24 +686,15 @@ def __getSingleMetadata( self, path ):
...
-          metadataDict['Lost'] = 1
-        else:
-          metadataDict['Lost'] = 0
-        if attributeDict['user.status'] == 'UNAVAILABLE':
-          metadataDict['Unavailable'] = 1
-        else:
-          metadataDict['Unavailable'] = 0
+    res = self._getExtendedAttributes( path )
+    # add extended attributes to the dict if available
+    if res['OK']:
```
chaen: attributeDict = res.get('Value', {})

chaen:
Review OK

PR #2610, opened 2015-09-17
[v6r12] HTCondorCE, Work In Progress

andresailer:
Here is an implementation for running jobs at HTCondorCEs.
I took some code from the remote_script/condorce, but had to adapt this, also the condor_submit needs a grid proxy, so LocalComputingElement does not work as far as I can tell.

Some things are still work in progress, mostly things marked with FIXME, but advice is required.

Pilot submission (and matching etc.) works.
Getting the pilot output works as long as the WMSAdministrator runs on the same machine, however the working directory is hardcoded at the moment, because I have no idea how to get that into the CE Instance when the WMSAdministrator comes calling. condor places the stdout/stderr files from the job in the folder where condor_submit is called, and I don't think know if there is a way to do that differently.

Another issue is the cleaning of the pilot wrapper and the output files from the working directory of the sitedirector. the condor_master takes care of actually submitting the jobs to the CEs, which comes with some delay from the condor_submit call, so the pilot wrapper script must no be removed directly after calling submit job.
So at some later point the pilot wrapper should be removed. Also at some point the out/err files should be cleaned up.

I haven't yet figured out how to get the Pilot_Reference parameter to the job.
There is
```
environment = CONDOR_JOBID=$(Cluster).$(Process)
```
in the sub file, but this doesn't seem to show up in the execution host's environment


chaen:
if it is still work in progress, was it supposed to be merged ?

andresailer:
It wasn't supposed to be merged, I just wanted some comments :(, there wasn't even a "review ok" post.
But none of these changes should affect anybody

atsareg:
This is a new independent development, nobody will suffer from this merge. But this is eventually a very useful development. In such cases I do not hesitate to merge to facilitate further testing.

PR #2615, opened 2015-09-23
[v6r14] Don't recreate systematicaly the FTS3 Context

chaen:
We now don't recreate the FTS3 Context at every submission or monitoring of a job, but keep it as object attribute. This leverages the memory leak of pycurl and nss


atsareg:
Review OK

PR #2617, opened 2015-09-23

atsareg:
Redone to replace #2586


Resources/Catalog/FileCatalog.py
```diff
@@ -1,13 +1,62 @@
...
 """
 
-import types, re
+import re
 
 from DIRAC  import gLogger, gConfig, S_OK, S_ERROR
 from DIRAC.ConfigurationSystem.Client.Helpers.Operations    import Operations
 from DIRAC.Core.Security.ProxyInfo                          import getVOfromProxyGroup
-from DIRAC.Resources.Utilities                              import checkArgumentFormat
+from DIRAC.DataManagementSystem.DB.FileCatalogComponents.Utilities import checkArgumentFormat
```
chaen: Why changing that ? It also applies to StorageElements no ?

Resources/Catalog/FileCatalog.py
```diff
@@ -94,26 +129,29 @@ def w_execute( self, *parms, **kws ):
...
     parms1 = parms[1:]
-    lfnsFlag = False
+
     for catalogName, oCatalog, master in self.writeCatalogs:
 
-      # Skip if metadata related method on pure File Catalog
-      if self.call in FileCatalog.write_meta_methods and not catalogName in self.metaCatalogs:
-        continue
+      # Skip if the method is not implemented in this catalog
+      if not oCatalog.hasCatalogMethod( self.call ):
```
chaen: we had agreed to use hasattr no ?

Resources/Catalog/FileCatalog.py
```diff
@@ -123,27 +161,27 @@ def w_execute( self, *parms, **kws ):
...
-          for lfn, message in res['Value']['Failed'].items():
-            # Save the error message for the failed operations
-            failed.setdefault( lfn, {} )[catalogName] = message
-            if master:
-              # If this is the master catalog then we should not attempt the operation on other catalogs
-              fileInfo.pop( lfn, None )
-          for lfn, result in res['Value']['Successful'].items():
-            # Save the result return for each file for the successful operations
-            successful.setdefault( lfn, {} )[catalogName] = result
+      for lfn, message in res['Value']['Failed'].items():
```
chaen: This will make an exception if the failure is on a non master catalog. You don't have res['Value'] then. 
You need to keep the 'else' clause that was here before

Resources/Catalog/FileCatalogClient.py
```diff
@@ -8,29 +8,37 @@
...
+    self.server = 'DataManagement/FileCatalog'
     if url:
-      self.setServer( url )
+      self.server = url
     self.available = False
-#    res = self.isOK()
-#    if res['OK']:
-#      self.available = res['Value']
+    
+  def __getRPC( self, rpc = None, url = '', timeout = 600 ):
```
chaen: but this is basically reimplementing the client class no ?

Resources/Catalog/FileCatalog.py
atsareg: We have essentially the same thing in 2 different places. This one is more complete. I am not sure that this is the best place to keep this utility but it should rather be in the DataManagement

Resources/Catalog/FileCatalog.py
atsareg: In the current clients it is implemented as hasattr, but in other potential clients it can be different.

Resources/Catalog/FileCatalog.py
atsareg: Yes, this should be fixed

Resources/Catalog/FileCatalogClient.py
atsareg: Yes, but without __getattr__ which will return something on ANY method name

Resources/Catalog/FileCatalog.py
chaen: But it should not be since we agreed to implement on the client side all the methods as a general rule

Resources/Catalog/FileCatalog.py
chaen: DataManagement/Utilities ?

atsareg: Yes, may be

Resources/Catalog/FileCatalogClient.py
chaen: Then I would suggest to update the Client constructor with a boolean (default True). Depending on the boolean, the __getattr__ is active or not. This avoids re implementing the __getRPC in all the clients all over DIRAC ?

chaen:
any progress on that ?

PR #2620, opened 2015-09-25
[v6r15] fix importing Time from DIRAC

pigay:
`from DIRAC import Time` was failing in JobWrapper


chaen:
It was decided in v6r15 to strip down the global __init__, so you have to fix the code doing 'from DIRAC import Time' to actually import it from the appropriate location

pigay:
OK. So I close this PR and wil propose the suggested fix in another one.

Thanks.

PR #2622, opened 2015-09-25
[v6r14] Fixed indentation and a debug log typo

fibbo:
Fixed Indentation: Updating metadata dict with attributeDict only works if there is an attributeDict


chaen:
review ok

PR #2623, opened 2015-09-25
fix TaskQueueDirector utilities import

pigay:
`from DIRAC import Time, List` failed.


pigay:
Bad branch... I'll reopen another one

PR #2624, opened 2015-09-25
[v6r15] fix TaskQueueDirector utilities import

pigay:
`from DIRAC import Time, List` failed


atsareg:
Review OK

chaen:
Since we anyway want to get ride of this useless module, may I recommande that we start replacing them when easy to do so ? In this module, it should be quite straight forward to use the standard python modules to do the same job

atsareg:
Well, the TaskQueueDirector itself is not used almost any more. As for the Time and List modules replacement, I do not see any urgency preferring not to change what is working.

PR #2625, opened 2015-09-25
[v6r14] WMS bug fixes

atsareg:
The use of 0o22 form of integer is not understood by 2.6 python interpreters causing pilots failure on some sites.
A logical bug was introduced when the Matcher client class was created resulting in that only the first job was assigned to the pilot which ran multiple jobs
And yes, the VOMS is the evil - need extra protection against endecodabe attributes

### DIFF
--- a/Core/Security/X509Certificate.py
+++ b/Core/Security/X509Certificate.py
@@ -202,8 +202,11 @@ class X509Certificate:
           if extBundle[0] == "VOMS attribute":
             attr = GSI.crypto.asn1_loads( str(extBundle[1]) ).get_value()
             attr = attr[0][0][1][0]
-            data[ 'attribute' ] = "%s = %s (%s)" % attr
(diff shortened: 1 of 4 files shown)


atsareg:
Merged without review as it is urgent and used in hot fixes in several installations

chaen:
About the octal notation I am curious. Do you know exactly which python version does not work ? 

guests-MacBook-Air:~ chris$ python2.6
Python 2.6.9 (unknown, Jul 14 2015, 19:46:31)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
>>> 0o22
18
>>> 022
18

atsareg:
I think some earlier 2.6, I do not know for sure because this is the case on just a couple of sites, may be the native python is even older there. In fact 0o22 is supposed to be the standard starting from 3.0.

PR #2632, opened 2015-10-07
[v6r12] owner/group/mode changes with recursion

calancha:
Declared the type (boolean) of the recursion parameter before exporting
following functions: changePathOwner, changePathGroup, changePathMode.
Also updated their documentation strings.


atsareg:
The recursive changes are not that trivial, this is done properly in v6r15 being prepared. I Close this PR and #2633

PR #2633, opened 2015-10-07
[v6r12] Recursive LFNs owner/group/mode changes

calancha:
Recursion wasnt implemented on _changePathFunction (argument recursive was ignored).
This commit and PR #2632 hopefully make possible to call: chown, chgrp and chmod
passing the argument '-R' in the file catalog client.

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
@@ -733,7 +733,7 @@ class FileManagerBase( object ):
   
 
   def setFileStatus( self, lfns, connection = False ):
-    """ Get set the group for the supplied files """
+    """ Set the status for the supplied files """
     connection = self._getConnection( connection )
     res = self._findFiles( lfns, ['FileID', 'UID'], connection = connection )
     if not res['OK']:
@@ -753,7 +753,7 @@ class FileManagerBase( object ):
 
 
   def _setFileStatus( self, fileID, status, connection = False ):
-    """ Set the file owner """
+    """ Set the file status """
     connection = self._getConnection( connection )
     if type( status ) in StringTypes:
       if not status in self.db.validFileStatus:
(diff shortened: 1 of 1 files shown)


chaen:
Hi,

concerning this PR as well as this one https://github.com/DIRACGrid/DIRAC/pull/2632

A proper recursive behavior has been implemented for v6r15 which is to be the next release. Rather than submitting patch (very much appreciated btw :-) ) to this very old version, could you rather consider upgrading to a more recent one ? 

Cheers, and thanks

andresailer:
Hi Chris,
Tino is one of my ILCDIRAC power-users, so no we can't update (InputDataResolution, xroot storage). :(
Sorry

calancha:
Hi, thank you for the information.
I am glad to hear that news. As Andre said i am just an user so i cannot include anything myself.
Hopefully i will enjoy such new version in the future.

chaen:
I see... then I let @atsareg look at it, but I would very strongly suggest to run the unit/integration tests on them before merging. Also, if it gets merged, I would avoid to propagate it to further versions.

DataManagementSystem/DB/FileCatalogComponents/FileManagerBase.py
```diff
@@ -1401,6 +1401,15 @@ def _changePathFunction( self, paths, credDict, change_function_directory, chang
     for p in result['Value']['Successful']:
       if result['Value']['Successful'][p]:
         dirList.append( p )
+        if recursive:
+           result = self.db.dtree.getSubdirectories(p)
```
andresailer: Is this a three space indent?

calancha: Upps, yes, i used 3 spaces instead of 2 in that if construct. Sorry for that.

andresailer:
I just tested this change, but it doesn't work. 
I would propose to not merge this and wait for the later version.

calancha:
Ok, thank you for having tested.

PR #2634, opened 2015-10-07
!!! [v6r16] Next step for multi protocol support

chaen:
This PR contains the next steps in order to implement multi protocol support for storage elements. It is however far from final now. Some cleanup are still needed, and will be done once the gfal2 plugins are production ready. 

Basically, when performing an action on the StorageElement, instead of looping over all the protocol plugins, we loop over a filtered list. This list is built taking into account which action is taken (read vs write), and is also sorted according to lists defined in the CS. 
Two lists were already existing: RegistrationProtocols and ThirdPartyProtocols. They were moved into the Operation section where they belong to, and two more lists added: AccessProtocols and WriteProtocols. They all default to 'srm' and 'dips'. The utilities to fetch these lists are moved to DMSHelper, and each of them can be locally overwritten in the SE definition.

The negotiation for third party transfer is also improved: it takes into account all possible protocols the source SE is able to produce, and all protocols the target is able to receive as input.

Note that as of now, the behavior should not change, because the most used plugin (SRM2) still internally uses the DefaultProtocol list defined in the StorageElement section of the CS. This sort of bypasses  all the logic. But as I said, once gfal2 is production ready, we can do the final cleanup.

This will have to be veeeeery carefully tested :-)

### DIFF
--- a/ConfigurationSystem/Client/Helpers/Resources.py
+++ b/ConfigurationSystem/Client/Helpers/Resources.py
@@ -288,10 +288,19 @@ def getCatalogPath( catalogName ):
   """
   return '/Resources/FileCatalogs/%s' % catalogName
 
-def getRegistrationProtocols():
-  """ Returns the Favorite registration protocol defined in the CS, or 'srm' as default """
-  return gConfig.getValue( '/Resources/FileCatalogs/RegistrationProtocols', ['srm', 'dips'] )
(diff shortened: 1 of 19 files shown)


atsareg:
This PR has conflicts that must be resolved. Would be nice to resolve them before leaving on vacations, Chris

PR #2640, opened 2015-10-09
[v6r14] fix bulk requests

chaen:
Lock the lines when selecting the requests to be assigned, update the LastUpdate time, and expose the assigned flag to the client


phicharp:
Review OK

PR #2641, opened 2015-10-09
[v6r14] Time Left utility revisited

phicharp:
I reviewed the TimeLeft utilities included the plugins:
* Make them consistent in terms of what is returned: all return real seconds
* Checked with concrete examples -> fix LSFTimeLeft not taking correctly normalization
* Use consistently always the same CPU power (not yet changed the names, i.e. normFactor and scaleFactor)
* Review JobAgent as well for what concerns time left

### DIFF
--- a/Core/Utilities/TimeLeft/BQSTimeLeft.py
+++ b/Core/Utilities/TimeLeft/BQSTimeLeft.py
@@ -6,7 +6,7 @@
     current CPU consumed and CPU limit.
 """
 
-from DIRAC import gLogger, gConfig, S_OK, S_ERROR
+from DIRAC import gLogger, gConfig, S_OK, S_ERROR, siteName
 from DIRAC.Core.Utilities.TimeLeft.TimeLeft import runCommand
 
 __RCSID__ = "$Id$"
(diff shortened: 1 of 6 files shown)


atsareg:
BQSTimeLeft is a legacy class ( no more BQS ). I wonder how the changes were tested, this is not trivial for all the cases.

phicharp:
Can we remove BQSTimeLeft? It is for sure not used as it had a bug that would have triggered an exception (undefined variable).
The other changes are mostly a rearrangement of the code in order to be easier to maintain... The only functional change is the one in LSFTimeLeft as there was a real bug there: the baseline now is that the plugins return times in real seconds rather than in a unit system that was unknown to the TimeLeft module (this was the bug in LSF). 
I agree these changes may go to a quick certification run, although the current situation is that it just doesn't work at most sites, therefore even if there is a bug it won't change much ;-). It is impossible to "test" these changes except submitting jobs... We already know it cannot work at GRIDKA as SGE there is not configured in the expected way and the information cannot be retrieved (number of jobs set to one at that site).

atsareg:
Ok, if it fixes the bug that makes it unusable anyway, I will merge it and I will remove the BQS class. I think you are understand well that even a "mostly a rearrangement of the code" is error prone, however.

phicharp:
Yes I know Andrei! ;-)
What we discussed this morning is that anyway Joel will install it on the certification system as he always does and we shall submit some jobs and watch the pilot log at CERN, GRIDKA and IN2P3 (for LSF and SGE)

PR #2651, opened 2015-10-15
[v6r14] Documentation fixes

andresailer:
Just some things I stumbled across while compiling my own (ILCDIRAC) documentation
This should go to whatever release the documentation is compiled from, please let me know if it is another branch...

Fixes for this:
http://diracgrid.org/files/docs/CodeDocumentation/Core/Utilities/ReturnValues.html
And this
http://diracgrid.org/files/docs/CodeDocumentation/Interfaces/API/Job.html#DIRAC.Interfaces.API.Job.Job.setExecutable

### DIFF
--- a/Core/Utilities/ReturnValues.py
+++ b/Core/Utilities/ReturnValues.py
@@ -1,19 +1,27 @@
 # $HeadURL$
 """
    DIRAC return dictionary
+
    Message values are converted to string
+
    keys are converted to string
 """
 
 import types
 
 def S_ERROR( messageString = '' ):
-  """ return value on error confition
+  """ return value on error condition
+
   :param string messageString: error description
   """
   return { 'OK' : False, 'Message' : str( messageString )  }
 
 def S_OK( value = None ):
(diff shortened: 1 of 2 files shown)


chaen:
In fact this is something I already wanted to point out. We should have for each release of DIRAC a corresponding release of DIRACDocs and TestDIRAC, because currently things are inconsistent. Also, it would be nice if the diracgrid website could expose the versioning of the documentation (it must be feasible since SQLAlchemy is doing it using sphynx : http://docs.sqlalchemy.org/en/rel_1_0/ )

Review OK btw :-)

andresailer:
Yes, it would be nice to have at least the version mentioned on the website.
Even better would be to be able to chose the version to be displayed, like in the python std lib documentation, which is also based on sphinx. Maybe somebody should open a feature request in DIRACDocs :)

chaen:
Somebody opened a feature request in DIRACDocs https://github.com/DIRACGrid/DIRACDocs/issues/64 :-)

PR #2665, opened 2015-10-21
[v6r12] fix: cleanDirectory fails for empty directories

andresailer:
cleanDirectory (e.g., called from dirac-dms-clean-directory) fails when the directory itself is empty,
```
Failed to clean directory. /ilc/prod/ilc/mc-dbd/ild/dst/250-TDR_ws/higgs_ffh/ILD_o1_v05/v01-16-p10_250/00005972/000 Failed to perform exists from any catalog
```
 because the function passes an empty list to removeFile here https://github.com/andresailer/DIRAC/blob/de7cc06e8d86f8ff0787e315faf76824688413ab/DataManagementSystem/Client/DataManager.py#L128
, removeFile then passes an empty list to exists which returns an error.

Now there are three locations to fix that: 
 * in cleanDirectory not passing an empty list
 * in removeFile returning success for empty input 
 * in exists to also return success in case of empty input (I guess, did not actually try that...)

I chose the second option here because it has the smallest footprint of changes as far as I can tell

### DIFF
--- a/DataManagementSystem/Client/DataManager.py
+++ b/DataManagementSystem/Client/DataManager.py
@@ -1007,6 +1007,8 @@ class DataManager( object ):
 
         'lfn' is the file to be removed
     """
+    if not lfn:
+      return S_OK( { 'Successful': {}, 'Failed': {} } )
     if force == None:
       force = self.ignoreMissingInFC
     if type( lfn ) == ListType:


chaen:
But I don't understand how removeFile can be called with an empty list. __getCatalogDirectoryContents is supposed to return in successful something like { 'emptyDir' : { 'Files' :{}, 'SubDirs' : {} } }, isn't it ?

andresailer:
you mean this function? https://github.com/DIRACGrid/DIRAC/blob/rel-v6r14/DataManagementSystem/Client/DataManager.py#L221
This function just returns S_OK( allFiles ) where allFiles is a dict (same in r12 or r14) so if the directory is empty allFiles is empty?

chaen:
Oh right... sorry I looked too quickly.
Review OK then

andresailer:
Ok, thanks

PR #2667, opened 2015-10-21
[v6r14] - improve the retry mechanism

zmathe:


Core/DISET/private/BaseClient.py
```diff
@@ -205,7 +205,7 @@ def __findServiceURL( self ):
       
     if len( urls ) == len( self.__bannedUrls ):
       self.__bannedUrls = []  # retry all urls
-      self.__retryDelay = 5 / len ( urls )  # we run only one service! In that case we increase the retry delay.
+      self.__retryDelay = 5 / len ( urls ) if len( urls ) > 1 else 1  # we run only one service! In that case we increase the retry delay.
```
atsareg: If one service - delay == 1, if 2 services - delay == 2.5. Is it what you wanted ?

Core/DISET/private/Transports/SSL/SocketInfoFactory.py
```diff
@@ -57,7 +57,7 @@ def __sockConnect( self, hostAddress, sockType, timeout, retries ):
     osSocket = socket.socket( sockType, socket.SOCK_STREAM )
     #osSocket.setblocking( 0 )
     if timeout:
-      osSocket.settimeout( 5 )
+      osSocket.settimeout( 0.5 ) # we retry 3 times, which is 3 sec + 5 times retry = 15 sec.
```
atsareg: Will 0.5 be OK for slow networks ? The total period of retries looks OK ( 15 sec ), this is in most cases enough for a service to restart. But may be it is better to have a slightly longer timeout but less number of retries.

Core/DISET/private/BaseClient.py
zmathe: no, What I want is to have maximum 2.5 second delay. This is the case when we run 2 service. But when we run more service I want the numbers to be decreased. For example: 5/3=1.66, 5/4=1.25
So, If we have more service, it make no sense to wait to much, because we have to retry very fast all the services.
I want to keep a constant number (which is 1) when 1 service running.

atsareg: Why is the shortest delay in case of 1 service ? Aren't you contradicting yourself when you say the more the number of services the less delay ?

Core/DISET/private/Transports/SSL/SocketInfoFactory.py
zmathe: 15 sec is not true any more. When we have ipv6 enabled we have a big problem... That is why I decreased the timeout. What I discovered is we have a list of addresses (ipv6 and ip4). ipv6 always  timeout. So 3 times 5 is 15 sec. For example:
2015-10-21 15:51:54 UTC tt[MG1M]  DEBUG: Connecting to: dips://lbtestvobox2.cern.ch:9167/Bookkeeping/TestBookkeepingManager
2015-10-21 15:51:54 UTC tt[MG1M]  DEBUG: Using proxy /tmp/x509up_u29854
retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.500597953796
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.500988006592
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.501138925552
retries 2 ('128.142.140.250', 9167) 2
2015-10-21 15:51:55 UTC tt[MG1M]  DEBUG:  Authenticated peer (/DC=ch/DC=cern/OU=computers/CN=lbtestvobox2.cern.ch)
2015-10-21 15:51:55 UTC tt[MG1M]  DEBUG: New session connecting to server at ('lbtestvobox2.cern.ch', 9167)
2015-10-21 15:51:55 UTC tt[MG1M]  DEBUG: New connection -> 128.142.140.250:9167
2015-10-21 15:51:56 UTC tt[MG1M]  DEBUG: Closing socket
{'OK': True, 'Value': [1825423375817125]}
Elapsed time: 2.51006507874

zmathe: I am going to test the Bookkeeping service and I will generate high workload. So, I can see how this settings will behaves.

atsareg:
We have 2 reasons to retry: 1. URL unavailable, definitively down 2. Service is restarting for whatever good reason. In case of 1. it is better to go to the next URL as soon as reasonable, that's why you decrease the delay, fine, but I wonder if 0.5 not too small as in slow networks we can have round trip time in hundreds of millisecs. So, 0.5 can be just not enough even if the service is up and running. In case of 2. we have to give the service enough time to restart where the total cumulated delay ( when all retries are used ) of 15 secs looks reasonable to me. Therefore, socket timeout of 1 sec and let say 3 times retry of all the URLs looks more adequate.

Core/DISET/private/BaseClient.py
zmathe: Okay, I increased to 3. In that case we will have 5 retry and each connection will be 3 retry. In that case we will wait 13 second.
But when ipv6 enabled we wait 5 * 3 * (3*0.5) = 22.5

rajanandakumar:
Why do you say that having ipv6 enabled is a big problem? I am interested ...

zmathe:
@rajanandakumar Because it does not work at all. For example:
2015-10-21 16:13:30 UTC tt[NGW8]  DEBUG: Connecting to: dips://lbtestvobox2.cern.ch:9167/Bookkeeping/TestBookkeepingManager
2015-10-21 16:13:30 UTC tt[NGW8]  DEBUG: Using proxy /tmp/x509up_u29854
retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.500771999359
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.501883029938
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.500822067261
retries 2 ('128.142.140.250', 9167) 2
2015-10-21 16:13:32 UTC tt[NGW8]  DEBUG:  Authenticated peer (/DC=ch/DC=cern/OU=computers/CN=lbtestvobox2.cern.ch)
2015-10-21 16:13:32 UTC tt[NGW8]  DEBUG: New session connecting to server at ('lbtestvobox2.cern.ch', 9167)
2015-10-21 16:13:32 UTC tt[NGW8]  DEBUG: New connection -> 128.142.140.250:9167

The interesting line: retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
As you can see 3 times timeout: EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.500771999359
And at the and it will work using ipv4.
I am using lxplus. I do not know why?

rajanandakumar:
Isnt this is exactly as expected?

zmathe:
@atsareg Yes, I think 15 sec is enough.  0.5 second maybe not enough but I am not sure. With the current settings. We have 3 minutes retry delay and 0.5 second timeout. The services up and running:
[zmathe@lxplus0009 ~]$ python tt.py
retries 2 ('128.142.152.39', 9135) 2
retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.501140117645
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.501131057739
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
0.501550912857
retries 2 ('128.142.140.250', 9167) 2
{'OK': True, 'Value': [1825423375817125]}
Elapsed time: 2.2272810936
[zmathe@lxplus0009 ~]$
As you can see we lost 0.5 + 3 sec because of ipv6. If the first address is ipv4 the same query:
[zmathe@lxplus0009 ~]$ python tt.py
retries 2 ('128.142.152.39', 9135) 2
retries 2 ('128.142.140.250', 9167) 2
{'OK': True, 'Value': [1825423375817125]}
Elapsed time: 0.726531028748
[zmathe@lxplus0009 ~]$ python tt.py

If I use 5 second timeout:
[zmathe@lxplus0009 ~]$ python tt.py
retries 2 ('128.142.152.39', 9135) 2
retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
5.00511312485
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
5.00050997734
EX: timed out ('2001:1458:301:bc::100:ec', 9167, 0, 0)
5.00482296944
retries 2 ('128.142.140.250', 9167) 2
{'OK': True, 'Value': [1825423375817125]}
Elapsed time: 15.7243411541
[zmathe@lxplus0009 ~]$

This is the current situation. I do not know why ipv6 timeout.

rajanandakumar:
Okay - I think I understand your problem. Currently the system is as follows - The connection is tried for all the addresses that are available for the destination server. If the client is a pure-ipv4 machine, connections to an ipv6 address will fail and similarly for a pure-ipv6 client.

Presumably a solution could be to determine in this module, whether the client is a pure-ipv4 (or pure-ipv6) machine and connect only to the destination addresses of the corresponding type. That will eliminate the error you are worried about.

I am happy to help, but am not sure where to get started on that however - any ideas appreciated. But to re-iterate, connecting to a ipv6 address from an ip4 client is certainly expected to fail. This is not a surprise and should not be considered as a disaster here.

zmathe:
@rajanandakumar It means my local machine works correctly. I am using the same code:
[zmathe@pclhcb63 LHCbDirac_v8r2p7]$ python ~/tt.py
retries 2 ('128.142.152.39', 9135) 2
retries 2 ('2001:1458:301:bc::100:ec', 9167, 0, 0) 10
retries 2 ('128.142.140.250', 9167) 2
{'OK': True, 'Value': [1825423375817125]}
Elapsed time: 0.75167298317
[zmathe@pclhcb63 LHCbDirac_v8r2p7]$
This is because the "Address family not supported by protocol" error message. In that case I do not try to connect to the service.

rajanandakumar:
Could you give the output of the command "ip addr" for your machine?

zmathe:
[zmathe@pclhcb63 ~]$ /sbin/ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 64:31:50:2c:c1:6d brd ff:ff:ff:ff:ff:ff
    inet 137.138.188.94/26 brd 137.138.188.127 scope global eth0
@rajanandakumar  My ip address:
3: virbr0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN
    link/ether 52:54:00:37:fc:a0 brd ff:ff:ff:ff:ff:ff
    inet 192.168.122.1/24 brd 192.168.122.255 scope global virbr0
4: virbr0-nic: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN qlen 500
    link/ether 52:54:00:37:fc:a0 brd ff:ff:ff:ff:ff:ff
[zmathe@pclhcb63 ~]$

zmathe:
I propose to separate the two issues as we are not using ipv6 in production. The issues are:
1.) make the retry mechanism faster (we must retry all urls much faster)
2.) improve the connection when ipv6 is used.

zmathe:
After a discussion with @atsareg  we decided to solve this issue two times. What we have in this patch release:
-reduced the socket timeout.
-I retry all urls 3 times if the number of urls < 2 otherwise 2 times.
-the first round will be 0 daly after that if we have single service we wait 5 seconds otherwise we wait 3/number of services
The result of this change:
*one service used and it is down the client has to wait <20 second
*more than one service and all of them down the client has to wait between 20 and 30 second.

atsareg:
Review OK

PR #2672, opened 2015-10-22
[v614] Update to add queue to the XRSL string

rajanandakumar:
This is to add the "queue" field for submission to ARC sites, which was discussed, but seems to have mysteriously vanished at some point. This has been tested in the certification instance by me.

The major assumption of this patch is that none of the batch systems we run on (PBS, GE, Condor, Torque, ... whatever) have a "-" in their name.

### DIFF
--- a/Resources/Computing/ARCComputingElement.py
+++ b/Resources/Computing/ARCComputingElement.py
@@ -206,6 +206,8 @@ class ARCComputingElement( ComputingElement ):
       jobdescs = arc.JobDescriptionList()
       # Get the job into the ARC way
       xrslString, diracStamp = self.__writeXRSL( executableFile )
+      gLogger.debug("XRSL string submitted : %S" %xrslString)
+      gLogger.debug("DIRAC stamp for job : %S" %diracStamp)
       if not arc.JobDescription_Parse(xrslString, jobdescs):
         gLogger.error("Invalid job description")
         break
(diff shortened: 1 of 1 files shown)


Resources/Computing/ARCComputingElement.py
```diff
@@ -180,6 +181,9 @@ def submitJob( self, executableFile, proxy, numberOfJobs = 1 ):
     """ Method to submit job
     """
 
+    # Assume that the ARC queues are always of the format nordugrid-<batchSystem>-<queue>
+    # And none of our supported batch systems have a "-" in their name
+    self.arcQueue = self.queue.split("-",2)[2]
```
rajanandakumar: Question for the experts : Does this have to be protected by a try-except?

chaen: I would say that unless you are 100% sure that you have at least two "-" in the name, you should use try/except yes. And if you are sure, then put in the comment :-)

rajanandakumar: Going over all the CEs in the configuration, this is 100% true. I have also asked the nordugrid guys, but they may take time to reply.

rajanandakumar: I mean the LHCb configuration is what I checked - I have access only to that ...

rajanandakumar: And response from the ARC developers below ... so, this is fine I suppose, unless the siteDirector gets the queue wrong (can it?).

============
Short answer:

I don't foresee this is going to be changed in the lifetime of these
systems, so you might go ahead and hardcode such naming structure.

atsareg:
Review OK

PR #2676, opened 2015-10-26
[v6r12] Fix job rescheduling on (glexec) error

sfayer:
We found that when a glexec wrapped job failed and rescheduleOnError was True, the rescheduling would trigger an exception. This patch fixes that problem and also returns S_OK() so that the outer layer of the JobAgent doesn't immediately mark the job as failed again.

Regards,
Simon


atsareg:
This is a real bug since long time, thanks. I will merge it into ealier releases.

PR #2679, opened 2015-10-27
[v6r12] Function returning trans. output lfns

calancha:
Current implementation cover the transformation input
quite well, but lacks of methods to check out easily the
transformation output.

Sometimes you want to list the output lfns in a
running production: you can easily do that in the dirac
file catalog but it's sensible to include such operation
also here.


andresailer:
Question to everybody else: Is "ProdID" always used as metadata for the production ID?

sposs:
Hi,
There are no constrains on this. It could be an Operations option to make it community dependent.

PR #2680, opened 2015-10-27
[v6r15] move missing wc/cpu time workaround fixes from batchPlugins to TimeLeft

pigay:
The batch plugins won't mix CPU and walltime values anymore, missing values are handled by the TimeLeft object.

This is the first part of the replacement of #2629

### DIFF
--- a/Core/Utilities/TimeLeft/MJFTimeLeft.py
+++ b/Core/Utilities/TimeLeft/MJFTimeLeft.py
@@ -78,7 +78,7 @@ class MJFTimeLeft:
     # We cannot get CPU usage from MJF
       
     consumed = {'CPU':cpu, 'CPULimit':cpuLimit, 'WallClock':wallClock, 'WallClockLimit':wallClockLimit}
-    self.log.debug( consumed )
+    self.log.debug( "MJF consumed: %s" % str( consumed ) )
 
     if cpuLimit and wallClock and wallClockLimit:
       return S_OK( consumed )
--- a/Core/Utilities/TimeLeft/MJFTimeLeft.py
+++ b/Core/Utilities/TimeLeft/MJFTimeLeft.py
@@ -73,16 +73,14 @@ class MJFTimeLeft:
       cpuLimit = int( urllib.urlopen(jobFeaturesPath + '/cpu_limit_secs').read() )
     except:
       self.log.warn( 'Could not determine cpu limit from $JOBFEATURES/cpu_limit_secs' )
-      cpuLimit = wallClockLimit
 
     wallClock = int(time.time()) - jobStartSecs
-    # We cannot get CPU usage from MJF, so for now use wallClock figure
-    cpu = wallClock
+    # We cannot get CPU usage from MJF
       
     consumed = {'CPU':cpu, 'CPULimit':cpuLimit, 'WallClock':wallClock, 'WallClockLimit':wallClockLimit}
     self.log.debug( consumed )
 
-    if cpu and cpuLimit and wallClock and wallClockLimit:
+    if cpuLimit and wallClock and wallClockLimit:
       return S_OK( consumed )
     else:
       self.log.info( 'Could not determine some parameters' )
--- a/Core/Utilities/TimeLeft/TimeLeft.py
+++ b/Core/Utilities/TimeLeft/TimeLeft.py
@@ -56,7 +56,16 @@ class TimeLeft( object ):
     if not self.batchPlugin:
       return 0
 
-    return self.batchPlugin.getResourceUsage().get( 'Value', {} ).get( 'CPU', 0.0 ) * self.scaleFactor
+    resourceDict = self.batchPlugin.getResourceUsage()
+
+    if 'Value' in resourceDict:
+      if resourceDict['Value']['CPU']:
+        return resourceDict['Value']['CPU'] * self.scaleFactor
+      elif resourceDict['Value']['WallClock']:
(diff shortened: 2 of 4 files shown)


Core/Utilities/TimeLeft/TimeLeft.py
```diff
@@ -76,10 +85,23 @@ def getTimeLeft( self, cpuConsumed = 0.0 ):
       return resourceDict
 
     resources = resourceDict['Value']
-    if not resources['CPULimit'] or not resources['WallClockLimit']:
+    self.log.verbose( resources )
```
phicharp: Please avoid this kind of logging without any context... It makes debugging impossible... Also when logging a full result dictionary it would be better to set it to "debug" level as this is really very low level ("debug" has a level lower than "verbose")

Core/Utilities/TimeLeft/TimeLeft.py
```diff
@@ -76,10 +85,23 @@ def getTimeLeft( self, cpuConsumed = 0.0 ):
...
+    if not resources['CPULimit'] and not resources['WallClockLimit']:
       # This should never happen
       return S_ERROR( 'No CPU or WallClock limit obtained' )
 
+    # if one of CPULimit or WallClockLimit is missing, compute a reasonable value
+    if not resources['CPULimit']:
+      resources['CPULimit'] = resources['WallClockLimit']
```
phicharp: This is not good: we should set a CPULimit lower than WallClock limit, although we don't know the ratio,  Iwould propose a factor 0.8 or 0.7
Similarly if wall clock limit is not set, we can set it to CPULimit / 0.8 in order to be only limited by CPULimit

Core/Utilities/TimeLeft/MJFTimeLeft.py
```diff
@@ -73,16 +73,14 @@ def getResourceUsage( self ):
...
     wallClock = int(time.time()) - jobStartSecs
-    # We cannot get CPU usage from MJF, so for now use wallClock figure
-    cpu = wallClock
+    # We cannot get CPU usage from MJF
       
     consumed = {'CPU':cpu, 'CPULimit':cpuLimit, 'WallClock':wallClock, 'WallClockLimit':wallClockLimit}
     self.log.debug( consumed )
 
-    if cpu and cpuLimit and wallClock and wallClockLimit:
+    if cpuLimit and wallClock and wallClockLimit:
```
phicharp: Why do you still keep this constraint?

Core/Utilities/TimeLeft/TimeLeft.py
pigay: fixed in 41f9fd8

Core/Utilities/TimeLeft/TimeLeft.py
pigay: This would change the behaviour of Timeleft utility I think. I didn't intend to do this with this PR.

phicharp: I agree, but we had recently an example where this was causing problems as the site had set CPULimit = 0.7*WallClockLimit but for a strange reason the CPULimit was not exposed :(
Jobs were killed by CPU limit then as it was reached before the WallClockLimit was reached... I agree it would be a hack rather than a real fix for a missing value as the factor would be random... An alternative is to just return an error if CPULimit is not provided by WallClockLimit is.

Core/Utilities/TimeLeft/MJFTimeLeft.py
```diff
@@ -73,16 +73,14 @@ def getResourceUsage( self ):
...
       self.log.warn( 'Could not determine cpu limit from $JOBFEATURES/cpu_limit_secs' )
-      cpuLimit = wallClockLimit
 
     wallClock = int(time.time()) - jobStartSecs
-    # We cannot get CPU usage from MJF, so for now use wallClock figure
-    cpu = wallClock
+    # We cannot get CPU usage from MJF
       
     consumed = {'CPU':cpu, 'CPULimit':cpuLimit, 'WallClock':wallClock, 'WallClockLimit':wallClockLimit}
     self.log.debug( consumed )
```
phicharp: Still I think you should add a context as first argument ;-)

Core/Utilities/TimeLeft/MJFTimeLeft.py
pigay: I want to keep the same behaviour as before. I removed the constraint on cpu because it's normal to have it missing (from the comment in the source). I left the other constraints that correspond to MJF errors because I can't test with MJF very carefully.

I realize we could remove the constraint on wallClock because it is unconditionnally computed at line 77. I'll add a commit if you agree.

Core/Utilities/TimeLeft/MJFTimeLeft.py
pigay: This one is not on my fault! You should blame 254cf51a instead ;)

More seriously, I fix in 7ccdefa

Core/Utilities/TimeLeft/TimeLeft.py
pigay: To be honest, I'd prefer if this problem could be fixed in a separate PR.

phicharp: OK fine... Indeed I had it real in a local branch here, but I will let you PR to be merged first otherwise we shall have conflicts and Andrei will not be happy...
At least the functionality is not changed... 
Review OK for me then!

PR #2688, opened 2015-10-30
v6r14 - fix gzip buffer size

roiser:
fix the buffer size for reading when zipping a file to maximum 2^31-1 bytes. Gzip internally does a crc32 check which will otherwise fail.


atsareg:
Review OK

PR #2690, opened 2015-10-30
v6r14 : extend the JobMonitoring Service 

roiser:
I want to propose to extend the JobMonitoring Service with this function, it would allow to retrieve multiple parameters for multiple jobs. Otherwise when going through a list of jobs and retrieving the same parameters for each of them it takes a very long time. This would allow to do it in one shot. 

My concrete example where I can use this is the monitoring of mesh processing, here I will need to retrieve multiple parameters, such as the JobGroup, RunNumber, Status  for a "helper site" to jobs it has done for a given job type in order to further digest which site it was contributing to for each job.

### DIFF
--- a/WorkloadManagementSystem/Service/JobMonitoringHandler.py
+++ b/WorkloadManagementSystem/Service/JobMonitoringHandler.py
@@ -242,6 +242,14 @@ class JobMonitoringHandler( RequestHandler ):
     return gJobLoggingDB.getJobLoggingInfo( jobID )
 
 ##############################################################################
+  types_getJobsParameters = [ ListType, ListType ]
+  @staticmethod
+  def export_getJobsParameters ( jobIDs, parameters ):
+    if not ( jobIDs and parameters ) : 
+      return S_OK( {} )
+    return gJobDB.getAttributesForJobList( jobIDs, parameters )
+
+##############################################################################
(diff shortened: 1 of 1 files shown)


WorkloadManagementSystem/Service/JobMonitoringHandler.py
```diff
@@ -242,6 +242,14 @@ def export_getJobLoggingInfo( jobID ):
...
 
 ##############################################################################
+  types_getJobsParameters = [ ListType, ListType ]
+  @staticmethod
+  def export_getJobsParameters ( jobIDs, parameters ):
+    if not ( jobIDs and parameters ) : 
+      return S_OK( {} )
+    return gJobDB.getAttributesForJobList( jobIDs, parameters )
+
+##############################################################################
```
atsareg: This is fine. Just one question - I think that getJobParameters() is better english, but I am not a native speaker. Ideal names in the code should read as normal language phrases

roiser: Up to you, I was following the spelling of the other functions, e.g. we have a function for getting the Status of multiple jobs which is called "getJobsStatus" and similar for other single parameters. 

On the other hand functions which only query one or multiple parameters of a job start with "getJob<parameter>". 

That was my reasoning for calling the function like that.

PR #2708, opened 2015-11-06
[v6r14] Move getCEStatus function back to using the ARC API.

rajanandakumar:
This is not an urgent patch and in principle, should give identical results to what already exists. Hopefully this code will be blind to what version of ldap or which glue schema is actually used to publish information in the future.

Here, we also avoid (unnecessary) knowledge of what VO this is, as that depends on the configuration which can be in a debatable state.

"Tested" in certification and by the Imperial College / GridPP instance.


atsareg:
Code review OK. Relying on the tests that you have done. But if this is not urgent ( I hope GridPP case is fixed elsewhere ), it should better go to v6r15

rajanandakumar:
Fine with me!

rajanandakumar:
Before I forget about this PR - is anything further required from me?

PR #2709, opened 2015-11-06
[v6r14] TransformationPlugin: Need to pass parameters to utils

andresailer:
As far as I can see the parameters for PluginUtilities are never set.
Without setting the params things like GroupSize are otherwise only taken from the default or the CS and not from individual transformations.

### DIFF
--- a/TransformationSystem/Agent/TransformationPlugin.py
+++ b/TransformationSystem/Agent/TransformationPlugin.py
@@ -47,6 +47,12 @@ class TransformationPlugin( PluginBase ):
       self.valid = False
     return self.valid
 
+  def setParameters( self, params ):
+    """ Need to pass parameters also to self.util
+    """
+    self.params = params
+    self.util.setParameters(params)
+
   def setInputData( self, data ):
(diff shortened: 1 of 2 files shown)


phicharp:
You are right, this was forgotten by whoever ported all this from LHCbDIRAC to DIRAC... We have exactly that method in TransformationPlugin of LHCbDIRAC...

phicharp:
There is also the following line that you may want to add:
    self.transID = params['TransformationID']

phicharp:
Review OK

andresailer:
the 
```python
self.transID = params['TransformationID']
```
is done in the PluginUtilities.setParameters function https://github.com/DIRACGrid/DIRAC/blob/rel-v6r14/TransformationSystem/Client/Utilities.py#L93
I don't find transID in TransformationPlugin itself. So that is not necessary, right?

phicharp:
Right... It';s just that in LHCbDIRAC we need it ;-) but you are right it is not used in the base class...

PR #2722, opened 2015-11-11
[v6r14] dirac-install: Make hashlib optional again

andresailer:
hashlib does not exist on SL5 by default (python 2.4) so use md5 instead (as it was before).
Also remove unused imports (pylint)


chaen:
review ok

PR #2735, opened 2015-12-07
[v6r14] HTCondorCE tweaks, log output

andresailer:
Some improvements for the HTCondorCE, still some open issues about the WorkingDirectory...
(Maybe adding a section in Operations for WorkingDirectory and how long the condor out/err/log files are kept?)

The Condor logging can now be obtained in the webinterface (that are the changes in WMSUtilities)

The pilot out and err files and the log files for each job are now placed in random locations.
Using find to find the files (which is fast enough...)
All log/out/err files older than 15 days are removed (using find again).

SIGTERM (instead of SIGKILL) is send to the application in case jobs are killed by the host site.

When pilots are put in held status we kill them in condor and mark them as aborted.
Some OSG sites allow opportunistic resources only 2 hours of uninterrupted running, after which jobs are held, but might be restarted later, to avoid this we have to kill the pilot. There are also other reasons (missing input files, outputfiles, resource offline) jobs are put in "Held" status all of which are bad and pilots are cheap enough to just kill them in this case.
Note that the "HELD" we use in this file should never make it outside as we set the status to "Aborted" after the pilot is killed

### DIFF
--- a/Resources/Computing/HTCondorCEComputingElement.py
+++ b/Resources/Computing/HTCondorCEComputingElement.py
@@ -59,9 +59,11 @@ def findFile( workingDir, fileName ):
   return S_OK(paths)
 
 def getCondorLogFile( pilotRef ):
-  """ return the location of the logFile belonging to the pilot reference """
+  """return the location of the logFile belonging to the pilot reference"""
   _jobUrl, condorID = condorIDFromJobRef( pilotRef )
-  resLog = findFile( '/opt/dirac/pro/runit/WorkloadManagement/SiteDirectorHT/', '%s.log' % condorID )
+  #FIXME: This gets called from the WMSAdministrator, so we don't have the same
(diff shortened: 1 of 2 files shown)


Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -50,12 +91,25 @@ def __writeSub( self, executable, nJobs ):
     """ Create the Sub File for submission
 
     """
-
     workingDirectory = self.ceParameters['WorkingDirectory']
     initialDir = '/'.join( workingDirectory.split('/')[:-1] )
```
atsareg: os.path.dirname( workingDirectory ) ?

andresailer: Changed to os.path.dirname

atsareg:
Review OK

PR #2736, opened 2015-12-09
[v6r15] Don't block the ReqProxy sweeping if one of the request is buggy

chaen:
If one of the request is malformed for a reason or another, the rest of the requests could not be swiped. This PR fixes this behavior.

### DIFF
--- a/RequestManagementSystem/Service/ReqProxyHandler.py
+++ b/RequestManagementSystem/Service/ReqProxyHandler.py
@@ -118,7 +118,7 @@ class ReqProxyHandler( RequestHandler ):
           gLogger.info( "sweeper: successfully put request '%s' @ ReqManager" % cachedName )
           os.unlink( cachedFile )
         except Exception as error:
-          gLogger.exception( "sweeper: hit by exception %s" % str( error ) )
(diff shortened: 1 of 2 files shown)


atsareg:
Review OK

chaen:
I added another fix in the Request state machine.
Currently, if a request has 2 Operations, with Op1 Waiting and Op2 Failed, the Request status will be Failed, which is not correct. 
Please review again

RequestManagementSystem/Service/ReqProxyHandler.py
```diff
@@ -117,9 +117,8 @@ def sweeper( cls ):
             continue
           gLogger.info( "sweeper: successfully put request '%s' @ ReqManager" % cachedName )
           os.unlink( cachedFile )
-        except Exception, error:
+        except Exception as error:
           gLogger.exception( "sweeper: hit by exception %s" % str( error ) )
```
phicharp: This is wrong!
should be
gLogger.exception( "sweeper: hit by exception", lException = error)
since str(error) is incorrect. Alternative: use repr(error)

chaen: Thanks, corrected

PR #2739, opened 2015-12-10
[v6r15] Do not return anything in DError, contrary to S_ERROR in SRM plugin

chaen:
Since this PR https://github.com/DIRACGrid/DIRAC/pull/2668 is not going to be merged in v6r15, I backported this important fix.

You can't add to DError. Also, the accounting is done in the StorageElement, no need to do it in the plugins


atsareg:
Why don't you just delete the Accounting related stuff but comment it out ?

chaen:
Corrected

PR #2759, opened 2016-01-07
[v6r14] Attempt to reconcile getCEStatus for LHCb and others

rajanandakumar:
This is a PR to presumably be discussed further in this thread.

Aim : Reconcile the usage of getCEStatus for the configuration as used for GridPP (robot certificates for pilot proxy) and LHCb (human certificates).

Attempt : Try to get the VO information. If a valid VO can be obtained, assume that the proxy is not a robot certificate and proceed. Otherwise proceed the other way.

Future of this PR :
1. Depends on the outcome of the infosys-discuss mailing list where a proper implementation of Glue2 on ARC CEs (among others) is being discussed. Once this is implemented by the sites (and ARC), the "robot certificate" method of using just the ARC api will presumably become the formally standard way of doing this.
2. If GridPP (and other instances?) move to using human certificates, the "if" statement can be removed.

This code is so far tested in the LHCb test instance and seems to work okay.

### DIFF
--- a/Resources/Computing/ARCComputingElement.py
+++ b/Resources/Computing/ARCComputingElement.py
@@ -303,7 +303,7 @@ class ARCComputingElement( ComputingElement ):
 
     result = S_OK()
     result['SubmittedJobs'] = 0
-    if (vo == '') :
+    if not vo:
       # Presumably the really proper way forward once the infosys-discuss WG comes up with a solution
       # and it is implemented. Needed for DIRAC instances which use robot certificates for pilots.
       endpoints = [arc.Endpoint( "ldap://" + self.ceHost + "/MDS-Vo-name=local,o=grid",
@@ -318,7 +318,7 @@ class ARCComputingElement( ComputingElement ):
       result['WaitingJobs'] = ceStats.WaitingJobs
     else:
       # The system which works properly at present for ARC CEs that are configured correctly.
-      # But for this we need the VO to be known
+      # But for this we need the VO to be known - ask me (Raja) for the whole story if interested.
       cmd = 'ldapsearch -x -LLL -H ldap://%s:2135 -b mds-vo-name=resource,o=grid "(GlueVOViewLocalID=%s)"' %(self.ceHost, vo.lower())
       res = shellCall( 0, cmd )
       if not res['OK']:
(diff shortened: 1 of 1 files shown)


Resources/Computing/ARCComputingElement.py
```diff
@@ -291,19 +294,49 @@ def getCEStatus( self ):
...
+    vo = ''
+    res = getVOfromProxyGroup()
+    if res['OK']:
+      vo = res['Value']
 
     result = S_OK()
-    result['RunningJobs'] = ceStats.RunningJobs
-    result['WaitingJobs'] = ceStats.WaitingJobs
     result['SubmittedJobs'] = 0
+    if (vo == '') :
```
chaen: if not vo:

Resources/Computing/ARCComputingElement.py
```diff
@@ -291,19 +294,49 @@ def getCEStatus( self ):
...
+      # The system which works properly at present for ARC CEs that are configured correctly.
+      # But for this we need the VO to be known
+      cmd = 'ldapsearch -x -LLL -H ldap://%s:2135 -b mds-vo-name=resource,o=grid "(GlueVOViewLocalID=%s)"' %(self.ceHost, vo.lower())
+      res = shellCall( 0, cmd )
+      if not res['OK']:
+        gLogger.debug("Could not query CE %s - is it down?" % self.ceHost)
+        return res
+      try:
+        ldapValues = res['Value'][1].split("\n")
+        running = [y for y in ldapValues if 'GlueCEStateRunningJobs' in y]
```
chaen: maybe a more meaningful name than 'y' ? :-)

Resources/Computing/ARCComputingElement.py
```diff
@@ -291,19 +294,49 @@ def getCEStatus( self ):
...
+        waiting = [y for y in ldapValues if 'GlueCEStateWaitingJobs' in y]
+        result['RunningJobs'] = int(running[0].split(":")[1])
+        result['WaitingJobs'] = int(waiting[0].split(":")[1])
+      except IndexError:
+        res = S_ERROR('Unknown ldap failure for site %s' % self.ceHost)
+        res['RunningJobs'] = 0
```
chaen: This does not really follow the standard that we set ourselves relatively recently (started with DError). S_ERROR should contain only 'OK' and 'Message' as key.

Resources/Computing/ARCComputingElement.py
```diff
@@ -291,19 +294,49 @@ def getCEStatus( self ):
...
+      # and it is implemented. Needed for DIRAC instances which use robot certificates for pilots.
+      endpoints = [arc.Endpoint( "ldap://" + self.ceHost + "/MDS-Vo-name=local,o=grid",
+                               arc.Endpoint.COMPUTINGINFO, 'org.nordugrid.ldapng')]
+      retriever = arc.ComputingServiceRetriever(self.usercfg, endpoints)
+      retriever.wait() # Takes a bit of time to get and parse the ldap information
+      targets = retriever.GetExecutionTargets()
+      ceStats = targets[0].ComputingShare
+      gLogger.debug("Running jobs for CE %s : %s" % (self.ceHost, ceStats.RunningJobs))
+      gLogger.debug("Waiting jobs for CE %s : %s" % (self.ceHost, ceStats.WaitingJobs))
+      result['RunningJobs'] = ceStats.RunningJobs
```
chaen: shouldn't it be 
result['Value']['RunningJobs'] = ceStats.RunningJobs  ?
And same bellow ?

rajanandakumar: Not sure ... as far as I can see, it works as it is currently. Also see lines 249, 250 in ComputingElement.py?

https://github.com/DIRACGrid/DIRAC/blob/integration/Resources/Computing/ComputingElement.py#L249

I will wait for your response on this. The rest of your comments are "of course - what was I thinking!" :)

chaen: Oh I see... yes there are places were we used to do that, so you obvously have to stay compatible... :-)

atsareg:
Review OK

PR #2760, opened 2016-01-12
[v6r14] Fix Service class

atsareg:
After enabling Service configuration to inherit another Service configuration with Module option, the Service name can be different from the name of the class of the service handler. The private self._name variable was still pointing to the class name. This was wrong and caused the StorageElement service to fail in case it was configured with the Module option.

Removed obsoleted _discoverHandlerLocation. This function is performed by the ModuleLoader.

### DIFF
--- a/Core/DISET/private/Service.py
+++ b/Core/DISET/private/Service.py
@@ -29,7 +29,7 @@ class Service:
 
   def __init__( self, serviceData ):
     self._svcData = serviceData
-    self._name = serviceData[ 'loadName' ]
+    self._name = serviceData[ 'modName' ]
     self._startTime = Time.dateTime()
     self._validNames = [ serviceData[ 'modName' ]  ]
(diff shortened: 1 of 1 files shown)


Core/DISET/private/Service.py
```diff
@@ -116,28 +116,6 @@ def initialize( self ):
 
     return S_OK()
 
-  def _discoverHandlerLocation( self ):
-    handlerLocation = self._cfg.getHandlerLocation()
-    if handlerLocation:
-      if handlerLocation.find( "Handler.py" ) != len( handlerLocation ) - 10:
```
phicharp: endswith()?

atsareg: This function is dropped altogether anyway

phicharp: oups, sorry, it was a removal ;-)

PR #2771, opened 2016-01-20
[v6r14] VOMSService outputs only string type

chaen:
Seal down the evilish suds.sax.text.Text type into VOMSService. The fix is already running in production in LHCb


phicharp:
Review OK!

Issue #2432, opened 2015-06-16
Feature Request: FileCatalogCLI: execfile, EOF


andresailer:
As I requested at the 4th (SIC) Dirac user workshop it would be nice to have a execfile method to execute a list of commands for the FC-CLI.
It would also be good if (all) the CLIs would understand the EOF (CTRL-D) character(?).

Currently I do
```
dirac-dms-filecatalog-cli < inputfile > output
```
which works nicely, as long as I don't forget to add "exit" as the last line of inputfile.

afortiorama:
Hi - I have the same problem.

PR #2787, opened 2016-02-02
Add EOF support for FC CLI

sposs:
as requested by #2432 

I can still do small things here and there... ;)


andresailer:
Thanks :) but why integration?
Also what about other CLIs?
Now if someone does execfile too...

sposs:
Oups, my bad... Which release do you want?

sposs:
recreated againts v6r14 in #2788

PR #2793, opened 2016-02-08
[v6r15] fix backward compatibility has access

chaen:
the signature in v6r14 used to be (operationName (str), paths (str, list, dict)) while in v6r15 it is (paths (dict), operationName (str)).
If the first argument arriving to the serving is a string, we swap the parameters. This should go away as soon as v6r16


atsareg:
Review OK

PR #2794, opened 2016-02-08
[v6r16] New fts3 only

chaen:
This PR contains the new FTS system designed to work only with FTS3. More doc and wiki instruction to follow once tested in certification


atsareg:
I guess we can drop the FTS2 completely in v6r16

chaen:
This system would not work with FTS2 anyway, and it is completely independant from the old system (the transition between both has to be made in a specific way though... I'll document all that)

chaen:
Hum interesting, merging and then pushing on an open PR then shows all the commits instead of only the merge commit....
in any case, I just added one commit here, which is to undo the comment @fstagni had done here https://github.com/DIRACGrid/DIRAC/pull/2816

PR #2802, opened 2016-02-10
[v6r14] Fix for Repo retrieve sandbox/outputdata

andresailer:
Retrieved and OutputData are strings '0' in the jobDict when repo file is read, need to cast to int
Replace #2801 for proper target


chaen:
review ok

chaen:
Maybe just a comment in the code to explain the cast ? :-)

andresailer:
added comments

PR #2803, opened 2016-02-10
[v6r15] rename cores to processors in resource description

pigay:
Renaming "Cores" to "Processors" was not complete here.


atsareg:
Review OK

PR #2806, opened 2016-02-10
[v6r14] Bug in DirectoryUsage table

chaen:
Bad accounting when removing several files from the same directory, and when removing replicas in a non existing SE

### DIFF
--- a/DataManagementSystem/DB/FileCatalogWithFkAndPsDB.sql
+++ b/DataManagementSystem/DB/FileCatalogWithFkAndPsDB.sql
@@ -910,7 +910,7 @@ DELIMITER ;
 
 
 -- ps_delete_files : delete files from file ids and update the DirectoryUsage table. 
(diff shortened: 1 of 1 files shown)


DataManagementSystem/DB/FileCatalogWithFkAndPsDB.sql
```diff
@@ -910,7 +910,7 @@ DELIMITER ;
 
 
 -- ps_delete_files : delete files from file ids and update the DirectoryUsage table. 
---                   CAREFUL : the cascade delete also removes the replicas
+--                   CAREFUL : the cascade delete also removes the replicas but will not update DierctoryUsage
```
atsareg: Spelling

atsareg:
Review OK, although I am not an expert in the MySQL stored procedures, I guess this was well tested

PR #2808, opened 2016-02-10
[v6r15] port tests for DirectoryUsage tabel from TestDIRAC

chaen:


atsareg:
These tests should work for both implementations of the Directory Tree ?

chaen:
In principle yes. And you tested them on your implementation last time you made changes, so I believe it is still the case. 
The tests I added here are just making sure that the accounting is correct, so he directory manager should be irrelevant. Though the tests could be improved still

PR #2809, opened 2016-02-10
[v6r14] Bug fixed

atsareg:
Bug fixed in __generateTarball


atsareg:
Trivial bug, I merge it straightaway

PR #2814, opened 2016-02-12
[v6r16] Response time monitor issue

Sbalbp:
As requested in #2138, modified the RequestHandler class to register the average response time of a service's methods using gMonitor.


zmathe:
It is recreated #2972

PR #2815, opened 2016-02-12
[v6r14] Using SE groups in ban/allowSE

phicharp:
Trivial change to be bale to allow/ban all SEs in an SE group at the same time (e.g. Tier1-ARCHIVE in LHCb)


chaen:
againt integration ?

PR #2818, opened 2016-02-16
[v6r14] Allow an SE group to be banned/allowed

phicharp:


chaen:
review ok

phicharp:
New PR with one more fix

PR #2820, opened 2016-02-17
[v6r14] Fix a crash in the RequestPreparationAgent 

chaen:
A corrupted staging request with an empty LFN made its way to the StorageManagementDB. This was causing the RequestPreparationAgent to crash


atsareg:
Review OK

PR #2821, opened 2016-02-17
[v6r14] Few minor fixes

phicharp:
* Allow SE group is dirac-admin-allow-se
* Fix reduction of number of files in TransformationAgent


chaen:
Review ok

PR #2824, opened 2016-02-18
[v6r15] Added default methods to the FileCatalogClientBase

atsareg:


Resources/Catalog/FileCatalogClientBase.py
```diff
@@ -48,4 +49,49 @@ def hasCatalogMethod( self, methodName ):
...
+      :param lfn paths: has to be formatted this way :
+                  { lfn : { se1 : pfn1, se2 : pfn2, ...}, ...}
+
+      :return: { successful : { lfn : [ ses ] } : failed : { lfn : { se : msg } } }
+
+    """
+    return S_OK( {'Failed' : {}, 'Successful' : dict.fromkeys( lfns, False )} )
+
+  @checkCatalogArguments
+  def getPathPermissions( self, lfns ):
```
chaen: Why not dict.fromkeys for this one ?

chaen:
The code is trivially okay, but doing so, aren't we going back to the base class we had before, and you so much wanted to get ride of ?

Another point is that the children classes also need to declare them in the list of methods. It would have been nice if instead of being module attributes, the lists were class attributes and this base class had the base methods in it, and the children just extend it.

atsareg:
Those methods are not just placeholders unlike the previous base class, so we are not going back. That's why they can help in case of pseudocatalogs. I think the method lists can indeed be class lists. In case the child does not want to expose some of the base class methods, it can overwrite rather than extend the base list.

PR #2829, opened 2016-02-19
[v6r16] DynamicMonitoringConsumer

Sbalbp:
Added a consumer class that connects to RabbitMQ, retrieving logs from the queue and sending them to ElasticSearch in an infinite loop.
Requires PR's #2619 and #2798 to be merged before.

### DIFF
--- a/FrameworkSystem/Consumer/DynamicMonitoringConsumer.py
+++ b/FrameworkSystem/Consumer/DynamicMonitoringConsumer.py
@@ -25,7 +25,22 @@ class DynamicMonitoringConsumer ( object ):
     """
     Reads messages from RabbitMQ and sends them to ElasticSearch
     """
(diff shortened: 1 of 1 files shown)


FrameworkSystem/Consumer/DynamicMonitoringConsumer.py
```diff
@@ -0,0 +1,66 @@
...
+      gLogger.error( result[ 'Message' ] )
+      return result
+
+    while True:
+      time.sleep( 1 )
+      # Get logs from the queue
+      result = self.rabbitMQ.get()
+      if result[ 'OK' ]:
+        # insert the logs into ElasticSearch
+        result = self.elasticDB.insertLog( result[ 'Value' ] )
```
zmathe: bulk insert is faster. It may require a bulk insert. But it is okay now.

zmathe:
It is recreated #2969

PR #2847, opened 2016-03-11
[v6r14] dirac-dms-user-lfns: Add error message when proxy is expired

andresailer:
When the proxy is expired the username field is not present, and the command fails with an exception.


atsareg:
Review OK

PR #2850, opened 2016-03-11
[v6r15] ConfigurationData.sync() performance

pigay:
Compressing `gConfigurationData.remoteCFG` data each time `sync()` is called takes only a few tens of a second. However, each user command does this a few times while it has no need to do so.

This PR should improve user experience a bit, especially with COMDIRAC cli.


chaen:
review OK

PR #2854, opened 2016-03-15
[v6r15] Cache DIRACPLAT in client installations

pigay:
scripts created with dirac-deploy-scripts will attempt to determine platform with `DIRAC.Core.Utilities.Platform.getPlatformString()`. This function uses `libc_ver()` which is fairly long. In user commands, it appends a delay about a few tens of seconds.

This PR caches the platform value at installation time for client installations through the DIRACPLAT environment variable.


pigay:
found in [dirac-bash](https://github.com/DIRACGrid/DIRAC/blob/integration/Core/scripts/dirac-bash) that DIRACPLAT could be initialized from the output of `dirac-platform` which seems better.

PR #2862, opened 2016-03-18
[v6r15] compute DIRAC platform only when requested

pigay:
Importing DIRAC module in a scripts implied computing of `DIRAC.platform` with `Platform.getPlatformString()` which is a bit long (see #2854).

This PR hides this field in a variable of the`Platform` module accessible through a `getPlatform()` method. All client scripts will be a few tens of seconds faster than before (except those who explicitly need platform string, like dirac-info).


atsareg:
Review OK

PR #2865, opened 2016-03-24
[v6r14] Make the StorageElementCache Thread safe

chaen:
For a yet unknown reason, the FTSAgent threads started to step on each other when using the StorageElement. One thread would ask for getFileMetadata, the other one for getURL, but the parameters of the two calls would be swapped (leading to exceptions, and to getFileMetadata returning an URL....). 
I tracked this down to the StorageElementCache not being thread safe, which is now fixed in this PR.


phicharp:
Review OK

PR #2867, opened 2016-03-31
[v6r14] Minor DMS fixes

phicharp:
* SE inheritance was not working well when some parameters were modified
* Fix DataManager for getActiveReplicas (first check Active before selecting disk SEs)


chaen:
review ok

PR #2868, opened 2016-04-01
[v6r15] Added default methods to the FileCatalogClientBase

atsareg:
Added default methods needed to be implemented by all the catalogs to be used in standard operations:

hasAccess()
exists()
getPathPermissions

In the default implementations the answer is True to all the requests

### DIFF
--- a/Resources/Catalog/FileCatalogClientBase.py
+++ b/Resources/Catalog/FileCatalogClientBase.py
@@ -7,6 +7,7 @@ __RCSID__ = "$Id$"
 
 from DIRAC.Core.Base.Client import Client
 from DIRAC.Core.Utilities.ReturnValues import S_OK
+from DIRAC.Resources.Catalog.Utilities import checkCatalogArguments
 
 class FileCatalogClientBase( Client ):
   """ Client code to the DIRAC File Catalogue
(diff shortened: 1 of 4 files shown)


chaen:
I think this implementation using class attributes is much better. 
However returning True by default for all permission checking is a bit dangerous no ?

Resources/Catalog/FileCatalogClientBase.py
```diff
@@ -35,17 +54,65 @@ def isOK( self, timeout = 120 ):
...
+      :param lfn paths: has to be formatted this way :
+                  { lfn : { se1 : pfn1, se2 : pfn2, ...}, ...}
+
+      :return: { successful : { lfn : [ ses ] } : failed : { lfn : { se : msg } } }
+
+    """
+    return S_OK( {'Failed' : {}, 'Successful' : dict.fromkeys( lfns, False )} )
+
+  @checkCatalogArguments
+  def getPathPermissions( self, lfns ):
```
chaen: Why not doing it the same way are the two others using fromkeys ?

atsareg:
@chaen : The default implementation to return True for all the permissions is for the pseudocatalogs where no permissions are assumed at all. But for those path filters will apply that you have developed

atsareg:
I have a strange glitch in the commit comments in my PyCharm... spotted, will try to fix

chaen:
yes I understand that, I was just thinking of new catalogs. But well, that does not happen everyday, so review ok

PR #2877, opened 2016-04-11
[v6r15] Fix bad merge of the SE

chaen:
In v6r15, the parameters is properly called "plugins", and not "protocols"


phicharp:
Review obviously OK

PR #2878, opened 2016-04-11
[v6r14] Properly checks if staging is allowed in JobScheduling

chaen:
All in the title. 
If everything went fine, a bool was returned, otherwise, S_ERROR was. The calling code was expecting only a boolean, which turns out to be True in case of problem.


phicharp:
Review OK

PR #2891, opened 2016-04-21
[v6r14] Gives string LFN to LFC

hmiyake:
Inside LcgFileCatalogClient, LFC wrapper does not accept unicode but only string.
This patch converts any basestring to string.


chaen:
Maybe you might just want to add a comment that this cast is for the lfcthr library, so that we know it is not useless, but otherwise, review OK

PR #2896, opened 2016-04-26
[v6r15] Improved user name evaluation from the DN

atsareg:
Improved utility to derive user name form the user DN


atsareg:
Merging as it is already running in production in FG-DIRAC for a while

PR #2906, opened 2016-05-03
[v6r14] use better (2D) plot for number of jobs running per pilot

zmathe:
Original:
8bae255422703a8e95ca152e300a4998
New:
20470a33a7abe73ca5cd4fe693dfce4e

### DIFF
--- a/AccountingSystem/private/DBUtils.py
+++ b/AccountingSystem/private/DBUtils.py
@@ -319,7 +319,7 @@ class DBUtils:
   
   def _getOverallMaxValue(self, dataDict):
     """
-    :param dict dataDict it returns the maximum value in overall period.
+    :param dict dataDict it returns the maximum value in the overall period.
     
     """
     maxValue = 0.0
--- a/AccountingSystem/private/DBUtils.py
+++ b/AccountingSystem/private/DBUtils.py
@@ -1,8 +1,6 @@
 import types
 from DIRAC.Core.Utilities import Time
 
-import numpy
-
 class DBUtils:
 
   def __init__( self, db, setup ):
@@ -324,7 +322,7 @@ class DBUtils:
     """
     maxValue = 0.0
     for i in dataDict.values():
-      currentMax =  numpy.max(i.values())
+      currentMax =  max(i.values())
       if maxValue < currentMax:
           maxValue = currentMax
     return maxValue
--- a/AccountingSystem/private/DBUtils.py
+++ b/AccountingSystem/private/DBUtils.py
@@ -314,4 +314,3 @@ class DBUtils:
           newData[ bt ] = 0.0
         newData[ bt ] += dataDict[ k ][ bt ]
     return newData
-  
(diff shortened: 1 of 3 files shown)


chaen:
Is this really something to go in a patch release ?

zmathe:
A new type of plot has introduced, which is only cerated when we set normalization value. It will not change the current functionality.

AccountingSystem/private/DBUtils.py
```diff
@@ -314,3 +314,16 @@ def _getBucketTotals( self, dataDict ):
           newData[ bt ] = 0.0
         newData[ bt ] += dataDict[ k ][ bt ]
     return newData
+  
+  def _getOverallMaxValue(self, dataDict):
+    """
+    It returns the maximum value in the overall period.
+    :param dict dataDict the format is { key: { time : [value1, value2, ...valueN] } } where key is the parameter what we want to plot, time is the certain period
```
chaen: shorter line please

AccountingSystem/private/DBUtils.py
```diff
@@ -314,3 +314,16 @@ def _getBucketTotals( self, dataDict ):
           newData[ bt ] = 0.0
         newData[ bt ] += dataDict[ k ][ bt ]
     return newData
+  
+  def _getOverallMaxValue(self, dataDict):
```
chaen: I am not quite sure what you are trying to do with this method. But I *think* that 
```max([ x for y in d.itervalues() for x in y.itervalues() ])``` does what you want no ?

File: Core/Utilities/Graphs/QualityMapGraph.py [hunk 3]
```diff
@@ -66,10 +66,17 @@ def __init__( self, data, ax, prefs, *args, **kw ):
 
     # Setup the colormapper to get the right colors
     self.cmap = LinearSegmentedColormap( 'quality_colormap', cdict, 256 )
-    #self.cmap = cm.RdYlGn
-    self.norms = normalize( 0, 100 )
+    
+    self.cmap = cm.RdYlGn
```
chaen: for better reading, I propose to put this into the else clause lower

AccountingSystem/private/DBUtils.py
phicharp: you can even remove the share brackets ;-)

AccountingSystem/private/DBUtils.py
```diff
@@ -314,3 +314,16 @@ def _getBucketTotals( self, dataDict ):
...
+    """
+    It returns the maximum value in the overall period.
+    :param dict dataDict the format is { key: { time : [value1, value2, ...valueN] } } where key is the parameter what we want to plot, time is the certain period
+    and value1,...valueN is the value.
+    """
+    maxValue = 0.0
+    for i in dataDict.values():
```
phicharp: itervalues()

atsareg:
Review OK

petricm:
Just a thought. What about a log scale with scale chosen so that 1 in the middle and the green red coloring?

atsareg:
It is a good starting point, and can be further improved based on the experience. I would prefer just green rather than blue. Tastes differ. I do not think log scale will be better, not so many jobs running per pilot

PR #2908, opened 2016-05-08
[v6r15] Documentation moved from DIRACDocs project to the docs directory

atsareg:


atsareg:
Merging since it is a mere copying over of the documentation files from DIRACDocs

PR #2916, opened 2016-05-10
[v6r14]  Fix for dirac-rms-cancel and TransformationCLI

andresailer:
RMS Cancel, also recently reported in diracgrid-forum

TransformationCLI: Protection against missing entry in dict


atsareg:
Review OK

PR #2941, opened 2016-05-27
[v6r14] Fix the ReqClient to allow it to talk to multiple ReqManager

chaen:
There was a bug in the ReqClient preventing to have multiple instance of the ReqManager.
Basically the client would talk to ```url1:9140/RequestManagement/ReqManager``` asking for ```url2:9140/RequestManagement/ReqManager```, resulting in the error message ```RequestManagement/ReqManager is not up in this server```


chaen:
Just a note about the new error found by landscape. The error is ```Use of super on an old style class```
However it is a false positive. See https://github.com/landscapeio/landscape-issues/issues/151

atsareg:
Landscape is reporting a new error in this PR

atsareg:
Ah, just noticed the comment about the landscape error ( we have many more comments now ). It is OK then

PR #2942, opened 2016-05-27
[v6r15] Protecting the parsing of pilot commands output.

CinziaLu:
Protecting the parsing of pilot commands output.
See #2928

### DIFF
--- a/WorkloadManagementSystem/PilotAgent/pilotCommands.py
+++ b/WorkloadManagementSystem/PilotAgent/pilotCommands.py
@@ -384,14 +384,15 @@ class CheckCECapabilities( CommandBase ):
                                                                         self.pp.queueName,
                                                                         " ".join( self.cfg ) )
     retCode, resourceDict = self.executeAndGetOutput( checkCmd, self.pp.installEnv )
+    if retCode:
+        self.log.error( "Could not get resource parameters [ERROR %d]" % retCode )
+        self.exitWithError( retCode )
     try:
       import json
       resourceDict = json.loads( resourceDict )
     except ValueError:
       self.log.error( "The pilot command output is not json compatible." )
       sys.exit( 1 )
-    if retCode:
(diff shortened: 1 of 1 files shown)


WorkloadManagementSystem/PilotAgent/pilotCommands.py
```diff
@@ -432,11 +435,15 @@ def execute( self ):
     checkCmd = 'dirac-wms-get-wn-parameters -S %s -N %s -Q %s %s' % ( self.pp.site, self.pp.ceName, self.pp.queueName,
                                                                        " ".join( self.cfg ) )
     retCode, result = self.executeAndGetOutput( checkCmd, self.pp.installEnv )
-
-    result = result.split( ' ' )
-    numberOfProcessor = int( result[0] )
-    maxRAM = int( result[1] )
-
+    if retCode:
+      self.log.warn( "Could not get resource parameters [ERROR %d]" % retCode )
```
chaen: If you have the error code, why is it just a warning and why are you trying to split the result a bit lower ?

CinziaLu: Because, at the beginning I did not want to stop the pilot if we did not have the parameters. 
Fixed now.

WorkloadManagementSystem/PilotAgent/pilotCommands.py
```diff
@@ -432,11 +437,16 @@ def execute( self ):
...
-
-    result = result.split( ' ' )
-    numberOfProcessor = int( result[0] )
-    maxRAM = int( result[1] )
-
+    if retCode:
+      self.log.error( "Could not get resource parameters [ERROR %d]" % retCode )
+      self.exitWithError( retCode )
+    try:
+      result = result.split( ' ' )
```
atsareg: I think here it should be rather just result = result.split(), we need here splitting by white spaces whatever the number of those

PR #2947, opened 2016-06-02
Rel v6r15 pre7

vmendez:
l fix the "hack" taking SubmitPools list as SubmitPool to match, because it only works when SumbitPools of the VO is just one submit pool


atsareg:
Conflicting PR. @vmendez: please contact me before another attempt

PR #2951, opened 2016-06-02
[v6r15] Bug fixes and better logging in the Matcher

atsareg:
Fixed a bug resulting in ignoring multiProcessor job requirement for sites presenting an empty Tag list.

Better and prettier logging of the matching parameters

### DIFF
--- a/Core/Utilities/PrettyPrint.py
+++ b/Core/Utilities/PrettyPrint.py
@@ -19,76 +19,107 @@ def int_with_commas(i):
 
 def printTable( fields, records, sortField='', numbering=True, 
                 printOut = True, columnSeparator = ' ' ):
-    """ Utility to pretty print tabular data
-    """
+  """ Utility to pretty print tabular data
 
-    stringBuffer = StringIO.StringIO()
+  :param list fields: list of column names
+  :param list records: list of records, each record is a list or tuple of string values
+  :param str sortField: name of the column by which the output will be sorted
+  :param bool numbering: flag for numbering rows
+  :param bool printOut: flag for printing into the stdout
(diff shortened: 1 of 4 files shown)


Andrew-McNab-UK:
Can I clarify what the behaviour will be after this change? 

Currently if I have a batch queue on a CE where the batch queue has the tag Multiprocessor, then single processor jobs without that tag still go there, and multiprocessor jobs with that tag also go to non-tagged queues/CEs elsewhere.

The sites tend to have a multiprocessor queue (e.g. "mcore") alongside a generic queue ("long") for single processor jobs. Will this change get the MP and SP jobs only to the right type of queue?

atsareg:
If a job has a tag requirement Multiprocessor it will only be matched by a pilot which is presenting this tag. So, it should be checked that the queue tag Multiprocessor is picked up by the pilot. And vice versa. If pilot only should take jobs with tag Multiprocessor then it should present a RequiredTag parameter with the value Multiprocessor. In this case the double matching is done. So, both job abd site requirement must be defined.
This change is fixing the bug that when pilot was presenting Tag parameter being an empty list, the job requirement was simply ignored.

pigay:
Review ok

Core/Utilities/PrettyPrint.py
```diff
@@ -9,8 +9,13 @@
 
 import StringIO
 
-def int_with_commas(i):
-  s = str(i)
+def int_with_commas( inputValue ):
+  """ Utility to make a string of a large integer with comma separated degrees
+      of thousand
+  :param int inputValue: value to be interpreted
```
andresailer: You need an empty line before the parameter list starts.

PR #2956, opened 2016-06-03
[v6r15] Fix landscape

petricm:
Landscape does not work at the moment because there is a missing line for pylint. Sorry was testing on a more advanced example, and then wanted to make it as simple as possible, apparently too simple.

### DIFF
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,3 +1,6 @@
+pylint:
+  run: true
+
 pep8:
   disable:
     - E114


atsareg:
Review OK

petricm:
I don't get it completely, why there is no landscape response in the conversation of the PR.

My commit
https://github.com/petricm/DIRAC/commit/496c7aca849ffac9c32a0c1d4977e01a78bcf57c
passed landscape
https://landscape.io/github/petricm/DIRAC/54

But it looks like that nothing was triggered for DIRACGrid

PR #2958, opened 2016-06-05
[v6r15] Fix dirac-rss-sync command to not call JobDB directly

atsareg:
Accessing ResourceStatusDB and JobDB through clients rather than directly.
Proper order in the script initialization and import statements

### DIFF
--- a/ResourceStatusSystem/scripts/dirac-rss-sync.py
+++ b/ResourceStatusSystem/scripts/dirac-rss-sync.py
@@ -19,13 +19,8 @@
 """
 
 from datetime import datetime, timedelta
-from DIRAC                                            import gConfig, gLogger, exit as DIRACExit, S_OK, version
-from DIRAC.Core.Base                                  import Script
-from DIRAC.WorkloadManagementSystem.DB.JobDB          import JobDB
-from DIRAC.ResourceStatusSystem.DB.ResourceStatusDB   import ResourceStatusDB
-from DIRAC.ResourceStatusSystem.Utilities             import Synchronizer, CSHelpers, RssConfiguration
-from DIRAC.ResourceStatusSystem.Client                import ResourceStatusClient
-from DIRAC.ResourceStatusSystem.PolicySystem          import StateMachine
+from DIRAC import version, gLogger, exit as DIRACExit, S_OK
+from DIRAC.Core.Base  import Script
 
 __RCSID__  = '$Id$'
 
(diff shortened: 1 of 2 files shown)


ResourceStatusSystem/scripts/dirac-rss-sync.py
```diff
@@ -127,10 +135,9 @@ def initSites():
     Initializes Sites statuses taking their values from the "SiteMask" table of "JobDB" database.
   '''
 
-  jobClient = JobDB()
-  rssClient = ResourceStatusDB()
+  rssClient = ResourceStatusClient.ResourceStatusClient()
```
vipersec: `ResourceStatusClient.ResourceStatusClient()` does not contain ResourceStatusDB's `addIfNotThere` function which is used after a few lines [here](https://github.com/atsareg/DIRAC/blob/32a6f5f310d42203e65bed28ded69defdc467998/ResourceStatusSystem/scripts/dirac-rss-sync.py#L151)

atsareg: Ah, I see. I did not pay attention that this client is not inheriting the base Client class. Is there a special reason for that ? With this inheritance all the service methods are immediately available as if they implemented in the client. Instead, there is an internal redirection which is doing essentially the same thing.

atsareg:
Not subject to this PR, but the ResourceStatusClient implementation must be reviewed. I think it can be ways simpler by making it inherit the Client base class.

ResourceStatusSystem/scripts/dirac-rss-sync.py
```diff
@@ -141,7 +148,7 @@ def initSites():
     insert = { 'Status': elements[0], 'Reason': 'Synchronized', 'Name': site, 'DateEffective': elements[1], 'TokenExpiration': Datetime,
              'ElementType': 'Site', 'StatusType': 'all', 'LastCheckTime': None, 'TokenOwner': elements[2] }
 
-    result = rssClient.addIfNotThere(insert, table)
+    result = rssClient.addIfNotThereStatusElement( insert, table )
```
vipersec: The `addIfNotThereStatusElement` gets string parameters and not dictionaries as far as I know.

ResourceStatusSystem/scripts/dirac-rss-sync.py
vipersec: Perhaps it should be changed in the near future.

atsareg: Ok, I made a fix  following your suggestions, it works

PR #2977, opened 2016-06-08
[v6r16] Change ExtraModule to Extensions

zmathe:
This fix the issue reported #2099. The backward compatibility is kept. But there is one exception:

dirac-install --extraModules 

I assume we are using -e instead of --extraModules.

### DIFF
--- a/Core/scripts/dirac-configure.py
+++ b/Core/scripts/dirac-configure.py
@@ -88,6 +88,7 @@ vo = None
 update = False
 outputFile = ''
 skipVOMSDownload = False
+extensions = ''
 
 def setGateway( optionValue ):
   global gatewayServer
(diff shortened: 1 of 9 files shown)


Core/scripts/dirac-configure.py
```diff
@@ -302,6 +311,16 @@ def forceUpdate( optionValue ):
   if newVO:
     setVO( newVO )
 
+if not extensions:
+  newExtensions = DIRAC.gConfig.getValue( cfgInstallPath( 'ExtraModules' ) )
+  if newExtensions:
+    DIRAC.gLogger.warn( "extraModules is deprecated please use extensions instead!" )
```
atsareg: Message should rather read:
ExtraModules option is deprecated, use Extensions option instead

zmathe:
done

atsareg:
Review OK

PR #2979, opened 2016-06-08
[v6r14] DIPStorage bug fix

atsareg:
In putFile when copying from another DIPStorageElement ( a la third party transfer ), the service name was not properly evaluated, fix for #2413

### DIFF
--- a/Resources/Storage/DIPStorage.py
+++ b/Resources/Storage/DIPStorage.py
@@ -106,8 +106,14 @@ class DIPStorage( StorageBase ):
     localCache = False
     srcDict = res['Value']
     if srcDict['Protocol'] in ['dips', 'dip']:
+      serviceDict = dict( srcDict )
+      serviceDict['Path'] = '/'.join( srcDict['Path'].split('/')[:3] )
+      serviceDict['FileName'] = ''
+      res = pfnunparse( serviceDict )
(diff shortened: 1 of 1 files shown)


Resources/Storage/DIPStorage.py
```diff
@@ -106,8 +106,14 @@ def __putFile( self, src_file, dest_url ):
     localCache = False
     srcDict = res['Value']
     if srcDict['Protocol'] in ['dips', 'dip']:
+      serviceDict = dict( srcDict )
```
chaen: Can you please add comments ? I don't really understand, in particular the ```[:3]```, nor why FileName is empty. Does that effectively make dips a third party protocol ?

atsareg: This is stripping off the file part in the dips url leaving only the service part as needed for the TransferClient. I have added a comment

chaen:
Review OK. However we should never reach this part of the code since dips was removed from the third party transfer list, so we would go via local cache anyway in the DataManager

atsareg:
With this fix the dips can be put back into the list of third party protocols. Right now ti will do the same thing as the DataManager using cache. But when eventually dips will do a real third party transfer, nothing will need changes in the configuration

PR #2986, opened 2016-06-14
[v6r14] Strip of comments in the job JDL

atsareg:
Strip of comments in the job JDL before submission to the JobManager. In some cases comments are screwing up the job processing


chaen:
Review OK

PR #2988, opened 2016-06-15
[v6r15] print request id for replicate and register requests

marianne013:
We have users who, especially for one off requests, use the dirac-dms-replicate-and-register-request command directly. Right now this command only prints the request name, which doesn't make allowances for duplicate request names. This patch now also prints the requestid(s), giving the user a chance to monitor their request(s).

### DIFF
--- a/DataManagementSystem/scripts/dirac-dms-replicate-and-register-request.py
+++ b/DataManagementSystem/scripts/dirac-dms-replicate-and-register-request.py
@@ -58,6 +58,7 @@ if __name__ == "__main__":
   count = 0
   reqClient = ReqClient()
   fc = FileCatalog()
+  requestIDs = []
   for lfnChunk in lfnChunks:
     metaDatas = fc.getFileMetadata( lfnChunk )
     if not metaDatas["OK"]:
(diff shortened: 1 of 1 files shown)


DataManagementSystem/scripts/dirac-dms-replicate-and-register-request.py
```diff
@@ -106,13 +107,15 @@ def getLFNList( arg ):
...
       continue
-
+    requestIDs.append( str( putRequest["Value"] ) )
     if not multiRequests:
       gLogger.always( "Request '%s' has been put to ReqDB for execution." % request.RequestName )
 
   if multiRequests:
     gLogger.always( "%d requests have been put to ReqDB for execution, with name %s_<num>" % ( count, requestName ) )
-  gLogger.always( "You can monitor requests' status using command: 'dirac-rms-show-request <requestName>'" )
+  if len(requestIDs) > 0:
```
chaen: Hi,

Thanks for the PR !
The prefered way of such tests according to PEP8 is ```if requestIDs:```

After this fix, review OK

marianne013:
OK, I think that's fixed now.

chaen:
Review OK

PR #2990, opened 2016-06-16
[v6r14] Bug fix to extract output and error files

yujikato:
Modifications in L659 and 666 is necessary to extract output and error files
for all the batch systems.  Result['Value'] is tuple but string is expected to be returned.

Modification in L605 is related to SSHTorque.
The jobStamp of Torque includes hostname of the head node.
But name of output file does not include it. So somehow tricky treatment is necessary.

We confirmed this code is working under Belle II environment.

### DIFF
--- a/Resources/Computing/SSHComputingElement.py
+++ b/Resources/Computing/SSHComputingElement.py
@@ -602,7 +602,14 @@ class SSHComputingElement( ComputingElement ):
     if 'OutputTemplate' in self.ceParameters:
       self.outputTemplate = self.ceParameters['OutputTemplate']
       self.errorTemplate = self.ceParameters['ErrorTemplate']
-    
+
+    if self.batchSystem == 'Torque':
+      output = self.outputTemplate % jobStamp.split('.')[0]
+      error = self.errorTemplate % jobStamp.split('.')[0]
+    else:
+      output = self.outputTemplate % jobStamp
+      error = self.errorTemplate % jobStamp
+      
     if self.outputTemplate:
       output = self.outputTemplate % jobStamp
       error = self.errorTemplate % jobStamp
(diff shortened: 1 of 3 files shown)


Resources/Computing/SSHComputingElement.py
```diff
@@ -602,7 +602,14 @@ def _getJobOutputFiles( self, jobID, host = None ):
...
       self.outputTemplate = self.ceParameters['OutputTemplate']
       self.errorTemplate = self.ceParameters['ErrorTemplate']
-    
+
+    if self.batchSystem == 'Torque':
+      output = self.outputTemplate % jobStamp.split('.')[0]
+      error = self.errorTemplate % jobStamp.split('.')[0]
+    else:
+      output = self.outputTemplate % jobStamp
+      error = self.errorTemplate % jobStamp
```
atsareg: This fix is Torque dependent in a generic class. This is not good. If you just leave only two lines:
  output = self.outputTemplate % jobStamp.split('.')[0]
  error = self.errorTemplate % jobStamp.split('.')[0]
the result will be the same as before for "non-Torque" sites and correct for "Torque" sites.
However, I would like to ask you to rather modify the BatchSystems/Torque.py module so that the host name is stripped of in the submitJob method before returning the resulting JobIDs. In this case, the Torque pecularities will be contained in the specific Torque module. Since you have the case of such site you can make a proper test.

atsareg:
Review OK

yujikato:
Modified to strip hostname at Torque.py instead of SSHComputintElement.py 
as requested.  Modified pilotCommands.py for the consistency.

The status 'C' is moved to "Running" as in this status pilot output is still in the buffere 
and can not retrieve the output.

These changes are confirmed to work at the Belle II environment.

PR #2998, opened 2016-06-17
Bugfix in proxy name and group + New VOMS return 1 when only Warning

vmendez:
to  "dirac-admin-get proxy -all" and VOMS2CSAgent

### DIFF
--- a/Core/Security/VOMS.py
+++ b/Core/Security/VOMS.py
@@ -279,8 +279,10 @@ class VOMS( BaseSecurity ):
     status, output, error = result['Value']
 
     if status:
-      self._unlinkFiles( newProxyLocation )
-      return S_ERROR( DErrno.EVOMS, 'Failed to set VOMS attributes. Command: %s; StdOut: %s; StdErr: %s' % ( cmd, output, error ) )
+      # This warning is not avoiding of having an AC:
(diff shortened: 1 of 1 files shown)


chaen:
If I understand well what it does, I am not sure it is a good idea. We would then end up having proxies in the system that are valid at the time of the job submission for example, but not when it comes to run the jobs or execute the requests. This possibility already exists today, but we have this minimum duration as safety margin, so it almost never happens.

chaen:
are we speaking about the same PR ? From what I see in the code, you just modified VOMS not to return an error if the error contains "The validity of this VOMS AC in your proxy is shortened" 
I don't see the relation with all what you say here. But again, maybe I just misunderstand what this PR is solving

vmendez:
From PR title, in previous explanation:
"Bugfix in proxy name and group" -> 2)
"New VOMS return 1 when only Warning" -> 1) 

Sorry if I did not give an extensive description, I guest now is clear enough, anyway the fix is few lines very trivial.

atsareg:
There is nothing in this PR concerning "Bugfix in proxy name and group" -> 2)

vmendez:
You are right, I close this PR and I am submitting the good one

PR #2999, opened 2016-06-20
good Bugfix in proxy name and group + New VOMS return 1 when only Warning

vmendez:


atsareg:
I see now. This is not a bug in the VOMS2CSAgent. The keyed arguments are actually interpreted by the decorator.

vmendez:
It is not interpreter, since
def __syncCSWithVOMS( self, vo ):
no variable asignment is not working for:
 result = self.__syncCSWithVOMS( vo, proxyUserName = voAdminUser, proxyUserGroup = voAdminGroup ) 
I don't know if 
#pylint: disable=unexpected-keyword-arg
but the fix is working anycase I would prefer to avoid trust in platform parameters changing for a simple update

If your VOMS is not asking AC for list of user, this is neither affecting your, since a simple proxy is enough. But if your VOMS is asking for AC, then you need to play with pylint parameters (not desiderable) or just accept this simple and clear fix.

ConfigurationSystem/Agent/VOMS2CSAgent.py
```diff
@@ -57,7 +57,7 @@ def execute( self ):
 
       self.log.info( 'Performing VOMS sync for VO %s with credentials %s@%s' % ( vo, voAdminUser, voAdminGroup ) )
 
-      result = self.__syncCSWithVOMS( vo, proxyUserName = voAdminUser, proxyUserGroup = voAdminGroup ) #pylint: disable=unexpected-keyword-arg
+      result = self.__syncCSWithVOMS( vo, voAdminUser, voAdminGroup )
```
sposs: This does not work as expected: the decorator NEEDS the keyword arguments.

ConfigurationSystem/Agent/VOMS2CSAgent.py
```diff
@@ -81,7 +81,7 @@ def execute( self ):
     return S_OK()
 
   @executeWithUserProxy
-  def __syncCSWithVOMS( self, vo ):
+  def __syncCSWithVOMS( self, vo, proxyUserName, proxyUserGroup ):
```
sposs: Specify default values, to None if needed, as otherwise there is a high risk of failure.

PR #3000, opened 2016-06-20
proxy user and groups works anycase

vmendez:
After your comments,

### DIFF
--- a/ConfigurationSystem/Agent/VOMS2CSAgent.py
+++ b/ConfigurationSystem/Agent/VOMS2CSAgent.py
@@ -57,7 +57,7 @@ class VOMS2CSAgent( AgentModule ):
 
       self.log.info( 'Performing VOMS sync for VO %s with credentials %s@%s' % ( vo, voAdminUser, voAdminGroup ) )
 
-      result = self.__syncCSWithVOMS( vo, proxyUserName = voAdminUser, proxyUserGroup = voAdminGroup ) #pylint: disable=unexpected-keyword-arg
+      result = self.__syncCSWithVOMS( vo, voAdminUser, voAdminGroup )  
       if not result['OK']:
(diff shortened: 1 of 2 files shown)


Core/Security/VOMS.py
```diff
@@ -279,8 +279,10 @@ def setVOMSAttributes( self, proxy, attribute = None, vo = None ):
     status, output, error = result['Value']
 
     if status:
-      self._unlinkFiles( newProxyLocation )
-      return S_ERROR( DErrno.EVOMS, 'Failed to set VOMS attributes. Command: %s; StdOut: %s; StdErr: %s' % ( cmd, output, error ) )
+      # This warning is not avoiding of having an AC:
+      if not error.find("The validity of this VOMS AC in your proxy is shortened"):
+        self._unlinkFiles( newProxyLocation )
+        return S_ERROR( 'Failed to set VOMS attributes. Command: %s; StdOut: %s; StdErr: %s' % ( cmd, output, error ) )
```
chaen: please keep the error number

ConfigurationSystem/Agent/VOMS2CSAgent.py
```diff
@@ -81,7 +81,7 @@ def execute( self ):
     return S_OK()
 
   @executeWithUserProxy
-  def __syncCSWithVOMS( self, vo ):
+  def __syncCSWithVOMS( self, vo, proxyUserName=None, proxyUserGroup=None ):
```
chaen: As @atsareg said, these two parameters are used by the decorator. In fact they are even poped from kwargs, so they will never make it to the method. And this is why there was the pylint helper above.

vmendez:
Without  pylint it is not working.
Proposed change works with or without pylint.

PR #3009, opened 2016-06-23
[v6r14] Allow staging at non local SE

phicharp:
Allow staging to take place at an SE accessible by protocol, while until now it was only local. This was preventing a job at NIKHEF to stage at SARA.


chaen:
against integration ?

PR #3010, opened 2016-06-23
[v6r14] Allow staging from an SE accessible by Protocol

phicharp:
Allow staging to take place at an SE accessible by protocol, while until now it was only local. This was preventing a job at NIKHEF to stage at SARA.
After discussion with the reviewer, try and get both the VO and the InputDataPolicy from the job manifest in order to do it more correctly. As the agent is not running with a proxy, Operations() cannot deduce the VO and it should be picked up in the job for multi-VO installations.

### DIFF
--- a/WorkloadManagementSystem/Executor/JobScheduling.py
+++ b/WorkloadManagementSystem/Executor/JobScheduling.py
@@ -355,11 +355,6 @@ class JobScheduling( OptimizerExecutor ):
 
   def __preRequestStaging( self, jobState, stageSite, opData ):
     from DIRAC.DataManagementSystem.Utilities.DMSHelpers import DMSHelpers
-    # Allow staging from SEs accessible by protocol
-    result = DMSHelpers().getSEsForSite( stageSite, connectionLevel = 'PROTOCOL' )
-    if not result['OK']:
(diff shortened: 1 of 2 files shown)


WorkloadManagementSystem/Executor/JobScheduling.py
```diff
@@ -354,7 +354,9 @@ def __resolveStaging( self, jobState, inputData, idSites ):
     return ( True, bestSites )
 
   def __preRequestStaging( self, jobState, stageSite, opData ):
-    result = getSEsForSite( stageSite )
+    from DIRAC.DataManagementSystem.Utilities.DMSHelpers import DMSHelpers
```
chaen: I would just reorganize the code differently. A bit bellow, you get the vo.
So instanciate the DMSHelpers after that, so that you can pass it the vo parameters. 
Also, the import can be done at the beginning of the file no ?

phicharp: You are right actually, since optimizers may not know which VO is concerned, we should get it from the manifest and pass it onto DMSHelpers()

chaen:
review ok

Issue #3020, opened 2016-06-30
[v6r14] "Source local file does not exist" errors with multiple ProxyStorage instances


ahaupt:
In case there are multiple ProxyStorage instances configured, DMS client has a high chance to run into a bug resulting in upload/download failures. Error message: "Source local file does not exist". As the sURL lookup is done twice two different sURLs for the ProxyStorage service could be returned.

In such a case the second ProxyStorage instance does not know about the file and fails.

PR #3021, opened 2016-06-30
[v6r14] Make sure transferClient connects to the same ProxyStorage instance

ahaupt:
Make sure transferClient connects to the same ProxyStorage instance to which the uploaded/downloaded file has been cached before.

In setups with multiple ProxyStorage instances two independent sURL lookups led to "file not found" errors in case they returned different results.

Should fix: #3020

### DIFF
--- a/Resources/Storage/ProxyStorage.py
+++ b/Resources/Storage/ProxyStorage.py
@@ -33,6 +33,8 @@ class ProxyStorage( StorageBase ):
     failed = {}
     successful = {}
     client = RPCClient( self.url )
+    # Make sure transferClient uses the same ProxyStorage instance.
+    # Only the this one holds the file we want to transfer.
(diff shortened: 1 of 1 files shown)


chaen:
Review OK. Maybe just add a comment in the code to specify why you do that. It is easy for someone coming after you not to see the interest and replace it back with self.url ;-)

ahaupt:
Well, it would be the first code comment in the hole file ...

chaen:
:D You have a (very good) point :-) But this is very old code, and we have decided since then to adopt better/good/any development practices :-) When we now write new code or modify old one, we adopt these practices and add the appropriate comments. However, you can feel free to add more that just this one comment in the file ;-)

chaen:
Thanks :-)
Review OK

PR #3025, opened 2016-07-05
[v6r14] RMS: Add possibility of constant delay

chaen:
Instead of incrementing the delay at each attempt, there might be (there is !) a wish to have a constant delay.


phicharp:
Review OK

PR #3046, opened 2016-07-26
[v6r15] CREAMComputingElement: Fix parsing of return value

andresailer:
printout stdout and stderr for failed calls. (Not entirely sure what is in [1] and what is in [2], or where the commands even put the output on errors, so print both just in case).

first change: the error string is more often in item 1 of `result['Value']`
e.g.:
```
ERROR:  Pilot submission failed with error: 1: 2016-07-26 15:36:56,585 FATAL - Received NULL fault; the error is due to another cause: FaultString=[connection error] - FaultCo
de=[SOAP-ENV:Client] - FaultSubCode=[SOAP-ENV:Client] - FaultDetail=[Connection timed out], 2:
``` 
(added `1:`, `2:` during debugging session)

second change: this was a bug, causing exception.
The 0 item is an int, instead at least 1 should be there

### DIFF
--- a/Resources/Computing/CREAMComputingElement.py
+++ b/Resources/Computing/CREAMComputingElement.py
@@ -120,7 +120,8 @@ class CREAMComputingElement( ComputingElement ):
       if result['OK']:
         if result['Value'][0]:
           # We have got a non-zero status code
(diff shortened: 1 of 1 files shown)


atsareg:
There is a strange result of the landscape check with new detected problems that have nothing to do with this PR. Would be nice to understand why this happened

PR #3048, opened 2016-07-27
[v6r15] ReplicateLFN: Check that there is no local file that could get overwr…

andresailer:
…itten when third party copy is not taking place.

This function gets called both via dirac-dms-replicate-lfn and in the filecatalog cli.

### DIFF
--- a/Interfaces/API/Dirac.py
+++ b/Interfaces/API/Dirac.py
@@ -1333,6 +1333,15 @@ class Dirac( API ):
     if not isinstance( localCache, basestring ):
       return self._errorReport( 'Expected string for path to local cache' )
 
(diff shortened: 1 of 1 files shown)


chaen:
review OK.

atsareg:
Why this should be an error ? Why not just overwrite it ? it is a cache. Or there can be a copy of the same name file going on in parallel ? Did you get such a case ?

andresailer:
yes, we did get that case. https://its.cern.ch/jira/browse/ILCDIRAC-698

atsareg:
Let's incorporate this PR. But I think it is better to generate a tmp unique file name instead of using the LFN basename. Then this error would never happen. But this needs changes in the DataManager.

PR #3054, opened 2016-08-03
[v6r16] MessageQueue resources

atsareg:
This PR replaces #2985, which was badly screwed by a faulty rebase. All the comments from #2985 are either incorporated or answered. 

This PR contains classes for managing MQ connections as DIRAC Resources. The current implementation is only for the Stomp MQ protocol which is provided by the RabbitMQ. This is a wrapper over the MQConnection classes by Sergio as in #2974, and replaces this PR.

Users are supposed to use MQListener and MQPublisher classes which can be used as simply as the following:

publisher = MQPublisher( "MyQueue" )
listener = MQListener( "MyQueue" )
publisher.put( message )
listener.get( message )

Or:
publisher = MQPublisher( "MyQueue" )
listener = MQListener( "MyQueue", callback = myCallback )
publisher.put( message )
listener.run()

Example of the MQ configuration description is the following:

Resources
{
MQServices
{
mardirac3.in2p3.fr
{
MQType = Stomp
Host = mardirac3.in2p3.fr
Port = 9165
User = guest
Password = guest
Queues
{
TestQueue
{
Acknowledgement = True
}
}
}
}
}

The queueName in the listener/publisher instantiation can be given as just the Queue name (if no ambiguities ) or fully qualified name like "mardirac3.in2p3.fr::TestQueue" to avoid ambiguities.

### DIFF
--- a/Resources/MessageQueue/MQListener.py
+++ b/Resources/MessageQueue/MQListener.py
@@ -4,16 +4,17 @@
     The MQListener can be used in two modes:
 
     Simple mode with a default callback allows to get messages one by one by an explicit
-    get() call:
+    get() call::
 
       listener = MQListener( queueName )
       listener.get()
 
     Consumer mode with a custom callback function provided in the constructor or through
-    setCallback call:
+    setCallback call::
 
       listener = MQListener( queueName, callback = myCallback )
-      listener run()
+      listener.run()
+
 """
 
 import time
--- a/Resources/MessageQueue/MQListener.py
+++ b/Resources/MessageQueue/MQListener.py
@@ -1,0 +1,80 @@
+""" MQListener class encapsulates listener functionality fully configurable
+    by the CS options
+
+    The MQListener can be used in two modes:
+
+    Simple mode with a default callback allows to get messages one by one by an explicit
+    get() call:
+
+      listener = MQListener( queueName )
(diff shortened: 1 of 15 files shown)


Resources/MessageQueue/MQListener.py
```diff
@@ -0,0 +1,80 @@
...
+    get() call:
+
+      listener = MQListener( queueName )
+      listener.get()
+
+    Consumer mode with a custom callback function provided in the constructor or through
+    setCallback call:
+
+      listener = MQListener( queueName, callback = myCallback )
+      listener run()
```
andresailer: `listener.run()`

Resources/MessageQueue/MQListener.py
```diff
@@ -0,0 +1,80 @@
+""" MQListener class encapsulates listener functionality fully configurable
+    by the CS options
+
+    The MQListener can be used in two modes:
+
+    Simple mode with a default callback allows to get messages one by one by an explicit
+    get() call:
```
andresailer: You need two colons to turn the indented parts into a nice block
i.e.:
```
get() call::

      listener = MQListener( queueName )
      listener.get()
```
(or prefix with ">>>" or use ".. code-block::" )

PR #3055, opened 2016-08-04
[v6r15] HTCondorCE: Make better configurable

andresailer:
Add CS options:
   * add extraOption string to the submission file, based on the ARCCE implementation.
   * set working directory where the pilot logs etc. are kept
   * make the time to keep logs configurable

I place this for now in Computing/CEDefaults, because the ARC extraString was there too, but maybe this should go to Operations somewhere?

Add documentation page for CEDefaults

### DIFF
--- a/Resources/Computing/HTCondorCEComputingElement.py
+++ b/Resources/Computing/HTCondorCEComputingElement.py
@@ -8,13 +8,28 @@
    Allows direct submission to HTCondorCE Computing Elements with a SiteDirector Agent
    Needs the condor grid middleware (condor_submit, condor_history, condor_q, condor_rm)
 
+   Configuration for the HTCondorCE submission can be done via the configuration system::
+
+     Resources/Computing/CEDefaults/HTCondorCE/WorkingDirectory
+     Resources/Computing/CEDefaults/HTCondorCE/DaysToKeepLogs
+
+   Additional configuration for the submission file can be done per CE, Site or all HTCondorCEs::
+
+     Top priority          : Resources/Sites/<Grid>/<Site>/CEs/<CE>/ExtraSubmitString
+     Second priority       : Resources/Sites/<Grid>/<Site>/ExtraSubmitString
+     Default for HTCondorCE: Resources/Computing/CEDefaults/HTCondorCE/ExtraSubmitString
+
+   It should be of the form::
+
+      request_cpus = 8 \\n periodic_remove = ...
+
 """
 
 import os
 import tempfile
 import commands
 
-from DIRAC                                               import S_OK, S_ERROR
+from DIRAC                                               import S_OK, S_ERROR, gConfig
 from DIRAC.Resources.Computing.ComputingElement          import ComputingElement
 from DIRAC.Core.Utilities.Grid                           import executeGridCommand
 from DIRAC.Core.Utilities.File                           import mkDir
@@ -23,6 +38,8 @@ from DIRAC.WorkloadManagementSystem.DB.PilotAgentsDB     import PilotAgentsDB
 from DIRAC.WorkloadManagementSystem.Agent.SiteDirector   import WAITING_PILOT_STATUS
 from DIRAC.Core.Utilities.File                           import makeGuid
 from DIRAC.Core.Utilities.Subprocess                     import Subprocess
+from DIRAC.Core.Utilities.SiteCEMapping                  import getSiteForCE
+
 
 from DIRAC.Resources.Computing.BatchSystems.Condor import parseCondorStatus, treatCondorHistory
 
@@ -31,6 +48,7 @@ __RCSID__ = "$Id$"
 CE_NAME = 'HTCondorCE'
 MANDATORY_PARAMETERS = [ 'Queue' ]
 DEFAULT_WORKINGDIRECTORY = '/opt/dirac/pro/runit/WorkloadManagement/SiteDirectorHT'
+DEFAULT_DAYSTOKEEPLOGS = 15
 
 def condorIDFromJobRef( jobRef ):
   """return tuple of "jobURL" and condorID from the jobRef string"""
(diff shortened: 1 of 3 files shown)


andresailer:
The extra strings are needed for example to solve this problem with running at UCSD https://ticket.opensciencegrid.org/28335

atsareg:
Why do you want to define extraString on the level of the site configuration ? I guess you have a site with multiple HTCondor CEs which should all share this option. I think still it is better to be defined only on the CE level, this is where it is logically belongs to. Sites can have CEs of different types, but also other resources. It is not appropriate to define particular CE options on this level.

Also, no extra implementation is needed to discover this option.

andresailer:
You mean this option: "Resources/Sites/<Grid>/<Site>/ExtraSubmitString", right?
It would just be really cumbersome to define the same option value 6 times, and then later having to change it. One could change the variablename to be more HTCondorCE specific to avoid confusion.
Anyway the same kind of option exists for the ARCCE, so why differ here?

What do you mean by: "Also, no extra implementation is needed to discover this option." ? Which is "this" option?

atsareg:
I have the same comment for the ARC CE as well, true. I still think that it is better to define on the CE level if even it can be defined several times the same thing. In this case it is logically consistent and CE parameters resolution is done by standard mechanism, you do not need to code __getExtraSubmitString() special method. 

I would suggest the following compromise. Let's introduce a new level of the general CE option definition: /Resources/Sites/<grid>/<site>/CEs/<GeneralCEOption>. It is inside the CEs section, so will not be confused with general site options. Then for each CE options will be resolved in the following sections in the order descending in preference:

/Resources/Computing/CEDefaults
/Resources/Computing/<CEType>
/Resources/Sites/<grid>/<site>/CEs
/Resources/Sites/<grid>/<site>/CEs/<CEName>

  This logic can be coded in the ComputingElement base class (minor changes). Then both ARC and HTCondor CEs are simplified dropping a special option resolution method which is only needed because the site level option is wanted. Comments ?

andresailer:
I think you mean "least to most" preferable, right?
The values inside angle-brackets are not shown, so I am not quite sure which options exactly you are referring to. I guess there should be something following the ending "/"s ? (CEs/...?)

The problem if you drop an option for ARCCE now people potentially have to change their config...

atsareg:
Yes, least to most preferable. So, options defined in these sections are automatically mixed into the CE parameters.

I do not know if anybody using ARCCE is using options on the site level. This is not the LHCb case, for example. But you are right, their config will have to change if we also change the ARC CE implementation. But this we can delay to the v6r16 and make a special announcement.

andresailer:
Ok, so will you change the ComputingElement parts?
Then I can get the extraString/WorkingDirectory/DaysToKeepLogs from the `self.ceParameters` dictionary, right?

atsareg:
Yes, right. See #3056

rajanandakumar:
@atsareg : Will you be updating the ARCComputingElement also?

atsareg:
Yes, this should be done. The only thing is that this can require changes in the configuration of some installations using ARC CE. Therefore we decided to postpone it to v6r16 and make a proper announcement of the change

rajanandakumar:
Indeed - I know that there is at least one installation using it, possibly more.

andresailer:
Adapted based on the changes in #3056

Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -50,8 +63,11 @@ def getCondorLogFile( pilotRef ):
   """return the location of the logFile belonging to the pilot reference"""
   _jobUrl, condorID = condorIDFromJobRef( pilotRef )
   #FIXME: This gets called from the WMSAdministrator, so we don't have the same
-  #working directory as for the SiteDirector unless we force it
-  resLog = findFile( DEFAULT_WORKINGDIRECTORY, '%s.log' % condorID )
+  #working directory as for the SiteDirector unless we force it, there is also
+  #no CE instantiated when this function is called so we can only pick this option up from one place
+  workingDirectory = gConfig.getValue( "Resources/Computing/CEDefaults/HTCondorCE/WorkingDirectory",
```
atsareg: This should be better taken from Resources/Computing/HTCondorCE/WorkingDirectory option. In this case later you can take it from ceParameters as other options

andresailer: Agreed with the location of the option, I didn't read the locations you added carefully enough.

This can only be set in one location, because if the default value were to be overwritten we could not find the log files when we call this function from the WMSAdministrator.

Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -72,15 +88,21 @@ def __init__( self, ceUniqueID ):
     self.outputURL = 'gsiftp://localhost'
     self.gridEnv = ''
     self.proxyRenewal = 0
+    self.extraSubmitString = self.ceParameters.get('ExtraSubmitString', '').decode('string_escape')
+
+    ## see note on getCondorLogFile, why we can only use the global setting
+    self.workingDirectory = gConfig.getValue( "Resources/Computing/HTCondorCE/WorkingDirectory",
```
atsareg: Here you can get the value from self.ceParameters

andresailer: I could, but then how do I make sure that I get the very same working directory when it is needed by the WMSAdministrator?

atsareg: By placing the option in the right place in the Configuration. This is what you have to do anyway.

atsareg:
I merge the PR now, but I still suggest you to think about a common way of getting the WorkingDirectory parameter

Resources/Computing/HTCondorCEComputingElement.py
andresailer: And what is the "right place"?

atsareg: Well, ideally the WMSAdministrator should not care about the CE parameters at all. It should rather use the CE object for all the CE related operations. and the CE object knows about its parameters

andresailer: So in fact you were referring to the problem addressed in the FIXME in getCondorLogFile function. So I can create a (HTCondorCE)ComputingElement instance in the getCondorLogFile function or I leave it to you to change the WMSAdministrator to create ComputingElement instances instead?

atsareg: The pilot logging needs certainly a complete rework. This is what is supposed to have been done in the already committed PRs ( v6r17 ? ). When and if this will be done and will be working, this log extracting mechanism will not be needed any more. So, let's keep it as it s for the moment.

andresailer: I tried instantiating a HTCondorCEComputingElement to get the log file, but the ceParameter locations at the Site and CE level are only resolved in the getQueues function that is called via the SiteDirector, so it doesn't work that way.

PR #3057, opened 2016-08-05
[v6r15]do not use the client for storing host information

zmathe:


andresailer:
This was opened against integration, it should go into the v6r15 branch

FrameworkSystem/Service/SystemAdministratorHandler.py
```diff
@@ -41,8 +41,7 @@ def initializeHandler( cls, serviceInfo ):
     hostMonitoring = cls.srv_getCSOption( 'HostMonitoring', True )
 
     if hostMonitoring:
-      client = SystemAdministratorClient( 'localhost' )
-      gThreadScheduler.addPeriodicTask( 60, client.storeHostInfo )
+      gThreadScheduler.addPeriodicTask( 60, cls.__storeHostInfo )
```
andresailer: This doesn't work unless storeHostInfo is a "@staticmethod"

FrameworkSystem/Service/SystemAdministratorHandler.py
```diff
@@ -633,8 +632,7 @@ def export_getComponentDocumentation( self, cType, system, module ):
     except Exception as _e:
       return S_ERROR( 'No documentation was found' )
 
-  types_storeHostInfo = []
-  def export_storeHostInfo( self ):
+  def __storeHostInfo( self ):
```
andresailer: Needs to be static method and then the self replaced by "SystemAdministratorHandler".
Also the gLogger.error below should get better information about who returned an error.

andresailer: And then of course readHostInfo needs to be staticmethod as well, but that function doesn't use self so it is OK to change that...

FrameworkSystem/Service/SystemAdministratorHandler.py
zmathe: yes. This is what I see.

FrameworkSystem/Service/SystemAdministratorHandler.py
```diff
@@ -632,11 +634,13 @@ def export_getComponentDocumentation( self, cType, system, module ):
     except Exception as _e:
       return S_ERROR( 'No documentation was found' )
 
-  def __storeHostInfo( self ):
+  @staticmethod
+  def __storeHostInfo( selfReference ):
```
andresailer: No argument is needed

FrameworkSystem/Service/SystemAdministratorHandler.py
```diff
@@ -632,11 +634,13 @@ def export_getComponentDocumentation( self, cType, system, module ):
...
     """
     Retrieves and stores into a MySQL database information about the host
+    :param object selfReference: it is the reference of the SystemAdministrator instance.
     """
-    result = self.__readHostInfo()
+    result = selfReference.__readHostInfo()
```
andresailer: This should be `SystemAdministratorHandler.__readHostInfo()`

FrameworkSystem/Service/SystemAdministratorHandler.py
zmathe: S_ERROR is fine... The error is properly sent....

andresailer: The output from
```
gLogger.error( result['Message'] )
```
is really confusing because it is not obvious where the error message comes from, e.g., the ComponentMonitoring

Also
it should be
```
return result
```
instead of creating a new S_ERROR structure.

zmathe: result contains the callstack as well. The gLogger.error does not print the callstack. and in the code it is clear where the error comes from

andresailer: In the log it just said "Unauthorized access", that is not clear at all.

zmathe: Yes, because the message is coming from the Framework, and it is not using the new S_ERROR structure.

zmathe:
Wrong branch

PR #3060, opened 2016-08-08
[v6r15] FileCatalog: fix expected type for getAncestor, getDescendents,

andresailer:
First argument is always turned into dict by checkCatalogArgument decorator in the client


DataManagementSystem/Service/FileCatalogHandler.py
```diff
@@ -323,7 +323,7 @@ def export_getReplicaStatus( self, lfns ):
     """ Get the status for the supplied replicas """
     return gFileCatalogDB.getReplicaStatus( lfns, self.getRemoteCredentials() )
 
-  types_getFileAncestors = [ ListType, [ ListType, IntType, LongType ] ]
+  types_getFileAncestors = [ [ ListType, DictType ], [ ListType, IntType, LongType ] ]
```
chaen: as of v6r15 we can use directly ```[list,dict]``` etc

atsareg:
Review OK

PR #3069, opened 2016-08-12
[v6r15] JobCleaningAgent: Option for removing Jobs in other states

marianne013:
This update adds an option (disabled by default) to remove Jobs from the dirac server irrespective of their state.
We found this useful for cleaning a large backlog of jobs that had been in the Waiting state for 3+ months.
Also updated the default polling time for this agent, as Job ages are given in days and a default polling time of 120 s seems excessive.
Updated the documentation to list all options for this agent.


atsareg:
Review OK

PR #3070, opened 2016-08-12
[v6r15] HostLogging: Bring back version and setup information

andresailer:
The `infoResult['Value']` contains the DIRAC version and Setup information which needs to be passed forward as well


atsareg:
Review OK

PR #3073, opened 2016-08-16
[v6r15] ValidateOutputDataAgent: use the shifter proxy

chaen:
Gone for almost 2 years.. this agent really needs to be reviewed because it does nothing


TransformationSystem/Agent/ValidateOutputDataAgent.py
```diff
@@ -47,6 +47,12 @@ def __init__( self, *args, **kwargs ):
   def initialize( self ):
     """ Sets defaults
     """
+
+    # This sets the Default Proxy to used as that defined under
```
atsareg: to be used

PR #3075, opened 2016-08-16
DataManager:fix an UnboundLocal Error for putAndRegister

cj501885963:
DataManager.putAndRegister : An UnboundLocal Error may be thrown because of the unreasonable location of variable ‘errStr’.


atsareg:
The fix is relevant, but please do the PR against the rel-v6r15 branch

PR #3084, opened 2016-08-23
[v6r15] SE: error in the accounting for putFile

chaen:
Logic error


phicharp:
Fed was too quick for me ;-)

PR #3089, opened 2016-08-25
[v6r15] Renew ARC pilots in the queued state.

sfayer:
This patch ensures that pilot jobs that are queued also get their proxies renewed on ARC-CE.

Regards,
Simon


atsareg:
Review OK

PR #3090, opened 2016-08-25
[v6r15] Miscellaneous small fixes

phicharp:
* Remove temporary fix in TransformationAgent (that had been there for 2 years ;-)
* Cache list of SEs in DMSHelpers
* Fix helper getStorageElementOptions() for handling BaseSE and Alias. Note: this is redundant with (part of) the StorageFactory and was causing troubles when introducing BaseSE earlier. It is only used by RSS

### DIFF
--- a/DataManagementSystem/scripts/dirac-dms-find-lfns.py
+++ b/DataManagementSystem/scripts/dirac-dms-find-lfns.py
@@ -10,6 +10,7 @@ if __name__ == "__main__":
 
   from DIRAC.Core.Base import Script
 
+  Script.registerSwitch( '', 'Path=', '    Path to search for' )
   Script.setUsageMessage( '\n'.join( [ __doc__.split( '\n' )[1],
                                        'Usage:',
                                        '  %s [options] metaspec [metaspec ...]' % Script.scriptName,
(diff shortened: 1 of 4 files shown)


chaen:
agaisnt integration ?

TransformationSystem/Agent/TransformationAgent.py
```diff
@@ -87,7 +87,7 @@ def initialize( self ):
 
     # for caching using a pickle file
     self.workDirectory = self.am_getWorkDirectory()
-    self.cacheFile = os.path.join( self.workDirectory, 'ReplicaCache.pkl' )
+    self.cacheFile = os.path.join( self.workDirectory, 'ReplicaCache_%s.pkl' % str() )
```
andresailer: Why the `% str()` ?

phicharp:
Oups, thanks André!

TransformationSystem/Agent/TransformationAgent.py
```diff
@@ -603,20 +603,14 @@ def __readCache( self, transID ):
...
-        self.replicaCache[transID] = cache.get( transID, {} )
+        self.replicaCache[transID] = {}
       else:
         cacheFile = open( fileName, 'r' )
         self.replicaCache[transID] = pickle.load( cacheFile )
-      cacheFile.close()
-      self._logInfo( "Successfully loaded replica cache from file %s (%d files)" %
-                     ( fileName, self.__filesInCache( transID ) ),
-                     method = method, transID = transID )
+        cacheFile.close()
```
chaen: use the ```with``` statement ?

chaen:
review OK

chaen:
re OK

DataManagementSystem/scripts/dirac-dms-find-lfns.py
```diff
@@ -10,6 +10,7 @@
 
   from DIRAC.Core.Base import Script
 
+  Script.registerSwitch( '', 'Path=', '    Path to search for' )
```
atsareg: In principle this is not necessary because Path is a valid metadata key. Historically this was not the case, so it is a separate argument in the call now. I wanted to get rid of it in fact. You can simply do:
    dirac-dms-find-lfns Path=/lhcb/user/a/atsareg SE=IN2P3-disk

PR #3091, opened 2016-08-25
[v6r15][IMPORTANT] FTS: use requests library

chaen:
use requests with no server verification, and add temporary fix for fts-rest

The use of requests is needed because of the memory leak and on top of this we are now for some reason hit by another curl bug (https://curl.haxx.se/mail/lib-2008-09/0197.html)

There is however a bug in the fts-rest library which does not propagate the server verification flag. I have made a PR, but until they merge it and release it (3.5.2, a month from now) I have added the fix in DIRAC (ftsSSLWrapper). As soon as we have the proper release in the bundle, we can remove this file.

This is an important PR, it should be in the next patch, it runs as hotfix in LHCb


phicharp:
Review OK

phicharp:
Review OK again

PR #3092, opened 2016-08-25
[v6r15] directory-sync: fixes for missing import, KeyError

andresailer:


DataManagementSystem/scripts/dirac-dms-directory-sync.py
```diff
@@ -111,7 +112,7 @@ def getSetOfRemoteSubDirectoriesAndFiles(path,fc,directories,files):
           return S_ERROR('Error: ' + res['Message'])
       return S_OK()
     else:
-      return S_ERROR("Error:" + result['Message'])
+      return S_ERROR("Error: %s" % result['Value'])
```
phicharp: These series of S_ERROR in a recursive method may end up with lots of "Error: " piling up, right?

In case of error, why not just `return result`

DataManagementSystem/scripts/dirac-dms-directory-sync.py
```diff
@@ -111,7 +112,7 @@ def getSetOfRemoteSubDirectoriesAndFiles(path,fc,directories,files):
...
   else:
     return S_ERROR("Error:" + result['Message'])
```
phicharp: `return result` for example here

DataManagementSystem/scripts/dirac-dms-directory-sync.py
andresailer: Well, the bug here was that the result is not S_ERROR at this point and therefore there is no `Message`. Doing `return result` would not be good in this case.

atsareg:
I agree with the last Philippe's comment, please update

andresailer:
If you feel strongly about the returns in this script please create an issue and assign it to Marko, meanwhile, please merge this PR to fix the bug.

PR #3094, opened 2016-08-26
[v6r15] Fix proxy renewal (for ARC-CE)

sfayer:
This patch ensures that a proxy of at least 3 hours is available to the updatePilotStatus function so that if it renews any proxies (via ce.getJobStatus), it's not renewing them with a very short proxy (which causes expiry problems on the ARC-CE).

### DIFF
--- a/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -531,7 +531,7 @@ class SiteDirector( AgentModule ):
       if not result['OK']:
         return result
       self.proxy = result['Value']
-      ce.setProxy( self.proxy, cpuTime - 60 )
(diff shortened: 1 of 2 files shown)


WorkloadManagementSystem/Agent/SiteDirector.py
```diff
@@ -531,7 +531,7 @@ def submitJobs( self ):
       if not result['OK']:
         return result
       self.proxy = result['Value']
-      ce.setProxy( self.proxy, cpuTime - 60 )
+      ce.setProxy( self.proxy, self.proxy.getRemainingSecs()['Value'] )
```
atsareg: It is generally not a good practice to not check the return status of the method. If even it can never return an error, the fact that the developer returns an S_OK assumes that eventually the implementation can be more sophisticated and may return S_ERROR as well

PR #3095, opened 2016-08-30
[v6r15] More options for the VOMS2CSAgent

atsareg:
Added more options for the VOMS2CSAgent:

- Possibility to automatically delete users who are no longer in the VOMS ( AutoDeleteUsers CS flag )
- Possibility to create user home directories in DFC ( MakeHomeDirectory CS flag )
- Possibility to generate more detailed report for the admins ( DetailedReport CS flag )

Plus several fixes in the logic of the agent.

### DIFF
--- a/ConfigurationSystem/Agent/VOMS2CSAgent.py
+++ b/ConfigurationSystem/Agent/VOMS2CSAgent.py
@@ -17,6 +17,7 @@ from DIRAC.ConfigurationSystem.Client.Helpers.Registry import getVOOption, getUs
 from DIRAC.Core.Utilities.Proxy                        import executeWithUserProxy
 from DIRAC.Core.Utilities.List                         import fromChar
 from DIRAC.Core.Utilities.PrettyPrint                  import printTable
+from DIRAC.Resources.Catalog.FileCatalog               import FileCatalog
 
 __RCSID__ = "$Id$"
 
@@ -37,6 +38,7 @@ class VOMS2CSAgent( AgentModule ):
     self.autoModifyUsers = False
     self.autoDeleteUsers = False
     self.detailedReport = True
+    self.makeFCEntry = False
 
   def initialize( self ):
     """ Initialize the default parameters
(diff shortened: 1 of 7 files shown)


andresailer:
Don't forget to update the documentation: http://dirac.readthedocs.io/en/latest/AdministratorGuide/Configuration/ConfReference/Systems/Configuration/Agents/VOMS2CSAgent/index.html

ConfigurationSystem/Agent/VOMS2CSAgent.py
```diff
@@ -75,37 +83,143 @@ def execute( self ):
...
+
+    diracVOMSMapping = result['Value']['DIRACVOMS']
+
+    records = []
+    groupDict = {}
+    multiDNUsers = {}
+    suspendedUsers = []
+    for user in userDict:
+      for group in userDict[user]['Groups']:
+        groupDict.setdefault( group, 0 )
```
andresailer: One could also use `from collections import defaultdict` for groupDict
https://docs.python.org/2/library/collections.html#collections.defaultdict

closier:
It seems that the options AutoAddVOMS is not take into account. which means that any group where a role is defined is filled even if the option AutoAddVOMS=False

closier:
It seems that the case where a "user" is defined in DIRAC CS but is not present in VOMS is not take into account. so if the VO is VOMS aware, we should check in both direction (VOMS - Cs but also CS - VOMS)

atsareg:
AudoAddVOMS is an option that can be defined for a group which defines if VOMS extension should be added to a proxy generated with dirac-proxy-init ( or equivalent ) command. This is not related to the work of the VOMS2CSAgent. If you want some DIRAC groups mapped on some VOMS roles not to be updated with users having this role, this is to be discussed whether this is a real requirement.

chaen:
Pelase add the [WIP] flag if you are still working on it, so we know we should not review yet

atsareg:
The module is ready for review. Actually, last modifs are responses to already made remarks and also a result of testing against the LHCb case

PR #3096, opened 2016-08-31
[v6r15] Add getFileDescendents/Ancestors to READ methods for FileCatalog

andresailer:
There should be some way to ensure that all functions are in the lists...
( This is not part of #3095, https://github.com/DIRACGrid/DIRAC/pull/3095/files#diff-38faa43f37e89cea3917fb128720076dR33 )


chaen:
Review OK, and agree with you

PR #3098, opened 2016-09-02
[v6r15] Changes to using index in accounting to speed-up removal query

ltomassetti:
now the delete query condition is:
startTime < _UNIX_TIMESTAMP() - 14months - maxBucketLength


phicharp:
Review OK

PR #3099, opened 2016-09-05
[v6r16] Monitoring the space left on DIRAC SEs

vipersec:
I have added a new command named `FreeDiskSpace` which is used to get the total and the remaining disk space of all dirac storage elements that are found in the CS and inserts the results in the `SpaceTokenOccupancyCache` table of `ResourceManagementDB` database.


DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +50,48 @@
...
+    elif size.upper() == "TB":
+      convert = 1.099511627776e12
+    else:
+      return S_ERROR( "No valid size specified" )
+
+    st = os.statvfs(path)
+    free = (st.f_bavail * st.f_frsize) / convert
+    return S_OK( round(free, 2) )
+
+def totalSpace(path, size = 'TB'):
```
chaen: You can most certainly factorize diskSpace and totalSpace.

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -176,6 +218,20 @@ def export_getMetadata( self, fileID ):
     """
     return self.__getFileStat( self.__resolveFileID( fileID ) )
 
+  types_getRemainingDiskSpace = [StringTypes]
```
chaen: Replace ```StringTypes``` with ```basestring```

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -176,6 +218,20 @@ def export_getMetadata( self, fileID ):
...
+  def export_getRemainingDiskSpace( self, path, size = 'TB' ):
+    """ Get the remaining disk space of the storage element
+        If no size is specified, terabytes will be used by default.
+    """
+    return diskSpace(path, size)
+
+  types_getTotalDiskSpace = [StringTypes]
+  def export_getTotalDiskSpace( self, path, size = 'TB' ):
```
chaen: Replace ```StringTypes``` with ```basestring```

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +50,48 @@
...
+    elif size.upper() == "MB":
+      convert = 1.048576e6
+    elif size.upper() == "GB":
+      convert = 1.073741824e9
+    elif size.upper() == "TB":
+      convert = 1.099511627776e12
+    else:
+      return S_ERROR( "No valid size specified" )
+
+    st = os.statvfs(path)
```
chaen: Exceptions might be thrown no ?

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -176,6 +218,20 @@ def export_getMetadata( self, fileID ):
     """
     return self.__getFileStat( self.__resolveFileID( fileID ) )
 
+  types_getRemainingDiskSpace = [StringTypes]
+  def export_getRemainingDiskSpace( self, path, size = 'TB' ):
```
chaen: Detail, but 'free' is probably mroe common than 'remaining'

ResourceStatusSystem/Command/RemainingDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+
+from datetime                                                   import datetime
+from DIRAC                                                      import S_OK, S_ERROR, gConfig
+from DIRAC.ResourceStatusSystem.Command.Command                 import Command
+from DIRAC.Core.DISET.RPCClient                                 import RPCClient
+
+__RCSID__ = '$Id:  $'
+
+
+class RemainingDiskSpaceCommand( Command ):
```
chaen: Same general remark concerning 'free' rather than 'remaining'

ResourceStatusSystem/Command/RemainingDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+
+    :param SE: String
+    :param protocol: String
+    :return: String
+    """
+    attributes = [ "Protocol", "Host", "Port", "Path"]
+
+    result = ""
+    for attribute in attributes:
+      res = gConfig.getValue( "/Resources/StorageElements/%s/AccessProtocol.1/%s" % ( SE, attribute ) )
```
chaen: This is not good because you make assumption as of the section name ```AccessProtocol.1```.  Maybe you should rather use the storageElementFactory rather than gConfig to get what you want.

ResourceStatusSystem/Command/RemainingDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+  '''
+
+  def __init__( self, args = None ):
+
+    super( RemainingDiskSpaceCommand, self ).__init__( args )
+
+    self.rpc = None
+    self.rsClient = None
+
+  def getUrl(self, SE, protocol = None ):
```
chaen: instantiate the SE and call getBasePath (or something very simmilar, I don't remember exactly)

DataManagementSystem/Service/StorageElementHandler.py
vipersec: I added this https://github.com/DIRACGrid/DIRAC/pull/3099/commits/7892d014b94b2bc39871d9e20a36d2ec06043f09 Is it enough? Or should I use a Try-Catch statement?

chaen: you need a try catch for all possible exception that might arrise. I am pretty sure there are OSError if the path does not exist for example

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +50,39 @@
...
+      st = os.statvfs(path)
+
+      if total:
+        # return total space
+        result = (st.f_blocks * st.f_frsize) / convert
+      else:
+        # return free space
+        result = (st.f_bavail * st.f_frsize) / convert
+
+    except Exception as e:
```
chaen: check the doc, but I think it is enough to catch OSError

vipersec: `Exception` catches all exceptions including `OSError` as far as I know :)

chaen: of course it does, but it is a bad practice to use a catch all if you can have a finer grain selection :) Pylint would tell you as well ;-)

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +50,39 @@
 MAX_STORAGE_SIZE = 0
 USE_TOKENS = False
 
+def diskSpace(path, size = 'TB', total = False):
```
atsareg: According to DIRAC coding conventions, method name should start with a verb, e.g. getDiskSpace()

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+  '''
+
+  def __init__( self, args = None ):
+
+    super( FreeDiskSpaceCommand, self ).__init__( args )
+
+    self.rpc = None
+    self.rsClient = None
+
+  def getUrl(self, SE, protocol = None ):
```
chaen: Seems like my previous comment was lost, but please don't use directly gConfig for that, use the StorageFactory, and don't construct the URLs yourself but use the getBaseURL method of the SE plugins

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+
+    urls = {}
+    for SE in SEs['Value']:
+      res = self.getUrl( SE, protocol )
+      if res:
+        urls.update( {SE: res} )
+
+    return urls
+
+  def doCommand( self ):
```
fstagni: Hi Steve,
this is not the way you code a command in DIRAC RSS: 
- a command should work on one element only
- this element is in the `params`
- the command should then be called from the CacheFeederAgent

fstagni: And you should provide a `doNew`, `doMaster`, `doCache` methods

chaen: Just in case there is no, it would be a good opportunity to write a doc about it.

vipersec:
So, in https://github.com/DIRACGrid/DIRAC/pull/3099/commits/fac178313573f3cd58d81d47b437b00e2ccdbcbb I added `doNew`, `doMaster`, `doCache` methods and changed `doNew` to work on one element only.

Is it ok to have a `for` loop to get all the storage element names in `doMaster`?

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,115 @@
...
+
+    :param SE: String
+    :param protocol: String
+    :return: String
+    """
+    attributes = [ "Protocol", "Host", "Port", "Path"]
+
+    result = ""
+    for attribute in attributes:
+      res = gConfig.getValue( "/Resources/StorageElements/%s/AccessProtocol.1/%s" % ( SE, attribute ) )
```
fstagni: Please follow the advice that Chris gave you few lines above

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,115 @@
...
+  def _prepareCommand( self ):
+    '''
+      FreeDiskSpaceCommand requires one argument:
+      - element : Resource
+
+    '''
+
+    if 'element' not in self.args:
+      return S_ERROR( '"element" not found in self.args' )
+    element = self.args[ 'element' ]
```
fstagni: `element` is in RSS usally either "Site" or "Resource". I think what you are looking for here is `name` (the name of the resource)

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,115 @@
...
+
+    return S_OK( element )
+
+  def doNew( self, masterParams = None ):
+    """
+    Gets the total and the free disk space of all DIPS storage elements that
```
fstagni: nope, just one

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,115 @@
...
+
+    self.rpc = RPCClient( elementURL, timeout=120 )
+    free = self.rpc.getFreeDiskSpace("/")
+    total = self.rpc.getTotalDiskSpace("/")
+    self.rsClient.addOrModifySpaceTokenOccupancyCache(endpoint = elementURL, lastCheckTime = datetime.utcnow(),
+                                                      free = free, total = total, token = element )
+
+    return S_OK()
+
+  def doCache( self ):
```
fstagni: This should be implemented to get it from the spaceTokenOccupancy cache.

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,115 @@
...
+    return S_OK()
+
+  def doMaster( self ):
```
fstagni: add a docstring

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +50,39 @@
...
+
+      if total:
+        # return total space
+        result = (st.f_blocks * st.f_frsize) / convert
+      else:
+        # return free space
+        result = (st.f_bavail * st.f_frsize) / convert
+
+    except OSError as e:
+      return S_ERROR( "Error %s", e )
```
fstagni: do instead
```
return S_ERROR( errno.EIO, "Error while getting the available disk space: %s", repr(e) )
```

fstagni:
BTW, it is ok to have a for loop to get all the storage element names in `doMaster`. 

And please add some doc.

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -95,15 +95,34 @@ def doNew( self, masterParams = None ):
     self.rpc = RPCClient( elementURL, timeout=120 )
     free = self.rpc.getFreeDiskSpace("/")
     total = self.rpc.getTotalDiskSpace("/")
-    self.rsClient.addOrModifySpaceTokenOccupancyCache(endpoint = elementURL, lastCheckTime = datetime.utcnow(),
-                                                      free = free, total = total, token = element )
+    result = self.rsClient.addOrModifySpaceTokenOccupancyCache(endpoint = elementURL, lastCheckTime = datetime.utcnow(),
+                                                               free = free, total = total, token = elementName )
+    if not result[ 'OK' ]:
+      return S_ERROR( DErrno.EMYSQL, "Query failed %s" % result )
```
fstagni: `result['Message']`

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
chaen: Once  merged https://github.com/DIRACGrid/DIRAC/pull/3123

```
In [4]: se = StorageElement(SE)
In [5]: se.getStorageParameters(protocol = protocol)
{'OK': True,
 'Value': {'Host': 'lhcb-se-dirac.cern.ch',
  'Path': '/DataManagement/StorageElement',
  'PluginName': 'DIP',
  'Port': '9148',
  'Protocol': 'dips',
  'SpaceToken': '',
  'StorageName': 'LogSE',
  'URLBase': 'dips://lhcb-se-dirac.cern.ch:9148/DataManagement/StorageElement',
  'WSUrl': ''}}
```

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
chaen: no,
```
if not result['OK']:
    return result
```
Let the error propagate

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+
+    elementName = self._prepareCommand()
+    if not elementName[ 'OK' ]:
+      return elementName
+
+    result = self.rsClient.selectSpaceTokenOccupancyCache(token = elementName)
+
+    if not result[ 'OK' ]:
+      return S_ERROR( DErrno.EMYSQL, "Query failed %s" % result['Message'] )
```
chaen: Same here

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +51,39 @@
...
 USE_TOKENS = False
 
+def getDiskSpace(path, size = 'TB', total = False):
+    """
+      Returns disk usage of the given path.
+      If no size is specified, terabytes will be used by default.
+      If total is set to true, the total disk space will be returned instead.
+    """
+
+    if size.upper() == "KB":
```
JanEbbing: I would replace this with something like
```python
UNIT_CONVERSION = { "KB" : 1024, "MB" : 1.048576e6, ... }
def getDiskSpace(...):
  size_to_convert = size.upper()
  if size_to_convert not in UNIT_CONVERSION:
    return S_ERROR( "no valid size" )
  convert = UNIT_CONVERSION[size_to_convert]
```

This is shorter, imo easier to understand, debug/maintain and in general better coding style. There's probably even better solutions (I don't like the total flag) but that should be ok.

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +51,39 @@
...
+      convert = 1.073741824e9
+    elif size.upper() == "TB":
+      convert = 1.099511627776e12
+    else:
+      return S_ERROR( "No valid size specified" )
+
+    try:
+      st = os.statvfs(path)
+
+      if total:
```
JanEbbing: Also here, slightly better imo:
```python
if total:
  queried_size = st.f_blocks
else:
  queried_size = st.f_bavail
result =  ( queried_size * st.f_frsize ) / convert
```

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -0,0 +1,102 @@
...
+    se = StorageElement(elementName)
+
+    elementURL = se.getStorageParameters(protocol = "dips")['URLBase']
+
+    if not elementURL:
+      gLogger.info( "Not a DIPS storage element, skipping..." )
+      return S_OK()
+
+    self.rpc = RPCClient( elementURL, timeout=120 )
+    free = self.rpc.getFreeDiskSpace("/")
```
JanEbbing: Might be (probably is) premature optimization, but do these queries take long?
You could split `getDiskSpace` even more so that you have (private) methods that take the result of a `os.statvfs` and return the free or the total space of a disk. Then 3 public methods `getFreeDiskSpace`, `getTotalDiskSpace` and `getDiskInfo` which returns a tuple ( freeSpace, totalSpace ) (would violate SRP but I could imagine a `os.stat('/')` to run quite a while).

DataManagementSystem/Service/StorageElementHandler.py
phicharp: and it is also much better to not define the units in floating point but as integers!

MB: 1024*1024
GB: 1024*1024*1024
etc...

phicharp:
Not related to this PR: I find the new way of commenting a PR absolutely tedious... It is not clear at all when a notification is sent to the participants, and since notifications are not sent to the author of it there is not way to check...
Anyway probably useless to say that here as nobody will be notified...
Can you please confirm you received something? ;-)

phicharp:
@vipersec : although approved can you take Jan's comments into account (and mine occasionally)

sposs:
I do receive all comments.

DataManagementSystem/Service/StorageElementHandler.py
```diff
@@ -50,6 +51,39 @@
...
+        # return total space
+        result = (st.f_blocks * st.f_frsize) / convert
+      else:
+        # return free space
+        result = (st.f_bavail * st.f_frsize) / convert
+
+    except OSError as e:
+      return S_ERROR( errno.EIO, "Error while getting the available disk space: %s", repr(e) )
```
andresailer: There should be a "%" instead of the comma after the second argument.

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
vipersec: The queries do not take long normally as far as I know. I will stick to the current solution for the time being.

DataManagementSystem/Service/StorageElementHandler.py
vipersec: Done and done :)

PR #3101, opened 2016-09-07
[v6r15] Miscellaneous small fixes

phicharp:
* Fix potential problem with JOBFEATURES (if the information is zero or absent we accept it)


atsareg:
Review OK

PR #3106, opened 2016-09-14
[v6r16] DynamicMonitoring Profiler

zmathe:
Added profiling utility used to retrieve dynamic information about a process. This is used by the dynamic monitoring system to get the information that will be stored into ElasticSearch.

### DIFF
--- a/Core/Utilities/Profiler.py
+++ b/Core/Utilities/Profiler.py
@@ -1,0 +1,122 @@
+"""
+Profiling class for updated information on process status
+"""
+
+import psutil
+import datetime
+from DIRAC import S_OK, S_ERROR
+
+class Profiler:
+  """
+  Class for profiling both general stats about a machine and individual processes.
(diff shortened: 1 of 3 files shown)


Core/Utilities/Profiler.py
```diff
@@ -0,0 +1,205 @@
...
+
+  def runningTime( self ):
+    """
+    Returns the uptime of the process
+    """
+    if self.process:
+      try:
+        start = datetime.datetime.fromtimestamp( self.process.create_time() )
+        result = ( datetime.datetime.now() - start ).total_seconds()
+      except psutil.NoSuchProcess as e:
```
sposs: A little decorator somewhere would save all the copy-paste...

zmathe: I have more problem with non tested code. I am not going to spend time to make decorator. Feel free and do it

chaen:
And it will wait forever when this gets merged.

PR #3109, opened 2016-09-19
Return error when pilot output is missing

alexanderrichards:
This patch hopefully fixes https://github.com/ic-hep/DIRAC/issues/64


alexanderrichards:
sorry was meant to push this against the ic-hep Dirac

PR #3125, opened 2016-09-29
[v6r15] FIX FTS: specify checksum type at submission

chaen:
From FTS server 3.5, the behavior changed with respect to the checksum type. To avoid that, always specify it


phicharp:
reviewed but not allowed as I don't have the karma...

chaen:
It seems only people with Write access to the repository can approve PR. This is not really nice...

phicharp:
Good: only Andrei and Federico can review code! Thanks guys!

PR #3126, opened 2016-09-29
[v6r15] Fix StorageManagementClient

phicharp:
If there are no active replicas, avoid an exception

Dummy change in CPUNormalisation

### DIFF
--- a/StorageManagementSystem/Client/StorageManagerClient.py
+++ b/StorageManagementSystem/Client/StorageManagerClient.py
@@ -64,12 +64,12 @@ def getFilesToStage( lfnList ):
     return S_ERROR( "Could not get metadata for files" )
   offlineLFNs = set( lfnList ) - onlineLFNs
 
-
   for offlineLFN in offlineLFNs:
     ses = lfnListReplicas['Value']['Successful'][offlineLFN].keys()
-    random.shuffle( ses )
-    se = ses[0]
-    offlineLFNsDict.setdefault( se, list() ).append( offlineLFN )
+    if ses:
+      random.shuffle( ses )
+      se = ses[0]
+      offlineLFNsDict.setdefault( se, list() ).append( offlineLFN )
 
   return S_OK( {'onlineLFNs':list( onlineLFNs ), 'offlineLFNs': offlineLFNsDict} )
 
--- a/StorageManagementSystem/Client/StorageManagerClient.py
+++ b/StorageManagementSystem/Client/StorageManagerClient.py
@@ -67,9 +67,7 @@ def getFilesToStage( lfnList ):
   for offlineLFN in offlineLFNs:
     ses = lfnListReplicas['Value']['Successful'][offlineLFN].keys()
     if ses:
-      random.shuffle( ses )
-      se = ses[0]
-      offlineLFNsDict.setdefault( se, list() ).append( offlineLFN )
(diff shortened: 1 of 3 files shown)


WorkloadManagementSystem/DB/JobDB.py
```diff
@@ -277,8 +277,10 @@ def getJobParameters( self, jobID, paramList = None ):
         if not ret['OK']:
           return ret
         paramNameList.append( ret['Value'] )
-      paramNames = ','.join( paramNameList )
-      cmd = "SELECT Name, Value from JobParameters WHERE JobID=%s and Name in (%s)" % ( e_jobID, paramNames )
+      if len( paramNameList ) == 1:
```
chaen: So I checked, you don't need a specific case for one element, the ```in``` clause is optimzied:
```
MySQL [FileCatalogDB]> explain select * from FC_StorageElements where SEName in ('SARA-USER');
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
| id | select_type | table              | type  | possible_keys | key    | key_len | ref   | rows | Extra |
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
|  1 | SIMPLE      | FC_StorageElements | const | SEName        | SEName | 129     | const |    1 | NULL  |
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
1 row in set (0.01 sec)

MySQL [FileCatalogDB]> explain select * from FC_StorageElements where SEName = 'SARA-USER';
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
| id | select_type | table              | type  | possible_keys | key    | key_len | ref   | rows | Extra |
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
|  1 | SIMPLE      | FC_StorageElements | const | SEName        | SEName | 129     | const |    1 | NULL  |
+----+-------------+--------------------+-------+---------------+--------+---------+-------+------+-------+
1 row in set (0.00 sec)

MySQL [FileCatalogDB]> explain select * from FC_StorageElements where SEName in ('SARA-USER', 'SARA-DST');
+----+-------------+--------------------+-------+---------------+--------+---------+------+------+-----------------------+
| id | select_type | table              | type  | possible_keys | key    | key_len | ref  | rows | Extra                 |
+----+-------------+--------------------+-------+---------------+--------+---------+------+------+-----------------------+
|  1 | SIMPLE      | FC_StorageElements | range | SEName        | SEName | 129     | NULL |    2 | Using index condition |
+----+-------------+--------------------+-------+---------------+--------+---------+------+------+-----------------------+
1 row in set (0.00 sec)
```

chaen:
Nop, can't approve either

phicharp:
nope

StorageManagementSystem/Client/StorageManagerClient.py
```diff
@@ -64,12 +64,12 @@ def getFilesToStage( lfnList ):
...
   offlineLFNs = set( lfnList ) - onlineLFNs
 
-
   for offlineLFN in offlineLFNs:
     ses = lfnListReplicas['Value']['Successful'][offlineLFN].keys()
-    random.shuffle( ses )
-    se = ses[0]
-    offlineLFNsDict.setdefault( se, list() ).append( offlineLFN )
+    if ses:
+      random.shuffle( ses )
```
JanEbbing: Why not `random.choice` ?

PR #3127, opened 2016-09-30
[v6r15] Minor fixes

phicharp:
* When updating input files status, if OK, clear the dictionary
    - This was forcing to do it twice
* In RemoveFile: remove all files that are not at a banned SE


chaen:
review OK

PR #3141, opened 2016-10-20
[v6r16] remove User from the Job Policy

zmathe:


atsareg:
Why is this change ?

zmathe:
This is because if a user using a certificate, the selector in the Accounting application will be not filled. However, if the user uses http (no certificate), the selectors will be filled. So in that case non authenticated users (http) have more rights, than the authenticated users...
There is another security which is at the plot creation level. So we check again the properties of the users. So If the user does not have the property the plot will not created.

zmathe:
I want to say if the user uses Job accounting and try to group by User, the plot will be empty as the user do not have the right to create the plot.

PR #3159, opened 2016-11-02
[v6r16] Various fixes

phicharp:
* dirac-dms-find-lfns was not working: make it partly work. However the quoted example doesn't work (filter CreationDate<2015-05-15 doesn't work)

* StalledJobAgent: send a kill signal to the job before setting it Failed. This should prevent jobs to continue running after they have been found Stalled and then Failed.
https://its.cern.ch/jira/browse/LHCBDIRAC-554

* Avoid space at the end of a gLogger message when the variable message is void

### DIFF
--- a/DataManagementSystem/scripts/dirac-dms-find-lfns.py
+++ b/DataManagementSystem/scripts/dirac-dms-find-lfns.py
@@ -33,11 +33,6 @@ if __name__ == "__main__":
     if opt == 'Path':
       path = val
 
-  if len( args ) < 1:
-    print "Error: No argument provided\n%s:" % Script.scriptName
-    Script.showHelp()
-    DIRAC.exit( -1 )
-
   fc = FileCatalog()
   result = fc.getMetadataFields()
   if not result['OK']:
@@ -47,8 +42,13 @@ if __name__ == "__main__":
   typeDict.update( result['Value']['DirectoryMetaFields'] )
   # Special meta tags
   typeDict.update( FILE_STANDARD_METAKEYS )
-  
-  gLogger.info( "MetaDataDictionary: %s" % metaDict )
+
+  if len( args ) < 1:
+    print "Error: No argument provided\n%s:" % Script.scriptName
+    Script.showHelp()
+    gLogger.notice( "MetaDataDictionary: \n%s" % str( typeDict ) )
+    DIRAC.exit( -1 )
+
 
   mq = MetaQuery( typeDict = typeDict )
   result = mq.setMetaQuery( args )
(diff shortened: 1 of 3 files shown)


DataManagementSystem/scripts/dirac-dms-find-lfns.py
```diff
@@ -62,7 +62,6 @@
   if not result['OK']:
     gLogger.error( 'Can not access File Catalog:', result['Message'] )
     DIRAC.exit( -1 )
-  lfnList = result['Value']
+  lfnList = sorted( result['Value'] )
 
-  for lfn in lfnList:
-    print lfn
+  gLogger.notice( '\n'.join( lfn for lfn in lfnList ) )
```
phicharp: Sorry, I didn't remember this...
B.t.w. why was the comment removed that you had put in that PR? Why not fix gLogger ;-)?

phicharp:
Still if someone could tell me why this doesn't work:

dirac-dms-find-lfns --Path /lhcb/debug/2012/SIM/00054394/0004 "SE=CERN-DEBUG" "CreationDate<2016-11-02"
Can not access File Catalog: FileCatalog error ( 1604 : Failed to perform findFilesByMetadata from any catalog)

as this is the example given in the help!

DataManagementSystem/scripts/dirac-dms-find-lfns.py
andresailer: Thanks, that is even better.
Can we get the dirac-dms-find-lfns fix into v6r15 as well? #3132

atsareg:
The problem with the CreationDate tag in the query is that the FileCatalog DB schema used by LHCb ( with stored procedures ) is not compatible with the original db schema, the table FC-FileInfo was dropped and merged with FC_Files. This is not trivial to fix. I would prefer that the LHCb's schema is made compatible, which means fixes also in the corresponding plugin code

phicharp:
OK, not really a show-stopper for LHCb, just I wanted to test that the example was working... and also use it for cleaning up the CERN-DEBUG storage, but I found another way of doing it using dirac-dms-list-directory that has a date limit...

chaen:
This is targeted at integration branch.

@atsareg can you be more specific about what is the problem with our managers ?

atsareg:
In the LHCb FileManager the CreationDate is in FC_Files table, whereas originally it was in FC-FileInfo. This is where it is failing now. Maybe this is not the only problem since the new schema was introduced essentially as a "monolitic" one as we have already discussed in the past.

PR #3162, opened 2016-11-05
[] Consumer scheme and CS synchronization for PilotLoggers

wkrzemien:
This is essentially rebased and fixed content of the PR #2976 All tests are passing. 
Probably should be cross-checked with   #3160 

1.Basic consumer structures to DIRAC
ConsumerModule - interface for all consumers
RabbitMQ.py module developed by Sergio Balbuena Pantigas is beeing used
2.CS RabbitMQ internal DB synchronization
RabbitMQAdmin is the module that encloses all administration task related to RabbitMQ server e.g. updating internal database.


wkrzemien:
This PR becomes  obsolete and is replaced by #3181

PR #3166, opened 2016-11-07
[v6r16] Fix with checksum propagation and handling

chaen:
Workflow: properly pass the checksum when uploading output.
DMS: handle properly checksum calculation in the DataManager

This avoids having "False" checksum in the DFC, resulting in problems in RMS/FTS/etc

### DIFF
--- a/DataManagementSystem/Client/DataManager.py
+++ b/DataManagementSystem/Client/DataManager.py
@@ -451,7 +451,15 @@ class DataManager( object ):
     if not checksum:
       log.debug( "Checksum information not provided. Calculating adler32." )
       checksum = fileAdler( fileName )
-      log.debug( "Checksum calculated to be %s." % checksum )
+      # Make another try
+      if not checksum:
+        log.debug( "Checksum calculation failed, try again" )
+        checksum = fileAdler( fileName )
+      if checksum:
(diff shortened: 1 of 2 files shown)


atsareg:
What can be the reason for the checksum to fail in the first place ? Is one retry is sufficient or it should be retried until successful ?

chaen:
That's a good question to which I have no answer, since it is something we have seen very very seldomly (9 times since the beginning in LHCb as far as I can say), and there is nothing in the log.

chaen:
The problem is understood:
* the checksum is calculated at some point
* the checksum is propagated to the failover client
* the checksum is *not* propagated to the DM (which is wrong, regardless of the rest)
* the checksum is computed in the DM, and there is no check whether the computation was done properly or not (which is wrong, no matter what)

Now, I can give you concrete examples of cases when computing the checksum failed (typically Input/Output error), but I have no idea what can cause it to fail, and then successfully copy the file. 

So indeed the comprehension of the problem is not complete, but I disagree in saying that better done nothing.

atsareg:
There was a typo in my previous comment which probably provoked the last comment. Sorry about that.

chaen:
No worries, better more explanations than not enough :-)

PR #3177, opened 2016-11-18
[v6r17] Update CAs when running dirac-proxy-init if they are out-of-date.

sfayer:
Hi,

We frequently have users run into problems because their CRLs have expired. This patch updates the CA and CRL bundles as part of the dirac-proxy-init process if none of the files in the certificates directory have been updated for 28 days. This should avoid the problem and not break things if they users have installed fetch-crl or some other method of updating the CRLs.

Regards,
Simon


atsareg:
This is fine. I would suggest only to check that X509_CERT_DIR is write accessible for the client before making the CRL's check. Quite often there is a site wide client installation that users are collectively using having only read access. Still a warning can be printed out in this case.

PR #3181, opened 2016-11-22
[v6r17] CS RabbitMQ internal DB synchronization

wkrzemien:
All tests passed. It can be safely merged because it does not affect any other DIRAC part.

This is rebased content of the PR #3162  without any consumer part.

RabbitMQAdmin is the module that encloses all administration task related to RabbitMQ server e.g. updating internal database. It's RabbitMQ specific.


atsareg:
This is all fine. The only thing is that the RabbitMQ stuff is scattered across Core and WorkloadManagement systems. However, the RabbitMQ can be used by WMS but also by other systems as well, in other words this is part of the DIRAC Framework. I suggest to put all the RabbitMQ classes and utilities into the FrameworkSystem/Service and Utilities packages

PR #3227, opened 2017-01-17
[v6r17] Various fixes

phicharp:
* Fix Logger such that subLogger works fine
* Don't set file Done in RemoveReplicas if there is an error


chaen:
Review OK

PR #3236, opened 2017-02-01
StorageFactory test fix

vipersec:
Fixed test `getStorageElementStatus` has been replaced with `getElementStatus`


chaen:
??? 
Can you explain ? I just wrote this thing last week, and it was behaving just like I expected it to, so why do you want to change it ?

PR #3246, opened 2017-02-08
[v6r17] JobWrapper documentation

chaen:
Add @phicharp 's finding about the watchdog in a temporary doc. It needs to be refined before making it to the admin section

### DIFF
--- a/docs/source/DeveloperGuide/Systems/WorloadManagementSystem/index.rst
+++ b/docs/source/DeveloperGuide/Systems/WorloadManagementSystem/index.rst
@@ -1,0 +1,35 @@
+================
+DIRAC JobWrapper
+================
+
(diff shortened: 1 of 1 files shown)


atsareg:
This is really good to have documentation ! While reading it I have found out that the job wrapper is completely changed compared to its original idea making it ( unnecessarily ) unusable in certain cases. Originally the job wrapper was a _single_ completely self-contained script that can be executed without any other argument or file dependency. As such it could be executed by _any_ ComputingElement ( InProcess, Batch system, etc ). Now it is a bunch of three files with one executable that assumes to find necessary extra files in the current directory. This will work for InProcess or Sudo CEs but not for others. It is a major design change that was done and I think it should be reverted back.

PR #3247, opened 2017-02-08
[v6r17] Fix scripts deployment for MacOS

atsareg:
1. It turns out that on recent MacOS ( Darwin version > 10.10 ), the behavior of "/usr/bin/env python" changed preventing finding the python interpreter in the $PATH. As a result the native python is picked up which is compiled differently, e.g. using UCS2 Unicode string treatment. The PR defines explicitely the python interpreter to be used in the DIRAC commands.
2. os.system() normally uses the environment from the calling shell. However DYLD_LIBRARY_PATH, for some obscure reason is not passed from the calling shell to the subshell of the os.system() command in Darwin > 10.10. In the PR DYLD_LIBRARY_PATH is set right in the os.system() command.

The changes should not affect non-MacOS environments.

### DIFF
--- a/Core/scripts/dirac-deploy-scripts.py
+++ b/Core/scripts/dirac-deploy-scripts.py
@@ -86,7 +86,7 @@ if len( sys.argv ) == 2:
 wrapperTemplate = wrapperTemplate.replace( '$PYTHONLOCATION$', pythonLocation )
 
 # On the newest MacOS the DYLD_LIBRARY_PATH variable is not passed to the shell of
-# the os.system() - bug or feature ?
+# the os.system() due to System Integrity Protection feature
 if platform.system() == "Darwin":
   wrapperTemplate += """
 sys.exit( os.system( 'DYLD_LIBRARY_PATH=%s python "%s"%s' % ( DiracLibraryPath, DiracScript, args )  ) / 256 )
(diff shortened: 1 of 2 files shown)


Core/scripts/dirac-deploy-scripts.py
```diff
@@ -77,9 +77,24 @@
...
 """
 
+# Python interpreter location can be specified as an argument
+pythonLocation = "/usr/bin/env python"
+if len( sys.argv ) == 2:
+  pythonLocation = os.path.join( sys.argv[1], 'bin', 'python' )
+wrapperTemplate = wrapperTemplate.replace( '$PYTHONLOCATION$', pythonLocation )
+
+# On the newest MacOS the DYLD_LIBRARY_PATH variable is not passed to the shell of
+# the os.system() - bug or feature ?
```
andresailer: It is a feature. The "System Integrity Protection" prevents inheritance of some variables for child processes when "protected" executables are used.
https://developer.apple.com/library/content/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html#//apple_ref/doc/uid/TP40016462-CH3-SW1

chaen:
Before merging replace maybe the question with Andre's answer, so we won't wonder in the future :)

PR #3250, opened 2017-02-09
[v617] executionLock optional argument to executeWithUserProxy()

atsareg:
This change allows to pass a lock object to a function call decorated with @executeWithUserProxy. The actual function will be executed with this lock. E.g.:

    res = getFilesToStage( inputData, proxyUserName = userName, proxyUserGroup = userGroup, executionLock = lockObject  )

Or just pass it as a string, then no lock object should be created beforehand:

    res = getFilesToStage( inputData, proxyUserName = userName, proxyUserGroup = userGroup, executionLock = 'ThisParticularLock'  )

### DIFF
--- a/Core/Utilities/Proxy.py
+++ b/Core/Utilities/Proxy.py
@@ -23,6 +23,7 @@ from DIRAC                                               import gConfig, gLogger
 from DIRAC.FrameworkSystem.Client.ProxyManagerClient     import gProxyManager
 from DIRAC.ConfigurationSystem.Client.ConfigurationData  import gConfigurationData
 from DIRAC.ConfigurationSystem.Client.Helpers.Registry   import getVOMSAttributeForGroup, getDNForUsername
+from DIRAC.Core.Utilities.LockRing                       import LockRing
 
 __RCSID__ = "$Id$"
 
@@ -41,6 +42,8 @@ def executeWithUserProxy( fcn ):
   :param str proxyUserDN: the user DN of the proxy to be used
   :param str proxyWithVOMS: optional flag to dress or not the user proxy with VOMS extension ( default True )
   :param str proxyFilePath: optional file location for the temporary proxy
+  :param str or object executionLock: optional lock object for execution of the original function;
+                                      if string, then named LockRing lock will be created
   """
 
   def wrapped_fcn( *args, **kwargs ):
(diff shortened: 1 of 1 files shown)


Core/Utilities/Proxy.py
```diff
@@ -41,6 +42,8 @@ def executeWithUserProxy( fcn ):
   :param str proxyUserDN: the user DN of the proxy to be used
   :param str proxyWithVOMS: optional flag to dress or not the user proxy with VOMS extension ( default True )
   :param str proxyFilePath: optional file location for the temporary proxy
+  :param str or object executionLock: optional lock object for execution of the original function;
```
andresailer: Does this properly render in the documentation?
maybe
```
:param executionLock: .....
:type executionLock: str or object
```
would be better

Core/Utilities/Proxy.py
```diff
@@ -71,6 +78,9 @@ def wrapped_fcn( *args, **kwargs ):
       if not result['OK']:
         return result
 
+      if executionLock:
+          executionLock.acquire()
```
chaen: the alignment is wrong, you have 2 extra spaces

chaen:
And just for safety, don't we want this lock to be recursive ? I can certainly imagine cases where we would end up calling two times executewithuserproxy

atsareg:
I guess you mean reentrant RLock object. I do not see immediately cases for an executeWithUserProxy decorated function to call another similarly decorated function. But may be this could be a case.

atsareg:
Ah, indeed the LockRing().getLock() argument is recursive :) Well, whatever R means here ...

chaen:
Yes, R like something :-) 
I do not think we now have a case, but I wouldn't be surprised if one method called with executewithuserproxy calls another method like that later on. Putting a RLock right now will probably save some headaches in the future :-)

phicharp:
Hi Andrei,
Why don't you hardcode the name of the lock inside the decorator rather... It doesn't harm is the decorator is always protected, right? Or then the optional argument can just be executionLock = False (by default) and I call it with executionLock = True
What do you think?
Otherwise we have to be damned sure that the same name is used in all potential contenders. which is not necessarily guaranteed
Cheers, Ph.

PR #3254, opened 2017-02-13
[v6r18] Use the fts-rest provided Request wrapper

chaen:
As of lcg-bundle 2017-01-27, we don't need our wrapper anymore


atsareg:
The lcg-bundle 2017-01-27 is available for SL6 and CC7 ?

chaen:
@closier can tell

PR #3265, opened 2017-02-21
[v6r17] set a very high number for the size of the EL result

zmathe:


Core/Utilities/ElasticSearchDB.py
```diff
@@ -334,20 +334,20 @@ def getUniqueValue( self, indexName, key, orderBy = False ):
       query.aggs.bucket( key,
                          'terms',
                          field = key,
-                         size = 1,
+                         size = 10000,
```
andresailer: Maybe it would be possible to define a constant somewhere?

zmathe: I hope I do not have to change very often. Constant in python make no sense. Using a variable is very dangerous as I can change the value any time..., but anyway as you prefer.

chaen: It makes sense for readability. If you are worried about people changing it at run time, I would recommend writing a comment on top of the variable (written in capital btw ;-) ). And if you are so super worried, we can make a wrapper class that forbid changing it after initialization :)  But this is very much an overkill IMHO

PR #3267, opened 2017-02-21
[v6r17] Fix RemoveReplica handler and SE metadata keys

phicharp:
* RemoveReplica: If an error was set it was never reset 

* Fixes for solving issue https://github.com/DIRACGrid/DIRAC/issues/2502 on SE metadata keys


chaen:
Review OK

PR #3275, opened 2017-03-02
[v6r17] Lower security log gzip memory usage

sfayer:
Hi,

When the security logs are rotated, the buffer size used is ~2GiB... This seems excessive and on our installation causes the gzip to fail most of the time; any buffer larger than the compression block size should be just as good (>64KiB). This patch lowers the buffer size to 1MiB,

Regards,
Simon

### DIFF
--- a/FrameworkSystem/private/SecurityFileLog.py
+++ b/FrameworkSystem/private/SecurityFileLog.py
@@ -74,11 +74,11 @@ class SecurityFileLog( threading.Thread ):
       gLogger.info( "Compressing file %s" % filePath )
       fd = gzip.open( "%s.gz" % filePath, "w" )
       fO = file( filePath )
-      bS = 1048576
-      buf = fO.read(2**31-1)
+      bS = 1048576 # 1MiB
+      buf = fO.read( bS )
       while buf:
         fd.write( buf )
-        buf = fO.read(2**31-1)
+        buf = fO.read( bS )
       fd.close()
       fO.close()
     except Exception as e:


chaen:
I think this was done purposely:
https://github.com/DIRACGrid/DIRAC/pull/2688

Maybe @roiser can comment

sfayer:
I think before #2688 it was just using read(), which could read even more than 2GiB in one go, this triggers a bug in the python2.7 zlib crc32 library (https://bugs.python.org/issue27130), so it was capped at 2^31-1, which is the largest block size that doesn't trigger the bug. I don't think making the block size even smaller should cause any problems.

Regards,
Simon

chaen:
I see, I remember now. Can you maybe just make sure about that and test your code on a 3GB file ? I am also a bit curious about the performance difference (even if here it does not really matter, it is just curiosity....).
Just for my understanding, it crashes on your installation because of lack of memory  ?

roiser:
@sfayer is correct on the statement above. Initially the read was getting everything of the file to zip even when bigger than 2 GB. then internally in the zipping there is a crc32 function which is limited to 2**31-1 bytes, that's why I restricted the reading of one chunk to this amount.

sfayer:
Yes, the crash on our system is it throwing a MemoryError due to lack of free memory on the machine... It doesn't cause any trouble, we just end up with a partially written gzip file for that day and the uncompressed CSV is left intact.

I've benchmarked this function using a 3GB file (on a modern machine with plenty of spare RAM & not much else going on):
 - 1MiB block size: 103 seconds
 - 2^31-1 block size: 540 seconds
 - unspecified (whole file) block size: <crash as expected after 20 seconds>

I was surprised the larger size was worse, I was expecting them to be about the same but there must be something in cpython that doesn't scale well to large buffers.

Regards,
Simon

chaen:
Oh ! Surprising results yes :-)
Thanks !

PR #3284, opened 2017-03-15
[v6r17] fixes in documentation

andresailer:
replace unicode with quotes
replace symlink with content, solves problems for some eclipse users

The scripts from the Tools folder should move to source so that the two different fakeEnvironment files are no longer necessary.

### DIFF
--- a/docs/Tools/fakeEnvironment.py
+++ b/docs/Tools/fakeEnvironment.py
@@ -1,1 +1,103 @@
-../source/fakeEnvironment.py
+''' fakeEnvironment
+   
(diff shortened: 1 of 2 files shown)


chaen:
```The scripts from the Tools folder should move to source so that the two different fakeEnvironment files are no longer necessary.``` Why not doing it in this PR ? is it difficult ?

andresailer:
Yes, well I would need to understand what some of the scripts are doing :)

andresailer:
Sorry, but I think this should go to v6r17

PR #3303, opened 2017-04-06
[v6r17] JobScheduling fixes

phicharp:
Add a CS flag to decide whether the optimizer checks the integrity of files only at Tape SEs or at all SEs.

Default is to only check at Tape SEs...


phicharp:
Likely tests will fail, but I have no time to update them today (if it is even possible)

PR #3304, opened 2017-04-06
[v6r17] Fix JobScheduling

phicharp:
* Add a CS flag to decide whether to check only replicas at tape SEs or all replicas. Default is to test only at Tape SEs
* If any file is not in the FC, make the job Failed instead of looping infinitely
* Avoid checking files if one replica was already found Online
* restrict usage of decorator to strict minimum, i.e. calling the SE.getFileMetadata

### DIFF
--- a/StorageManagementSystem/Client/StorageManagerClient.py
+++ b/StorageManagementSystem/Client/StorageManagerClient.py
@@ -20,12 +20,28 @@ def getFilesToStage( lfnList, jobState = None, checkOnlyTapeSEs = None, jobLog =
     return S_OK( {'onlineLFNs':[], 'offlineLFNs': {}, 'failedLFNs':[], 'absentLFNs':{}} )
 
   dm = DataManager()
+  if isinstance( lfnList, basestring ):
+    lfnList = [lfnList]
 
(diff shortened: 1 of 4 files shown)


phicharp:
The latest changes have been tested successfully and in real life with a job that had no input files in FC

phicharp:
tests remain to be fixed as return error messages were changed...

phicharp:
Please re-review as it seems new push doesn't cancel the approval

andresailer:
Please add the new option to the documentation
https://github.com/DIRACGrid/DIRAC/blob/integration/docs/source/AdministratorGuide/Configuration/ConfReference/Systems/WorkloadManagement/Executors/JobScheduling/index.rst

phicharp:
Finding out where the documentation is is a real challenge... is there any logic in this?
In addition the markup is just awful...
If you are interested, there is a rendering web site to check what you have done is not destroying everything:
https://livesphinx.herokuapp.com

andresailer:
Thanks!
emacs, of course, has a rst mode, which makes editing the documentation files less painful. Finding them is still a major pain

PR #3333, opened 2017-05-08
[v6r17] Implement PlatformCheck code in JobScheduling module.

sfayer:
Hi,

This code enhances the JobScheduling executor to check the platform specified by the user is actually available at a subset of the sites they've picked. The new tests are only enabled if CheckPlatform = True (default is False) in the CS, so backwards compatibility is maintained. I've removed the old "PlatformCheck" stub from the JobSanity module as this was never implemented and it makes more sense to do it in the JobScheduling module after the list of active sites for the job has been retrieved.

At the same time I've also adjusted the code so that the job manifest is only fetched once, which simplifies some of the other functions slightly.

Regards,
Simon

### DIFF
--- a/WorkloadManagementSystem/Executor/JobSanity.py
+++ b/WorkloadManagementSystem/Executor/JobSanity.py
@@ -65,14 +65,6 @@ class JobSanity( OptimizerExecutor ):
       finalMsg.append( "%s LFNs" % result[ 'Value' ] )
       self.jobLog.info( "%s LFNs" % result[ 'Value' ] )
 
-    #Platform check # disabled
-    if self.ex_getOption( 'PlatformCheck', False ):
-      result = self.checkPlatformSupported( jobState )
-      if not result[ 'OK' ]:
-        return result
-      finalMsg.append( "Platform OK" )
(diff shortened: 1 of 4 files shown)


fstagni:
Looks good to me. @atsareg ?

atsareg:
This new feature is fine to me. One remark - the jobs not finding proper sites should rather be declared failed, I think. The unavailable Platform will most likely be due to mistype in the job requirements. So, putting them on hold will most likely keep them in the JobDB until deleted manually. Failed jobs a clearly signaling the problem and will be eventually cleaned up automatically.

atsareg:
Does anybody understand why Travis is failing in this PR ?

andresailer:
Intermittent error on the CI side? Just restart the build via the button on travis.

atsareg:
Yes, looks like a stalled build process. However, I do not see how to restart the build. The "build" button on the Travis page just prints out some image info. Any ideas ?

andresailer:
If you click on details, there should be the button on the top right.
image

atsareg:
Ah, yes. Was not properly logged in Travis. After logging in, the button is there. Thanks

PR #3354, opened 2017-05-30
[v6r18] WMSAdministratorHandler: linting the code

miloszz:
Fixing ~25 pylint warnings during the workshop


petricm:
Using `@staticmethod` instead of `@classmethod` is probably better here

PR #3365, opened 2017-06-06
[v6r19] Update to gfal 2.13.3 API

chaen:
This will be for when we get the new LcgBundle with gfal 2.13.3 in it


chaen:
This seems to work fine, but I'll wait till the issue with the test machine is fixed to give a final go

PR #3413, opened 2017-07-26
[v6r19] Replace gLogger and fix some issues due to the changes

aldbr:
This pull request just replaces the logging system *gLogger* by the new one and fix some issues due to the changes.

### DIFF
--- a/Core/Base/API.py
+++ b/Core/Base/API.py
@@ -65,21 +65,31 @@ class API( object ):
 
   #############################################################################
 
-  def __getstate__( self ):
-    """ called when pickling the object. Also used in copy.deepcopy """
-    from DIRAC.FrameworkSystem.private.logging.SubSystemLogger import SubSystemLogger
-    d = dict( self.__dict__ )
-    if isinstance( d['log'], SubSystemLogger ):
-      d['log'] = d['log']._subName
-    return d
+  def __getstate__(self):
+    """ Return a copied dictionary containing all the attributes of the API.
+        Called when pickling the object. Also used in copy.deepcopy.
 
-  #############################################################################
+    :return: dictionary of attributes
+    """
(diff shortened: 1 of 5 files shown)


Interfaces/API/Job.py
```diff
@@ -1223,4 +1223,27 @@ def runLocal( self, dirac = None ):
 
     return dirac.submitJob( self, mode = 'local' )
 
+  #############################################################################
+
+  def __getstate__(self):
```
chaen: This is not from you. Why is it in your PR ?

aldbr: I updated the existing methods in API

chaen:
OK for me. Maybe @andresailer wants to check this getstate method ?

Core/Base/API.py
```diff
@@ -65,21 +65,31 @@ def __init__( self ):
...
-    self.__dict__.update(d)
-    if isinstance( d['log'], basestring ):
-      self.log = gLogger.getSubLogger( d['log'] )
+    :return: dictionary of attributes
+    """
+    from DIRAC.FrameworkSystem.private.standardLogging.Logging import Logging
+    state = dict(self.__dict__)
+    # Replace the Logging instance by its name because it is not copyable
+    # because of the thread locks
+    if isinstance(state['log'], Logging):
```
andresailer: To keep this backward compatible, also check for SubSystemLogger instance (keeping the other import)
```
if isinstance(state['log'], (SubSystemLogger, Logging) ):
```

andresailer:
`__setstate__` seems to only differ in whitespace and variable names

Core/Base/API.py
aldbr: Ok I kept these lines

PR #3420, opened 2017-08-04
[v6r19] Fix the import logging issue

aldbr:
This pull request mainly concerns the following code which caused a double display of each log: 
`import logging`
`logging.basicConfig(...)`

These lines are replaced by a gLogger method which enables or disables the display of the logs providing by all the external libraries. I also added a test class for this new method and updated the documentation.

By the way, I change the test path from *FrameworkSystem/test/testLogging/tests/* to *FrameworkSystem/test/testLogging/*

### DIFF
--- a/FrameworkSystem/private/standardLogging/LoggingRoot.py
+++ b/FrameworkSystem/private/standardLogging/LoggingRoot.py
@@ -49,6 +49,9 @@ class LoggingRoot(Logging):
     # initialize the root logger
     # actually a child of the root logger to avoid conflicts with other libraries which used 'logging'
     self._logger = logging.getLogger('dirac')
+    # prevent propagation to the root logger to avoid conflicts with external libraries
+    # which want to use the root logger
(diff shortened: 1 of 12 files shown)


FrameworkSystem/private/standardLogging/LoggingRoot.py
```diff
@@ -210,3 +213,23 @@ def __configureLevel(self):
       self._setLevel(LogLevels.DEBUG)
       self.showHeaders(True)
       self.showThreadIDs(True)
+
+  def enableLogsFromExternalLibs(self, isEnabled=True):
```
atsareg: This function signature looks confusing. I would suggest to make two functions:

  enableLogsFromExternalLibs()
  disableLogsFromExternalLibs()

Those can just call the original function with the corresponding flag

aldbr: Okay, I added the methods

PR #3442, opened 2017-08-17
[v6r19] Fix StorageElement inheritance and add a new Resources section

aldbr:
This pull request:

+ fixes different weird behaviours in Storage Element, as the inheritance of the protocol sections
+ creates a new Resources section named *StorageElementBases* containing only Base SEs
+ updates the documentation

### DIFF
--- a/docs/source/AdministratorGuide/Systems/DataManagement/index.rst
+++ b/docs/source/AdministratorGuide/Systems/DataManagement/index.rst
@@ -43,92 +43,125 @@ DIRAC provides an abstraction of a SE interface that allows to access different 
 
 In order to factorize the configuration, it is possible to use BaseSE, which acts just like inheritance in object programming. You define a SE just like any other. This SE can then be refered to by another SE. This new SE will inherit all the configuration from its parents, and can override it.  For example::
 
-   CERN-EOS
-   {
-     BackendType = Eos
-     SEType = T0D1
-     AccessProtocol.1
-     {
-       Host = srm-eoslhcb.cern.ch
-       Port = 8443
-       PluginName = GFAL2_SRM2
-       Protocol = srm
-       Path = /eos/lhcb/grid/prod
-       Access = remote
-       SpaceToken = LHCb-EOS
-       WSUrl = /srm/v2/server?SFN=
-     }
-   }
-   CERN-DST-EOS
-   {
-     BaseSE = CERN-EOS
-   }
-   CERN-USER
-   {
-     BaseSE = CERN-EOS
-     PledgedSpace = 205
-     AccessProtocol.1
-     {
-      PluginName = GFAL2_SRM2
-      Path = /eos/lhcb/grid/user
-      SpaceToken = LHCb_USER
-     }
-   }
-
-
-This definition would be strictly equivalent to::
-
-   CERN-EOS
-   {
-     BackendType = Eos
-     SEType = T0D1
-     AccessProtocol.1
-     {
-       Host = srm-eoslhcb.cern.ch
-       Port = 8443
-       PluginName = GFAL2_SRM2
-       Protocol = srm
-       Path = /eos/lhcb/grid/prod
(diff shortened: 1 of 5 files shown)


aldbr:
I just add a commit for a very little condition improvement in StorageFactory

chaen:
providing the tests succeed, review OK.
Please add in the wiki for v6r19 that as of this release, ```ProtocolName``` is not supported anymore in the configuration and must be replaced by ```PluginName```

aldbr:
Alright, I added the information

docs/source/AdministratorGuide/Systems/DataManagement/index.rst
```diff
@@ -43,92 +43,125 @@ DIRAC provides an abstraction of a SE interface that allows to access different
...
         Access = remote
-        SpaceToken = LHCb_USER
+        SpaceToken = LHCb-EOS
         WSUrl = /srm/v2/server?SFN=
       }
-   }
+    }
 
-Note that in case of overwriting, the protocol sections are matched using the PluginName attribute, and not the section name, which is irrelevant at all.
+Note that base SE must be separated from the inherited SE in two different sections. You can also notice that the name of the protocol section can be a plugin name. In this way, you do not need to specify a plugin name inside.
```
phicharp: Don't we want to make that a rule at some point?
The current implementation eases the migration, but at some point we should not have the PluginName any longer, right?

chaen:
We must have it if we ever want to use the same plugin twice (and we will want soon I suspect ;-) )

chaen:
For reference, this was agreed at this bild dev: https://indico.cern.ch/event/608327/

docs/source/AdministratorGuide/Systems/DataManagement/index.rst
atsareg: My experience shows that giving "meaningful" names to sections is often confusing. So, please, keep PluginName as a mandatory parameter and do not try to interpret the section name. 

Eventually (not now), I suggest to introduce yet another section, e.g. StoragePlugins, in the SE definition in which all the per plugin sections will go. Then there will be no confusion that each section corresponds to a plugin. This is how it is done for Queues in the CE sections. So, SEs should be described symmetrically. But even in this case I would still keep PluginName as mandatory. This is what was foreseen for v7r0. A crude example below:  

  StorageElements
  {
    SE1
    {
       Option1 = value
       Option2 = value
       StoragePlugins 
       {
          SomeName_1
          {
              PluginName = name1
              ...
           } 
           SomeName_2
           {
               PluginName = name 2
            }
       }
    }     
  }

atsareg:
We now have a CS section inheritance in several places. We should think of making it a generic feature of the CS. So, that the inheritance is only interpreted by the CS client and all the other consumers of the CS data are just getting a ready to use data structure

phicharp:
AFAIS, what was implemented is exactly what had been approved in February. Backward compatibility has been preserved as requested.
Therefore I don't see why this PR was closed. If users want to keep the previous behavior, they can keep it... We in LHCb will use the new feature which is much clearer than having sections called randomly `AccessProtocol.1` and for which one doesn't know what they mean... 
@atsareg: please re-open this PR

atsareg:
This was not at all my intention to close the PR. Just pressed the wrong button when commenting. Sorry about that.

phicharp:
Why the review is limited to the few developers who have write access to the repository? This is absurd, isn't it?

phicharp:
@aldbr : Alexandre, you should rebase your PR with integration apparently...

aldbr:
@phicharp Ok, I rebased it with integration

atsareg:
@phicharp: review is indeed limited by the Github to those who have write privileges for the repo. But we know each other. So, if you leave a comment "Review OK", this is taken into account in the same way as other reviews when considering merging a given PR.

phicharp:
Review OK (again)

PR #3447, opened 2017-08-18
[v6r17] HTCondor transfer files

VladimirRomanovsky:
Transfer output files(only log and err) for remote scheduler


andresailer:
This also affects the local scheduler. But its OK, and I guess it solves the problem where condor was transferring LDSB files to the submission machine.(?)

VladimirRomanovsky:
Andre, your guess is correct.

PR #3351, opened 2017-05-29
[v6r18] Correct documentation for PluginName

chaen:
Replace ProtocolName with PluginName

PR #3455, opened 2017-08-25
[v6r19] SEFactory backward compatibility fix

atsareg:
BEGINRELEASENOTES
*Resources
FIX: StorageElementFactory - preserve backward compatibility with respect to using protocol section name as the name of plugin in the absence of PluginName option 
ENDRELEASENOTES

### DIFF
--- a/Resources/Storage/StorageFactory.py
+++ b/Resources/Storage/StorageFactory.py
@@ -145,6 +145,7 @@ class StorageFactory( object ):
     for protocolSection, protocolDetails in self.protocols.iteritems():
       pluginName = protocolDetails.get( 'PluginName' )
       if pluginName is None:
+        gLogger.warn( "No MANDATORY PluginName option is found in the SE protocol section" )
         # Try ProtocolName/protocolSection for backward compatibility
         pluginName = protocolDetails.get( 'ProtocolName', protocolSection )
       if pluginList and pluginName not in pluginList:
--- a/Resources/Storage/StorageFactory.py
+++ b/Resources/Storage/StorageFactory.py
@@ -143,7 +143,10 @@ class StorageFactory( object ):
     turlProtocols = []
     # Generate the protocol specific plug-ins
     for protocolSection, protocolDetails in self.protocols.iteritems():
-      pluginName = protocolDetails.get('PluginName', protocolSection) 
+      pluginName = protocolDetails.get( 'PluginName' )
+      if pluginName is None:
+        # Try ProtocolName/protocolSection for backward compatibility
+        pluginName = protocolDetails.get( 'ProtocolName', protocolSection )
       if pluginList and pluginName not in pluginList:
         continue
       protocol = protocolDetails['Protocol']
--- a/Resources/Storage/StorageFactory.py
+++ b/Resources/Storage/StorageFactory.py
@@ -146,8 +146,8 @@ class StorageFactory( object ):
       pluginName = protocolDetails.get( 'PluginName' )
       if pluginName is None:
         gLogger.warn( "No MANDATORY PluginName option is found in the SE protocol section" )
-        # Try ProtocolName/protocolSection for backward compatibility
-        pluginName = protocolDetails.get( 'ProtocolName', protocolSection )
+        # Try protocolSection for backward compatibility. This is to be dropped as deprecated
+        pluginName = protocolSection
       if pluginList and pluginName not in pluginList:
         continue
       protocol = protocolDetails['Protocol']


Resources/Storage/StorageFactory.py
```diff
@@ -143,7 +143,10 @@ def getStorages( self, storageName, pluginList = None, hideExceptions = False ):
     turlProtocols = []
     # Generate the protocol specific plug-ins
     for protocolSection, protocolDetails in self.protocols.iteritems():
-      pluginName = protocolDetails.get('PluginName', protocolSection) 
+      pluginName = protocolDetails.get( 'PluginName' )
```
chaen: We announced at the DUW that this feature was being dropped at the next release

chaen: It is even in the "How to upgrade": https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-v6r19#abandon-of-the-protocolname-option

And the doc was updated as of v6r18: https://github.com/DIRACGrid/DIRAC/pull/3351

atsareg: It does not harm to have backward compatibility. On the other hand, we should not use the section name as the plugin name. So, if we drop the backward compatibility, we should also impose that PluginName option is mandatory

chaen: 1. It does as it makes the code more complex
2. Having the section name as plugin name was discussed and agreed so I do not understand why you are now coming back on that. In any case, pluginName is still supported

atsareg: I do not agree with using the section name as the plugin name. You yourself made arguments against that (possibility of different plugins of the same type). One can always call the section the same way as the plugin name but this is just a section name which is there for readability.

chaen: I believe that the BILD meeting when this was discussed would have been the place to object. You can't just now put this as a Mandatory attribute on your own decision, this is not really fair.

atsareg: ??? We are speaking about the quality of the system, it is not about "fair or not fair". Having section name used as plugin name have several severe drawbacks:
1. Two ways to define the same thing is bad because can lead to confusions
2. If at some point we will need another non-protocol section on the same level this will not be possible.
3. It does not follow the same patterns as in the similar Site/CE part of the CS and we should limit the number of  different patterns for the same purpose.

phicharp: @atsareg: I don't understand then why we have meetings and a DUW where things are presented and discussed if months later you take an authoritarian decision to no longer agree... 
Your argument on the meaning of the section name can be transposed to e.g. the StorageElement. Why do we call the section with the SE name? It could be called anything and the name of the SE could be an option StorageElementName in there... Would it make the CS more readable?
What has been implemented by Alexandre implements exactly what was presented and agreed in the BiLD meeting. It does not prevent whoever wants to continue using a meaningless name as section name to do it and define the PluginName inside the section. 
As a matter of fact these sections are just defining the different plugins and it makes a complex CS like LHCb's much more readable.
The meaningless section name was "useful" at some point as (wrongly) the plugins were considered in alphabetical order! But now this makes just no sense...

atsareg: I like your statement about "an authoritarian decision to no longer agree" ;) So, whenever I disagree with something this is an authoritarian decision. Point taken. Vive la democracie ! 

I do not want to reproach Alexandre at all, he has done a good job to implement what was agreed in a BiLD meeting and other discussions at CERN. I do not remember if I was present on this meeting, most likely not, or this was my fault not to make the remark we are discussing here. However, I asked explicitly in the review remarks of the initial PR and some subsequent ones NOT TO USE section names as a plugin names. But this my request was democratically ignored.

My point is that in order to avoid any confusion, section names should never be used as "meaningful" values, e.g. pluginNames unless they are put in special upper section which can contain only sections of a well defined nature. For example, Sites section contains only sections per site, in which CE section contains only sections per CE in which Queues section contains sections per queue. This is what we agreed many years already ago when we discussed the new v7r0 CS structure which unfortunately never materialized. But the idea was correct and resulted from painful experience and long discussions.

The current "meaningless" sections AccessProtocol.X are there for historical reasons dating back to the days where we had only 2 level hierarchy in the CS. So, this was a poor man approach to express "AccessProtocol section with per protocol subsections". Since now we are free to introduce new levels of sections in the CS, the "AccessProtocol.X" style is certainly deprecated. But you try to actually resurrect it just giving these sections new "meanningful" names.   

As for the readability, I just do not see how this is affected at all - you are always free to use plugin section names that makes sense. However, if you make section names as plugin names but you are still obliged to have a possibility to redefine PluginName inside, then this is certainly a source of confusion because admins will have to remember two ways to define plugin name while "for readability" they are accustomed to just one way because it is used 99% of times. So, with mandatory PluginName option you does not loose readability and you have a stricter and less confusing CS schema. This is a clear advantage. 

We have to decide on that now before the release is finally out to give a clear message in the release instructions. Please, concentrate on my arguments and not on my "authoritarian" behavior.

atsareg: By the way, from the coding point of view, there are practically no changes ( one line ), so Alexandre's work is not affected. This is more an administration/operations issue and avoiding nasty confusion in the future

atsareg: As for the StorageElement section - thank you for this example. This section is dedicated for per SE subsections only. No other options or sections of another nature can go there. In the v7r0 RFC the StorageElement section was proposed to move to the Site sections on the same level as the CE section name. This is what we will certainly do in the future, I hope.

phicharp: Andrei, precisely the reason we would like to set the section name as the plugin name (if the user so wishes) is an operational one... We have to manage all these SEs and it is much clearer if we know without opening the section which plugin it refers to!
It just is quite impractical to have to put an option in the section that just repeats the same information!

Since the functionality we propose doesn't prevent other users to continue doing exactly the same thing as now, i.e. using an explicit PluginName option, I don't understand why you continue insisting that this option is mandatory.

Would you be happier if there would be a section "Plugins" in each SE where all the plugins will be defined? We do not foresee to have other sections in the SE definitions but the plugins, therefore IMHO this would be a real overkill... 

In the Sites definition the situation is different as we have to define CEs and Network information, thus the one more level...

As for "authoritarian", my remark was not anti-democratic! It is just that you seem to be the only person of your opinion, and you have all authority to approve yourself your own PR and therefore modify a decision that had been taken (I think) democratically in a meeting where  everyone could express himself, and at which you were present... 

There is no need to anyone to change anything when the release comes out. It is just one additional possibility offered to VOs, which we should of course mention in the documentation, but there is no need to change the CS at all...

Bon week-end!

phicharp: if you have forgotten, here are the minutes of the meeting:
https://indico.cern.ch/event/608327/

phicharp: Finally concerning the argument that we may be willing to add sections others than plugins, which may be valid, this would imply that we **must** add a section called "Plugins" into which we can then put the plugins. This would require are real change to all users. Do we want to do that?
Our CS is made in such a way that all sections' names do have a meaning, either as a container of other sections, or as descriptive sections. It is impossible to have at the same level sections that are descriptive and have different contents.
To finish, the current plugin sections are the only ones throughout the CS whose name is totally meaningless! I it normal?

phicharp: Sorry, clicked on the wrong button when adding my comment ("Start a review" rather than "Add single comment")... I was not willing to approve the changes (which anyway I don't have the karma for!)

atsareg: OK, indeed that this is my fault that I have not raised this point in this meeting as I said before.
My bad. 

In this release we are also changing ProtocolName -> PluginName, so everybody will have to update their CS definitions in any case, either by changing the option name or section name. My preferred solution would be to have a Plugins section where all the corresponding subsections will go indeed. This is what we will have to do eventually, why not now. 

Now you say that you do not plan to have other sections than plugins on this level, but later when you will need it you will just not be even able to do so. You cut your possibilities because of saving one click. And even this click is not necessary in most cases if you give relevant names to these sections. You can be very well organized for that and trust what you are doing. But you are not alone and others can be different. And in general, now if you see GFAL2_SRM section in fact you can not be 100% sure that this is not redefined inside even non-intentionally and is not actually the cause of the problem that you are debugging. To me - you gain almost nothing and you create potential sources of confusion, you cut flexibility for the future and you make different structure compared to other similar sections. But you do not want to discuss these arguments preferring to insist that it is too late because it was discussed in a BiLD meeting. 

If you have noticed, this PR just adds a warning log message and does not change anything else. And it is not yet merged despite my authoritarian character. I just want you to understand my arguments and look at this issue in a bit wider context than just "saving one click for operations of a data manager".

atsareg:
I guess our last comments crossed. All the DIRAC installations will have to make CS changes with this release, so we have to decide what are these changes. I think my last comment answers your last comment already ;)

atsareg:
Current plugin section names are inherited from the past as I tried to explain before. Of course, these names unlike before are not needed in this form any more. DIRAC admins are free to change them in whatever way, not necessarily to call them after the plugin names

Resources/Storage/StorageFactory.py
chaen: By using the word "fair", I intended to keep the discussion calm, which I obviously failed at.
Let's not make this personal, it is useless on such a github PR, so I am purposely going to "forget" some points here.

Andrei, your points are taken, and are valid. 
Indeed, we force the users to change their CS. But this change was prepared by announcing long in advance, we kept backward compatibility for quite some version, the doc has been updated, and the change was explicitly repeated at the DUW. So if some users failed at knowing this change, they probably did not upgrade DIRAC since version v4r0p2. 

Given the success of v7r0 that we know, I think one should refrain from trying to stick too much to what was described at the time. Not that everything is to be thrown away, but v7r0 as is should not be an objective in my opinion.
But of course, one could rethink the CS structure. It wouldn't hurt in fact...

Also, we should have a discussion (for example, at a BILD meeting) about what we do with backward compatibility. I am sorry, but yes, backward compatibility has a cost: it makes the code much more difficult to read and to maintain. And Daniela's issue on the forum is one of the many example. We have all encountered some very shadowy part of DIRAC, which we do not understand, which appears to be just for backward compatibility reason not to break a VO with one user. These things will backfire at us eventually. I understand and even agree with the idea that one should not have breaking changes from one version to the other. But I refuse to suffer in my daily life for operational AND development aspects because there are a couple of lazy administrator in the world. So I would be in favor of a strong rule like "backward compatibility are kept for 2 major versions". To be discussed...

Bon weekend to everyone !

phicharp: I think there is a misunderstanding at some point, most likely because I didn't express myself clearly...

When a section contains a list of sections, all these sections must describe the same type of entity. This is by construction. When the SE is built, it gets the list of protocols/plugins by getting the sections contained in its own section. Therefore it is _by construction_ not possible that a section inside, say, CERN-RAW, describes anything but a protocol/plugin. I hope we agree on that.
So if we believe that there may be other entities that describe an SE, what we should do is to add one level, namely /Resources/StorageElements/<SEName>/Plugins/ in which we put all the plugins sections. 
Then I hope we would agree that the plugin sections should be names after the name of the plugin, just like SE sections are named after the name of the SE (without an additional option "Name = <SEName>" inside the section. 

Concerning the replacement of _ProtocolName_ by _PluginName_, this is something that has been there for very long with backward compatibility. In LHCb there is no longer a single _ProtocolName_. It was announced at the DUW and hopefully communities had time to prepare their CS. If this is not the case, I would suggest we re-establish this backward compatibility, which is not a big deal. It then makes 3 ways to give the name of the plugin to a section, but this is hopefully temporary... The documentation should state what is the order of precedence in case users want to use the 3 possibilities at once and give 3 different names, but this would be really weird...

If any change w.r.t. what Alexandre has coded, I would agree that we should allow (not force in a first instance, only give the possibility) to put the plugins on a _Plugins_ section. The code would look there first and if it doesn't exist, would get the sections directly in the SE section... Again one more backward compatibility to keep track of and decide when to break it...

PR #3481, opened 2017-10-10
[v6r20] new fts3 only last rebase

chaen:
Replaces https://github.com/DIRACGrid/DIRAC/pull/3371
I will add info in the twiki

BEGINRELEASENOTES

*DataManagementSystem
NEW: new FTS3 only agent, to replace the existing one

ENDRELEASENOTES


chaen:
@andresailer please use that one

schr476:
Belle II will install/test on our migration server.
(Sorry for the delay)

PR #3483, opened 2017-10-10
[v6r19] RunLocal: bring back treatment of files in working local submission d…

andresailer:
Another fix for #3370

### DIFF
--- a/Interfaces/API/Dirac.py
+++ b/Interfaces/API/Dirac.py
@@ -872,8 +872,9 @@ class Dirac( API ):
           elif os.path.isdir( isFile ):
             self.log.debug("Input Sandbox %s is a directory, found in the user working directory, copying it" % isFile)
             shutil.copytree( isFile, os.path.basename( isFile ), symlinks = True )
-          elif os.path.exists( isFile ):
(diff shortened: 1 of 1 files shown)


Interfaces/API/Dirac.py
```diff
@@ -872,8 +872,9 @@ def runLocal(self, job):
           elif os.path.isdir( isFile ):
             self.log.debug("Input Sandbox %s is a directory, found in the user working directory, copying it" % isFile)
             shutil.copytree( isFile, os.path.basename( isFile ), symlinks = True )
-          elif os.path.exists( isFile ):
```
andresailer: This check is already done before the else, because it checks if the file is in the current working directory.
Also the following string says something about copying while it should say it is already in the current working directory

Interfaces/API/Dirac.py
```diff
@@ -872,8 +872,9 @@ def runLocal(self, job):
...
-            self.log.debug("Input Sandbox %s is a file, found in the current working directory, copying it" % isFile)
+          elif os.path.exists(os.path.join(curDir, os.path.basename(isFile))):
```
andresailer: Here we check that the file is in the directory where the submission is called (`curDir`), then the file is copied

PR #3512, opened 2017-10-25
[v6r19] Reinitialize ObjectLoader and Logging after the CS is fully loaded

chaen:
There is a bootstrap issue with the new logging system. The global sequence of initialization goes as follow:

* load local dirac.cfg
* instantiate gLogger and hence ObjectLoader singleton
* load global CS

Now, if there are extensions, they are most of the time declared in the central CS, which is loaded after the ObjectLoader is used for the first time. And since it is a singleton, it will not "refresh" them.
This PR is to fix this issue. After loading the central CS, we retrigger the initialization of the gLogger, and force the ObjectLoader to reconsider the extensions

PS: sorry for all the white space mess, the original author did not have correct settings

BEGINRELEASENOTES

* Configuration
FIX: retrigger the initialization of the logger and the ObjectLoader after all the CS has been loaded

ENDRELEASENOTES

### DIFF
--- a/ConfigurationSystem/Client/LocalConfiguration.py
+++ b/ConfigurationSystem/Client/LocalConfiguration.py
@@ -184,8 +184,8 @@ class LocalConfiguration( object ):
       return S_ERROR( str( e ) )
     return S_OK()
 
-  def __initLogger( self, componentName, logSection ):
-    gLogger.initialize( componentName, logSection )
+  def __initLogger( self, componentName, logSection, forceInit = False ):
+    gLogger.initialize( componentName, logSection, forceInit = forceInit )
 
     if self.__debugMode == 1:
       gLogger.setLevel( "VERBOSE" )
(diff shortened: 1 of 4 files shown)


atsareg:
I do not know why the ObjectLoader is a singleton, may be it was just "optimization". The changes done make singleton aspect meaningless. So, I would suggest just dropping the singleton property of the ObjectLoader as a much simpler solution.

chaen:
I don't really know either. My guess is that it was to give the initial parameters only once. I agree that it is maybe not the most elegant way, but I find it a bit safer than removing the singleton. However, we have quite many places in the code where this logic is re-implemented. I'd rather collect all these use cases and make one utility that embraces them all.

atsareg:
I think that singleton is a more rigid solution which should be made on purpose where you need a globally unique object. This is certainly not the case of the ObjectLoader which is a rather simple stateless (its state is in the CS) utility. So, more simpler solution, the better. Your fix is a good prove for that.

atsareg:
Merged by hand from chaen/rel-v6r19_FIX_objectLoaderLogging because of conflicts

PR #3527, opened 2017-11-07
[v6r20] CI: add formatting job

andresailer:
This adds a job to the CI that will check that any changed lines or new files are formatted according to the agreed style.
The check only does something meaningful for PRs because it is impossible to figure out with what to diff the code otherwise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
     - CHECK=py.test
     - CHECK=docs
     - CHECK=pylint
+    - CHECK=format
 
 
 # command to install dependencies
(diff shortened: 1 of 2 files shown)


atsareg:
IIUC this will be applied for all the changed lines including "beautification". This risks to fail the Travis test too often, I am afraid. May be it is possible to make it a separate test from the rest of Travis enabled tests ?

andresailer:
I hope all future beautification is according to the formatting rules, that this test will fail otherwise is the whole point.

andresailer:
Rebased, so here is the job that failed the formatting test https://travis-ci.org/DIRACGrid/DIRAC/jobs/298479199 for the second commit in this PR

andresailer:
@fstagni test success is reported now

PR #3530, opened 2017-11-07
[v6r19] dirac-dms-user-lfn: always prepend '*' for wildcard

andresailer:
without a leading \* for the wildcard no matches were ever found unless the full path is specified or '\*' is prepended in the wildcard expression. This makes the wildcard more intuitive to use, will not affect anybody negatively. Full path will still match even with leading '\*'

BEGINRELEASENOTES

*DMS

CHANGE: dirac-dms-user-lfns: the wildcard flag will always assume leading "\*" to match files, unless the full path was specified in the wildcard no files were previously matched

ENDRELEASENOTES


phicharp:
Review OK

PR #3531, opened 2017-11-07
[v6r20] Allow command modules with underscores in the names

atsareg:
Joining the pylint rush :) . This fix allows to write commands with module names like dirac_xxx_yyy.py. While installation the command wrappers will be anyway generated as dirac-xxx-yyy, so no surprises for the users. The modules can be renamed except for dirac-install.py and dirac-deploy-scripts.py for the moment. Once the new versions are in place ( in a release ), original dirac-install.py and dirac-deploy-scripts.py can be also renamed.

BEGINRELEASENOTES
*Core
CHANGE: dirac-deploy-scripts, dirac-install - allow command modules with underscores in their names in order for better support for the code checking tools 
ENDRELEASENOTES

### DIFF
--- a/Core/scripts/dirac-deploy-scripts.py
+++ b/Core/scripts/dirac-deploy-scripts.py
@@ -123,7 +123,7 @@ if not rootPath:
   sys.exit( 1 )
 
 targetScriptsPath = os.path.join( rootPath, "scripts" )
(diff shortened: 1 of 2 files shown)


chaen:
Cool :)
Did you also see https://github.com/DIRACGrid/DIRAC/issues/3523 ?

PR #3573, opened 2017-12-12
[v6r19] Add new fields in TransformationMonitor, fix ReplicateAndRegister

phicharp:
BEGINRELEASENOTES
* TransformationSystem
CHANGE: add possibility to get extension-specific tasks and files statuses in TransformationMonitor web application

* RMS
NEW: add option --ListJobs to dirac-rms-request to list the jobs for a set of requests

* DMS
FIX: in DMSHelpers, fix case when no site is found for an SE
FIX: in ReplicateAndRegister, don't try and get SE metadata is replica is inactive

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
+++ b/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
@@ -24,6 +24,7 @@ __RCSID__ = "$Id $"
 
 # # imports
 import re
+from collections import defaultdict
 # # from DIRAC
 from DIRAC import S_OK, S_ERROR, gLogger
 from DIRAC.Core.Utilities.Adler import compareAdler, hexAdlerToInt, intAdlerToHex
@@ -273,7 +274,7 @@ class ReplicateAndRegister(DMSRequestOperationsBase):
 
     toSchedule = {}
 
-    errors = {}
+    errors = defaultdict(int)
     for opFile in self.getWaitingFilesList():
       opFile.Error = ''
       gMonitor.addMark("FTSScheduleAtt")
@@ -299,21 +300,21 @@ class ReplicateAndRegister(DMSRequestOperationsBase):
         gMonitor.addMark("FTSScheduleFail")
         if noMetaReplicas:
           err = "Couldn't get metadata"
-          errors[err] = errors.setdefault(err, 0) + 1
+          errors[err] += 1
           self.log.verbose(
               "unable to schedule '%s', %s at %s" %
               (opFile.LFN, err, ','.join(noMetaReplicas)))
           opFile.Error = err
         elif noReplicas:
           err = "File doesn't exist"
-          errors[err] = errors.setdefault(err, 0) + 1
+          errors[err] += 1
           self.log.error("Unable to schedule transfer",
                          "%s %s at %s" % (opFile.LFN, err, ','.join(noReplicas)))
           opFile.Error = err
           opFile.Status = 'Failed'
         elif badReplicas:
           err = "All replicas have a bad checksum"
-          errors[err] = errors.setdefault(err, 0) + 1
+          errors[err] += 1
           self.log.error("Unable to schedule transfer",
                          "%s, %s at %s" % (opFile.LFN, err, ','.join(badReplicas)))
           opFile.Error = err
@@ -403,17 +404,17 @@ class ReplicateAndRegister(DMSRequestOperationsBase):
       self.log.info("Trying transfer using replica manager as FTS failed")
     else:
       self.log.info("Transferring files using Data manager...")
-    errors = {}
+    errors = defaultdict(int)
     for opFile in waitingFiles:
       if opFile.Status == 'Failed':
         err = "File already Failed"
-        errors[err] = errors.setdefault(err, 0) + 1
+        errors[err] += 1
         continue
       if opFile.Error in ("Couldn't get metadata",
                           "File doesn't exist",
                           "All replicas have a bad checksum",):
         err = "File already in error status"
-        errors[err] = errors.setdefault(err, 0) + 1
+        errors[err] += 1
         continue
 
       gMonitor.addMark("ReplicateAndRegisterAtt", 1)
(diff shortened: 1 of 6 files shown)


phicharp:
@zmathe : Zoltan, is it sufficient to add these parameters in the method `getTransformationSummaryWeb` for them to be included in the web display? I suspect there is another place where the columns are defined, right?

zmathe:
The new fields have to be added also to WebApp. I will modify later today.

phicharp:
OK, but can these fields also be taken from the CS or does it need to be hardcoded, which is not very nice...

zmathe:
It has to be hardcoded: https://github.com/DIRACGrid/WebAppDIRAC/blob/integration/WebApp/static/DIRAC/TransformationMonitor/classes/TransformationMonitor.js#L40

phicharp:
Too bad... However the statuses 'NotProcessed' and 'Removed' are LHCb-specific, therefore should be in LHCbDIRAC...

zmathe:
The dataFields can be overwritten in LHCbWeb

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -29,203 +29,208 @@
...
+    allReplicas = dataManager.getReplicas(opFile.LFN, getUrl=False)
     if allReplicas['OK']:
-      allReplicas = allReplicas['Value']['Successful'].get( opFile.LFN, {} )
+      allReplicas = allReplicas['Value']['Successful'].get(opFile.LFN, {})
       if not allReplicas:
-        ret['NoReplicas'].append( None )
+        ret['NoReplicas'].append(None)
         noReplicas = True
       else:
-        # We try inactive replicas to see if maybe the file doesn't exist at all
```
phicharp: here is the change

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -29,203 +29,208 @@
...
     if fcChecksum:
-      if hexAdlerToInt( fcChecksum ) != False:
+      if hexAdlerToInt(fcChecksum) is not False:
         opFile.Checksum = fcChecksum
-        opFile.ChecksumType = fcMetadata['Value']['Successful'][opFile.LFN].get( 'ChecksumType', 'Adler32' )
+        opFile.ChecksumType = fcMetadata['Value']['Successful'][opFile.LFN].get('ChecksumType', 'Adler32')
       else:
         opFile.Checksum = None
 
+  # If no replica was found, return what we collected as information
```
phicharp: ... and return if no replica is found

chaen:
review ok

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -239,42 +244,41 @@ def _addMetadataToFiles( self, toSchedule ):
...
-      return S_OK( "%s targets are banned for writing" % ",".join( bannedTargets['Value'] ) )
+      return S_OK("%s targets are banned for writing" % ",".join(bannedTargets['Value']))
 
     # Can continue now
-    self.log.verbose( "No targets banned for writing" )
+    self.log.verbose("No targets banned for writing")
 
     toSchedule = {}
 
+    errors = {}
```
andresailer:
```
from collections import defaultdict
errors = defaultdict(int)
```

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -283,218 +287,238 @@ def ftsTransfer( self ):
...
-          toSchedule[opFile.LFN] = [ opFile, validReplicas, validTargets ]
+          toSchedule[opFile.LFN] = [opFile, validReplicas, validTargets]
       else:
-        gMonitor.addMark( "FTSScheduleFail" )
+        gMonitor.addMark("FTSScheduleFail")
         if noMetaReplicas:
-          self.log.warn( "unable to schedule '%s', couldn't get metadata at %s" % ( opFile.LFN, ','.join( noMetaReplicas ) ) )
-          opFile.Error = "Couldn't get metadata"
+          err = "Couldn't get metadata"
+          errors[err] = errors.setdefault(err, 0) + 1
```
andresailer: when using defaultdict
```
errors[err] += 1
```

phicharp: Yes, thanks for reminding me, this is a great feature! I'll do it

phicharp:
Please review... hopefully that's it for this week!

phicharp:
Diffs: https://github.com/DIRACGrid/DIRAC/pull/3573/files?w=1

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -239,262 +245,318 @@ def _addMetadataToFiles( self, toSchedule ):
...
 
+    delayExecution = 0
+    errors = defaultdict(int)
     for opFile in self.getWaitingFilesList():
       opFile.Error = ''
-      gMonitor.addMark( "FTSScheduleAtt" )
+      gMonitor.addMark("FTSScheduleAtt")
       # # check replicas
-      replicas = self._filterReplicas( opFile )
+      replicas = self._filterReplicas(opFile)
```
atsareg: Code for evaluating filtered replicas is similar in ftsTransfer and in dmTransfer. Might be worth making it a common utility

RequestManagementSystem/Client/ReqClient.py
```diff
@@ -402,125 +406,138 @@ def resetFailedRequest( self, requestID, allR = False ):
...
     key += ': '
   blanks = offset * ' '
-  if mainItem and type( mainItem ) == type( {} ):
-    output += "%s%s%s\n" % ( blanks, key, '{' ) if blanks or key else ''
-    for key in sorted( mainItem ):
-      prettyPrint( mainItem[key], key = key, offset = offset )
-    output += "%s%s\n" % ( blanks, '}' ) if blanks else ''
-  elif mainItem and type( mainItem ) == type( [] ) or type( mainItem ) == type( tuple() ):
-    output += "%s%s%s\n" % ( blanks, key, '[' if type( mainItem ) == type( [] ) else '(' )
+  if mainItem and isinstance(mainItem, type({})):
```
atsareg: There are multiple replacements for isinstance() which is good but better to make it properly, I mean:

   isinstance(mainItem, type({})) -> isinstance(mainItem, dict)

And similar for other replacements in this file

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
phicharp: It is the same method that is called...

RequestManagementSystem/Client/ReqClient.py
phicharp: Ah! This is done automatically by autopep8! I though it was doing it properly, probably though only if the original code is `if type(x) == types.dict`.
I will fix them by hand

PR #3574, opened 2017-12-12
[v6r19] Fix inheritance of SRM2 protocols List

chaen:
Because of some hardcoded path, the protocols list for SRM2 was not correctly inherited.
This PR fixes this issue.

It also makes the Stager survive inheritance
I tested the SRM change with our integration tests

BEGINRELEASENOTES

*Resources
FIX: Use parameters given at construction for SRM2 protocols List

*StorageManagement
FIX: use StorageElement object to get disk cache size

ENDRELEASENOTES

### DIFF
--- a/Resources/Storage/GFAL2_SRM2Storage.py
+++ b/Resources/Storage/GFAL2_SRM2Storage.py
@@ -47,6 +47,16 @@ class GFAL2_SRM2Storage( GFAL2_StorageBase ):
     self.__setSRMOptionsToDefault()
 
 
+    # This lists contains the list of protocols to ask to SRM to get a URL
+    # It can be either defined in the plugin of the SE, or as a global option
+    if 'ProtocolsList' in parameters:
+      self.protocolsList = parameters['ProtocolsList'].split(',')
+    else:
+      self.log.debug( "GFAL2_SRM2Storage: No protocols provided, using the default protocols." )
+      self.protocolsList = self.defaultLocalProtocols
+      self.log.debug( 'GFAL2_SRM2Storage: protocolsList = %s' % self.protocolsList )
+
+
 
   def __setSRMOptionsToDefault( self ):
     ''' Resetting the SRM options back to default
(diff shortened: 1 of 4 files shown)


phicharp:
Review OK

andresailer:
What are the errors that are fixed by this PR?
We also have some issues when using GFAL2_SRM2 that I don't understand, but maybe this will solve them.

chaen:
Ah true, I wanted to put them ! 
The error is triggered when you have a StorageElement (srm based) that inherits from another StorageElement. As of v6r19, the Base SE should be in the dedicated section (StorageElementBases).
However, one parameter was still looked for in StorageElement.

Example:
```
StorageElementBases
{
  Chris-Disk
  {
    [...]
  }
}
StorageElements
{
  Chris-USER
  {
    BaseSE = Chris-Disk
    [...]
  }
}
```

The following call will fail:

```

[chaen@localvm6 etc]$ cat /tmp/testInh.py 
parseCommandLine()

from DIRAC.Resources.Storage.StorageElement import StorageElement

se = StorageElement('CHRIS-USER')
print se.getURL('/lhcb/user/c/chaen/zozo.xml')
```
```
[chaen@localvm6 etc]$ python /tmp/testInh.py 
StorageFactory._getStorageOptions: Failed to get storage status CHRIS-USER: Cannot get URL for ResourceStatus/ResourceStatus in setup Test-Setup: RuntimeError('Option /DIRAC/Setups/Test-Setup/ResourceStatus is not defined',)
StorageFactory._getStorageOptions: Failed to get storage options. CHRIS-Disk: Path /Resources/StorageElements/CHRIS-Disk does not exist or it's not a section
  File "/tmp/testInh.py", line 11, in <module>
    print se.getURL('/lhcb/user/c/chaen/zozo.xml')
  File "/home/chaen/dirac/DIRAC/Resources/Storage/StorageElement.py", line 738, in getURL
    protocols = self.turlProtocols
  File "/home/chaen/dirac/DIRAC/Resources/Storage/StorageElement.py", line 1083, in __getattr__
    raise AttributeError( "StorageElement does not have a method '%s'" % name )
AttributeError: StorageElement does not have a method 'turlProtocols'
```

What errors do you have ? Can you send them to me (by email maybe, or open an issue)

andresailer:
Thanks, this relates to StorageElementBases. I have other problems (discussing via slack).

PR #3586, opened 2017-12-23
[v6r20] Print DEncode traceback information

chaen:
BILD SPOILER ALERT:
we want to be able to spot where a serialization that would fail or be incorrect in json happens in DISET.
There are currently 2 known cases: tuples, dictionaries whose keys are numbers.
This PR adds a function that will print the callstack as well as the arguments of the call when such cases are met. The output looks something like that:
```
============================================= Encoding tuples =============================================
  File "/tmp/ping.py", line 2, in <module>
    print RPCClient('dips://lbtestvobox.cern.ch:9197/DataManagement/FileCatalog').ping()

  File "/home/chaen/dirac/DIRAC/Core/DISET/RPCClient.py", line 43, in __call__
    return self.__doRPCFunc( self.__remoteFuncName, args )

  File "/home/chaen/dirac/DIRAC/Core/DISET/RPCClient.py", line 88, in __doRPC
    return self.__innerRPCClient.executeRPC( sFunctionName, args )

  File "/home/chaen/dirac/DIRAC/Core/DISET/private/InnerRPCClient.py", line 46, in executeRPC
    retVal = self._proposeAction( transport, ( "RPC", functionName ) )

  File "/home/chaen/dirac/DIRAC/Core/DISET/private/BaseClient.py", line 527, in _proposeAction
    retVal = transport.sendData( S_OK( stConnectionInfo ) )

  File "/home/chaen/dirac/DIRAC/Core/DISET/private/Transports/BaseTransport.py", line 120, in sendData
    sCodedData = DEncode.encode( uData )

  File "/home/chaen/dirac/DIRAC/Core/Utilities/DEncode.py", line 313, in encode
    g_dEncodeFunctions[ type( uObject ) ]( uObject, eList )

  File "/home/chaen/dirac/DIRAC/Core/Utilities/DEncode.py", line 290, in encodeDict
    g_dEncodeFunctions[ type( dValue[key] ) ]( dValue[key], eList )

  File "/home/chaen/dirac/DIRAC/Core/Utilities/DEncode.py", line 275, in encodeTuple
    g_dEncodeFunctions[ type( uObject ) ]( uObject, eList )

  File "/home/chaen/dirac/DIRAC/Core/Utilities/DEncode.py", line 268, in encodeTuple
    printDebugCallstack()

Calling frame: ('/home/chaen/dirac/DIRAC/Core/DISET/private/Transports/BaseTransport.py', 120)
With arguments {'uObject': {'OK': True,
             'Value': (('DataManagement/FileCatalog',
                        'LHCb-Certification',
                        'lhcb'),
                       ('RPC', 'ping'),
                       '')}}
====================================================================================================
```

It is disabled by default and only active when the environment variable  `DIRAC_DEBUG_DENCODE_CALLSTACK` is set to `True`

BEGINRELEASENOTES

*Core
NEW: Add conditional printout of the traceback when serializing/deserializing non json compatible object in DEncode (enabled with DIRAC_DEBUG_DENCODE_CALLSTACK environment variable)

ENDRELEASENOTES

### DIFF
--- a/Core/Utilities/DEncode.py
+++ b/Core/Utilities/DEncode.py
@@ -23,10 +23,9 @@ import traceback
 from pprint import pprint
 
 
-
 # Setting this environment variable will enable the dump of the debugging
 # call stack
-DIRAC_DEBUG_DENCODE_CALLSTACK=bool(os.environ.get('DIRAC_DEBUG_DENCODE_CALLSTACK', False))
+DIRAC_DEBUG_DENCODE_CALLSTACK = bool(os.environ.get('DIRAC_DEBUG_DENCODE_CALLSTACK', False))
 
 
 def printDebugCallstack():
@@ -47,18 +46,15 @@ def printDebugCallstack():
     # Get all the arguments of the call
     allArgs = inspect.getargvalues(frame)
     # Keep only the arguments that are parameters of the call, as well as their value
-    return dict([(argName, allArgs.locals[argName]) for argName in allArgs.args ])
-
+    return dict([(argName, allArgs.locals[argName]) for argName in allArgs.args])
 
   tb = traceback.format_stack()
   frames = inspect.stack(context=100)
-
 
   # print the traceback that leads us here
   # remove the last element which is the traceback module call
   for line in tb[:-1]:
     print line
-
 
   # Now we try to navigate up to the caller of dEncode.
   # For this, we find the frame in which we enter dEncode.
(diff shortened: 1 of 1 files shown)


chaen:
@fstagni @atsareg this should be well along the lines of what we discussed

chaen:
No courageous people to review ? :)

Core/Utilities/DEncode.py
```diff
@@ -16,236 +16,369 @@
...
+import os
+
+import inspect
+import traceback
+from pprint import pprint
+
+
+# Setting this environment variable will enable the dump of the debugging
+# call stack
+DIRAC_DEBUG_DENCODE_CALLSTACK = bool(os.environ.get('DIRAC_DEBUG_DENCODE_CALLSTACK', False))
```
andresailer: This means any non-empty value for the environment variable will enable the printing, not just 'True'

Core/Utilities/DEncode.py
```diff
@@ -16,236 +16,369 @@
...
+
+
+def printDebugCallstack():
+  """ Prints information about the current stack as well as the caller parameters.
+      The purpose of this method is to track down all the places in DIRAC that might
+      not survive the change to JSON encoding.
+
+      :returns: nothing, just prints
```
andresailer: I.e. `:returns: None`

Core/Utilities/DEncode.py
```diff
@@ -16,236 +16,369 @@
...
+
+        :returns: dict {param name: value}
+    """
+    # Get all the arguments of the call
+    allArgs = inspect.getargvalues(frame)
+    # Keep only the arguments that are parameters of the call, as well as their value
+    return dict([(argName, allArgs.locals[argName]) for argName in allArgs.args])
+
+  tb = traceback.format_stack()
+  frames = inspect.stack(context=100)
```
andresailer: why 100, is that the maximum stack level? or just not expected to be exceeded?

andresailer:
One should mention the environment variable in the release notes.

Core/Utilities/DEncode.py
chaen: Correct. I am assuming only sensible persons that would use only True or False :D What do you recommend ? That I keep this behavior and change the doc, or enforce the env variable to be True ?

Core/Utilities/DEncode.py
chaen: Hum, my bad, I should have put a name there. It's just that the default is 1, so you do not see far enough by default

Core/Utilities/DEncode.py
andresailer: Well exporting the environment variable to "False" would still enable it, because os.environ only contains strings as the values and any non-empty string is `True`. But I don't think the behaviour needs to be changed, just needs to be clear that exporting 0/False/Off still enables this feature.

chaen: Silly me. When codding at 5am, one tends to forget that bool does not do a lexical cast
```
In [3]: bool('False')
Out[3]: True
```

:D 
Okay, fixing all that. Thanks for the review

Core/Utilities/DEncode.py
```diff
@@ -16,236 +16,372 @@
...
+_dateType = type(_dateTimeObject.date())
+_timeType = type(_dateTimeObject.time())
 
 g_dEncodeFunctions = {}
 g_dDecodeFunctions = {}
 
-#Encoding and decoding ints
-def encodeInt( iValue, eList ):
-  eList.extend( ( "i", str( iValue ), "e" ) )
+# Encoding and decoding ints
```
andresailer: Maybe better to put this in the function docstring so autopep doesn't separate it from the function by two lines.

chaen: Sure. I can do that for all of them

PR #3596, opened 2018-01-16
[v6r19] Logging: add space after message only if varmessage is not empty

### LINKED ISSUE #3587: [v6r19] Logging: space added after message
If there is no variable message the new logging adds a space at the end of the message, this breaks the output from, e.g., dirac-dms-find-lfns when redirected to a file and then passing the file to, e.g., dirac-dms-lfn-replicas. Because the last lfn has an additional space in the line it cannot be found.

andresailer:
BEGINRELEASENOTES

*Framework
FIX: Remove the space after log messages if no variable message is printed, fixes #3587 

ENDRELEASENOTES

- [x] Need to fix the tests that expect additional space

### DIFF
--- a/FrameworkSystem/private/standardLogging/Logging.py
+++ b/FrameworkSystem/private/standardLogging/Logging.py
@@ -412,6 +412,7 @@ class Logging(object):
       # not either.
       extra = {'componentname': self._componentName,
                'varmessage': sVarMsg,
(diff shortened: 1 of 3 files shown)


chaen:
Kudo !!

andresailer:
Ok, fixed the tests, added varmessage to one of the comparisons

PR #3638, opened 2018-03-14
[v6r20] dirac-external-requirements restored

atsareg:
The dirac-external-requirements command is restored. It is called in the end of the dirac-install process to collect all the extra python modules requested in the DIRAC package and in the extension packages. It allows to define extra python module dependencies in the releases.cfg of each package
and not put it into the Externals. 

BEGINRELEASENOTES
*Core
CHANGE: dirac-external-requirements - reimplemented to use preinstalled pip command rather than 
                 the pip python API 
ENDRELEASENOTES

### DIFF
--- a/Core/scripts/dirac-externals-requirements.py
+++ b/Core/scripts/dirac-externals-requirements.py
@@ -25,13 +25,16 @@ __RCSID__ = "$Id$"
 # Default installation type
 instType = "server"
 
+
 def setInstallType(val):
   global instType
   instType = val
   return S_OK()
 
+
 Script.registerSwitch("t:", "type=", "Installation type. 'server' by default.", setInstallType)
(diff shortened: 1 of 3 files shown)


chaen:
This means that the compilation will happen on every client ?

atsareg:
This is meant for pip installable modules. If pip involves compilation, then yes, each client will compile. If it is a pure python, then it is quite light. Although I do not think that this mechanism should be used for client/pilot installations. Rather for the server ones. This would allow not to specify python modules in the Externals making it more stable

andresailer:
Could you add the "RequiredExternals section"  to the documentation, please: http://dirac.readthedocs.io/en/latest/DeveloperGuide/DevelopmentModel/DIRACProjects/index.html#releases-schema

PR #3664, opened 2018-04-18
[v6r20] FTS3DB: fix filter statements, add kick and delete functionality

andresailer:
I our development instance no more FTS3 jobs were submitted. Turns out the filter statements were changed for pylint. the "is None" comparison results in the statement "WHERE false = 1"

https://stackoverflow.com/questions/16093475/flask-sqlalchemy-querying-a-column-with-not-equals#16093713

BEGINRELEASENOTES

*DMS

FIX: FTS3DB: sqlalchemy filter statements with "is None" do not work and result in no lines being selected
NEW: FTS3Agent and FTS3DB: add functionality to kick stuck requests and delete old requests

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/DB/FTS3DB.py
+++ b/DataManagementSystem/DB/FTS3DB.py
@@ -476,7 +476,7 @@ class FTS3DB(object):
 
       ftsOps = session.query(FTS3Operation.operationID)\
           .filter(FTS3Operation.lastUpdate < (func.date_sub(func.utc_timestamp(),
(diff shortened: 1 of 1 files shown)


andresailer:
I added the `.distinct()` for the operationsID query

andresailer:
Following close discussions with @chaen via slack, added functionality for kicking and deleting

atsareg:
@chaen: final approval ?

PR #3676, opened 2018-05-03
[v6r20] FTS3: add monitoring

chaen:
I realized that we do not have any monitoring info in the new FTS3 systems.
@andresailer can you please review that one ? 

BEGINRELEASENOTES
*DMS
NEW: add accounting report in the FTS3Agent
ENDRELEASENOTES


fstagni:
Is it absolutely necessary that this is part of v6r20? If possible, I would prefer not.

chaen:
@fstagni I am afraid it is, otherwise we are totally blind. I am surprised I overlooked that at the time.

DataManagementSystem/Client/FTS3Job.py
```diff
@@ -441,29 +488,82 @@ def submit( self, context = None, ftsServer = None, ucert = None, pinTime = 3600
...
+    for fileDict in filesInfoList:
+      file_state = fileDict['file_state'].capitalize()
+      if file_state in FTS3File.FTS_SUCCESS_STATES:
+        successfulFiles.append(fileDict)
+
+    job_metadata = jobStatusDict['job_metadata']
+    # previous version of the code did not have dictionnary as
+    # job_metadata
+    if isinstance(job_metadata, dict):
+      sourceSE = jobStatusDict['job_metadata'].get('sourceSE')
```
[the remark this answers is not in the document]
chaen: which `job_metadata` ?

DataManagementSystem/Client/FTS3File.py
```diff
@@ -1,42 +1,45 @@
...
+                'Active',  # From FTS: File went active
+                'Finished',  # From FTS: File finished gracefully
+                'Canceled',  # From FTS: Canceled by the user
+                'Staging',  # From FTS: When staging of a file is requested
+                'Failed',  # From FTS: File failure
+                'Defunct',  # Totally fail, no more attempt will be made
+                'Started',  # From FTS: File transfer has started
+                ]
+
   FINAL_STATES = ['Canceled', 'Finished', 'Defunct']
```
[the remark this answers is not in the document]
chaen: I just added a comment on that

DataManagementSystem/Client/FTS3Job.py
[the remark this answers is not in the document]
andresailer: the dictionary that is used for the `isinstance` check, just another "name" for the same thing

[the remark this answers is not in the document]
chaen: ah yes ! Thanks

chaen:
@andresailer is it ok now ?

andresailer:
I didn't see any entries in the accounting and added a printout for the result from commit().
This is what I get
```
2018-05-16 08:12:29 UTC DataManagement/FTS3Agent/Account WARN: Failed to commit accounting Invalid values: no value for RegistrationTime, value for key RegistrationTime is not numerical type, no value for Regist
rationOK, value for key RegistrationOK is not numerical type, no value for RegistrationTotal, value for key RegistrationTotal is not numerical type
```

chaen:
Hum, copy paste is never good :-( Apologies, I will fix it right now

chaen:
voila

andresailer:
Thanks. I pulled the update.

andresailer:
```
2018-05-16 13:55:04 UTC DataManagement/FTS3Agent/Account INFO: Succeeded to commit accounting
```
image

chaen:
@andresailer this should do with what we discussed this morning

DataManagementSystem/private/FTS3Utilities.py
```diff
@@ -96,9 +99,50 @@ def generatePossibleTransfersBySources( ftsFiles, allowedSources = None ):
...
+
+
+def selectUniqueRandomSource(ftsFiles, allowedSources=None):
+  """
+      For a list of FTS3files object, select a random source, and group the files by source.
+
+      :param allowedSources : list of allowed sources
+      :param ftsFiles : list of FTS3File object
+
+      :return  S_OK({ sourceSE: [ FTS3Files] })
```
andresailer: `:return:  S_OK...` (second colon missing)

DataManagementSystem/private/FTS3Utilities.py
```diff
@@ -96,9 +99,50 @@ def generatePossibleTransfersBySources( ftsFiles, allowedSources = None ):
...
+  for ftsFile in ftsFiles:
+
+    if ftsFile.lfn in filteredReplicas['Failed']:
+      _log.error("Failed to get active replicas", "%s,%s" %
+                 (ftsFile.lfn, filteredReplicas['Failed'][ftsFile.lfn]))
+      continue
+
+    replicaDict = filteredReplicas['Successful'][ftsFile.lfn]
+
+    # pick a a random source
```
andresailer: `pick a random source` (one 'a' too much)

DataManagementSystem/private/FTS3Utilities.py
```diff
@@ -96,9 +99,50 @@ def generatePossibleTransfersBySources( ftsFiles, allowedSources = None ):
...
+  # For all files, check which possible sources they have
+  res = _checkSourceReplicas(ftsFiles)
+  if not res['OK']:
+    return res
+
+  filteredReplicas = res['Value']
+
+  for ftsFile in ftsFiles:
+
+    if ftsFile.lfn in filteredReplicas['Failed']:
```
andresailer: If the file no longer exists, can we set the ftsFile.status to Defunct so the job gets cleaned out of the system?

chaen: A non existing file will be filtered already at the previous step, when grouping by target. I will check what we do when this happens. However in principle it should not happen, since it would already be filtered out when going from RMS to FTS

DataManagementSystem/Agent/FTS3Agent.py
```diff
@@ -445,5 +451,18 @@ def execute(self):
...
+
+        :param ftsJob: the FTS3Job from which we send the accounting info
+    """
+
+    dataOp = DataOperation()
+    dataOp.setStartTime(fromString(ftsJob.submitTime))
+    dataOp.setEndTime(fromString(ftsJob.lastUpdate))
+
+    dataOp.setValuesFromDict(ftsJob.accountingDict)
+    dataOp.commit()
```
andresailer: Would it be better to use "delayeCommit" to reduce the amount of queries to the DataStoreClient?

DataManagementSystem/private/FTS3Utilities.py
chaen: Hum, no such filtering at the moment WITHIN FTS, but as I said, it should never enter FTS then. Would you prefer me to add more checks for that ?

DataManagementSystem/Agent/FTS3Agent.py
chaen: Can do ! I did not even know about that

DataManagementSystem/private/FTS3Utilities.py
andresailer: I had it happen that some file was no longer existing and the Operation/Job got stuck in a cycle.
It could be that that was caused by the debugging activities.

chaen:
Great ! It is also a good test for staging :)

andresailer:
Don't worry the staging part was thoroughly tested already :)

DataManagementSystem/private/FTS3Utilities.py
andresailer: The problem occurred like this:
Created a production to move files from A to B
Some files always failed, because they were not correctly stored on A, the file size was 0. 
I deleted the file. Thus it was already in the FTS system but no longer existing.

atsareg:
Are you converging on this PR ? Can it be approved and merged ?

chaen:
I am not sure I can do the last request as a straight forward thing, so if @andresailer is ok I would merge it just like it is, and try to implement what he wants a bit later (so we can release it for the workshop)

PR #3680, opened 2018-05-09
[v6r20] Doc: add a documented dirac.cfg

chaen:
When looking at big projects outside, you can see a full configuration files, with all the options documented. I think it is a very nice way of doing. 
Since I am exploring the code to write doc for the DUW, I took the opportunity to start such a file. 
I know, it's work to keep it up to date. But I think it is totally worth it. 
What do you think ? 

BEGINRELEASENOTES

*Docs
NEW: add a documented dirac.cfg 

ENDRELEASENOTES

### DIFF
--- a/dirac.cfg
+++ b/dirac.cfg
@@ -1,0 +1,234 @@
+Systems
+{
+   DataManagementSystem
+   {
+      Agents
+      {
+         #http://dirac.readthedocs.io/en/latest/AdministratorGuide/Systems/DataManagement/fts.html#cleanftsdbagent
+         # Used to clean the database
+         CleanFTSDBAgent
+         {
+            DeleteGraceDays = 21 # time after which deleting a job in a final status
+            DeleteLimitPerCycle = 100 # max number of jobs to delete per agent cycle
+         }
+         # http://dirac.readthedocs.io/en/latest/AdministratorGuide/Systems/DataManagement/fts.html#ftsagent
+         # Agent to perform the fts transfers
+         FTSAgent
+         {
+
+            MaxFilesPerJob = 10 # maximum number of files in a single fts job
+            MaxRequests = 1000 # maximum number of requests to look at per agent's cycle
+            MaxThreads = 60 # maximum number of threads
+            MaxTransferAttempts = 256 # maximum number of time we attempt to transfer a file
+            MinThreads = 10 # minimum number of threads
+            MonitoringInterval = 1800 # interval between two monitoring of an FTSJob in second
+            PinTime = 18000 # when staging, pin time requested in the FTS job in second
+            ProcessJobRequests = True # if this agent is meant to process job only transfers (see `http://dirac.readthedocs.io/en/latest/AdministratorGuide/Systems/DataManagement/fts.html#multiple-ftsagents)
+         }
+      }
+      Services
+      {
+         # http://dirac.readthedocs.io/en/latest/AdministratorGuide/Systems/DataManagement/dfc.html#filecataloghandler
+         FileCatalogHandler
+         {
+            DatasetManager = DatasetManager
+            DefaultUmask = 0775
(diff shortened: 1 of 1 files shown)


atsareg:
Yes, documenting the configuration of all the DIRAC components is very important. What you have put here is probably a good way to do that, I mean providing the documentation as a commented example of the configuration files ( in cfg format ). However, If everything will go into one file it will be huge and difficult to find relevant pieces. 
I suggest that we do the following using this PR as a starting point. In each Administrator Guide/Systems/<System_Name> chapter we introduce subchapters Services, Agents, Databases, Executors in which we include subsubchapters per component following a standard template: short description of the functionality, dependencies and a commented cfg format configuration with all the possible options, defaults and examples. 
Similar documentation can be provided for the Resources with the docs following the configuration structure. And for the Operations as well.
So, the Administrator guide will be structures almost exactly as the Configuration for these three chapters.

chaen:
In fact, I think it is quite the opposite: having a single file with all existing option is a wonderful starting point. For example, I want to change something related to a protocol, so I would grep for protocol on this file, and this will give me all the places in the documentation where I find this. Then, using the links in the comment, I can narrow down to what I need. At least, this is how I use such configuration files. 

As for structuring the Admin doc like the code/configuration, it does not sound like a good idea to me because there is no logic. I would be more tempted to go the other way around, e.g. "this is the DataManagement system, it is used to do this and that. if you want functionality X, you need to have agents A,B and C and services U,V and W".

atsareg:
There is certainly a logic in how the DIRAC code is structured. It is not by chance that there is a DataManagementSystem subpackage in the code. This logical structuring is functional, so for administrators it is quite natural to find necessary bits following the first classification by System and then by component. You have a different logic in mind which also makes sense. However, mixing the two together can and will confuse administrators. The DIRAC code structure is also reflected in the Configuration structure. So, making the docs to follow the same logic will simply follow the already established logic. We can have another 'view" into the system "by functionality" which can be unstructured or made like a kind of FAQ collection. This can have additional explanations and links into above described documentation following the standard logic

chaen:
I think we agree on the overall. Then it is just a matter of where we put things. I would go for something like the following:

* Code documentation: perfect just as it is 
* Developer: I think we are getting there. We explain how to develop components, how dirac works under the hood, etc, and we link to Code documentation
* Administrator: here I think it should go by functionality, so effectively, by system as you say. But rather than having then a chapter per components, I would start from the functionality and drift down to the component, including the configuration options of that specific component. On that specific point, I would say I would completely get ride of the configuration section as it is now. I would keep a Configuration System section to explain what it is, what are the agents/services, and describe the global structure (Resources, Operations, etc) and what they are meant to contain, but I would not list the configuration of every components there. Of course, there should be general chapter on how to install, unrelated to system per se.
* HowTo/FAQ: targeted at end users, or administrator, which would start from stories "I want to replicate dataset", "I want to submit a job", etc. And there I would link to script documentation, and if for example the actions you want to perform requires a set of agents, I would link (in a 'what you need to be installed' section) to the Administrator doc. 

Ultimately, I think that an administrator should not even know how the code is structured :-)

chaen:
I propose we continue the discussion here, but are you OK with merging that in the meantime so I can add more on it while writing the doc ?

atsareg:
I think we agree and please continue working on it. I agree that the Configuration chapter in the Admin guide shoud be dropped as it is now and the corresponding material distributed in the per system chapters. This is actually what I proposed but was not very clear. And I like the idea to document the configuration as commented cfg format files rather than tables as it is now which are difficult to write and also to read.

chaen:
Great ! :-) I just have an issue with style checking, the new version of pycodestyle is extremely picky, and does not limit itself to modified files for some reason ! I'll check if I can fix it

chaen:
Hum, travis does not like retargeting branch. Let me close that one, I'll recreate it

PR #3688, opened 2018-05-16
[v6r20] FileCatalog - fix quotes after escaping input strings

atsareg:
Fixing a bug introduced when escaping input path strings

BEGINRELEASENOTES
*DMS
FIX: DirectoryLevelTree - fix bug when escaping input path strings
ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
@@ -34,7 +34,7 @@ class DirectoryLevelTree(DirectoryTreeBase):
     if not dpath['OK']:
       return dpath
     dpath = dpath['Value']
-    req = "SELECT DirID,Level from FC_DirectoryLevelTree WHERE DirName='%s'" % dpath
+    req = "SELECT DirID,Level from FC_DirectoryLevelTree WHERE DirName=%s" % dpath
     result = self.db._query(req, connection)
     if not result['OK']:
       return result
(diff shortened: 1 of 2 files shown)


DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
```diff
@@ -49,11 +49,13 @@ def findDir(self, path, connection=False):
...
     """ Find DirIDs for the given path list
     """
-    dpaths = ','.join(["'" + os.path.normpath(path) + "'" for path in paths])
-    dpaths = self.db._escapeString(dpaths)
-    if not dpaths['OK']:
-      return dpaths
-    dpaths = dpaths['Value']
+    dpathList = []
+    for path in paths:
+      dpath = self.db._escapeString(os.path.normpath(path))
```
chaen: but if you do not put quotes around, your `in` query will be like `in (a,b,c)` instead of `in ('a', 'b', 'c')` no ?

DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
```diff
@@ -253,12 +255,6 @@ def getDirectoryPaths(self, dirIDList):
       return S_OK({})
 
     dirListString = ','.join([str(d) for d in dirs])
-
-    dirListString = self.db._escapeString(dirListString)
```
chaen: Why do you remove this part ? You have no guarantee that it is indeed integers that you have there

atsareg: We escape user input strings, e.g. lfns that come through the client interface. Those are numerical IDs internal to the catalog, no need to escape them

DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
atsareg: _escapeString itself adds quotes around the resulting string. This was actually creating problems

DataManagementSystem/DB/FileCatalogComponents/DirectoryLevelTree.py
chaen: oh ok, sorry, I thought this method was exposed outside, my bad

atsareg:
Added errno.ENODATA, thanks for suggestion

PR #3711, opened 2018-06-07
[v6r19] Fix  DMSHelpers

phicharp:
BEGINRELEASENOTES

*DMS
FIX: if an SE is at no site, return None and not []
ENDRELEASENOTES


phicharp:
`
In [3]: dh.getLocalSiteForSE('Manchester-Test')
Out[3]: {'OK': True, 'Value': None}
`

PR #3712, opened 2018-06-07
[v6r19] Fix DMSHelpers

phicharp:
BEGINRELEASENOTES
*DMS
FIX: return None as LocalSite if an SE is at no site
ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Utilities/DMSHelpers.py
+++ b/DataManagementSystem/Utilities/DMSHelpers.py
@@ -292,8 +292,10 @@ class DMSHelpers(object):
   def getLocalSiteForSE(self, se):
     """ Get the site at which the SE is """
     sites = self._getLocalSitesForSE(se)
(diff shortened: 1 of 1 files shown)


phicharp:
PR redone as Travis was not happy the first time since the PR was on the wrong branch

chaen:
That is not good, it breaks the code at quite a few places, for example

```
            res = getSitesForSE(se, gridName='LCG')
            if res['OK'] and len(res['Value']) > 0:
```

phicharp:
I don't think so... This is not the same method... `getLocalSiteForSE()` returns a single site name or `None`

phicharp:
`In [3]: dh.getSitesForSE('NCBJ-BUFFER')`
`Out[3]: {'OK': True, 'Value': ['LCG.NCBJ.pl']}`

chaen:
Oups ! I thought from the diff that getSitesForSE was the modified method.

PR #3714, opened 2018-06-08
Fix zero value of CPU usage in monitoring

xianghuzhao:
The cpu usage plot in Component Monitoring always show zero value. This is because the Profiler instance is created and deleted in a single loop, and cpu usage could not be calculated unless it is called twice or more. Here I created a global dict to hold all Profiler instance persistently.

Although this works fine, I am still not certain that it is the correct way to use global variable in a service handler.

BEGINRELEASENOTES

*Framework
FIX: SystemAdministrator - Get the correct cpu usage data for each component

ENDRELEASENOTES


zmathe:
Maybe you can create the PR against v6r20 instead of integration.

PR #3726, opened 2018-06-20
[v6r20]handle exception when the result is not a string

zmathe:
BEGINRELEASENOTES

*TransformationSystem
FIX: If the result can not be evaluated, it can be converted to list

ENDRELEASENOTES

### DIFF
--- a/TransformationSystem/Agent/TransformationCleaningAgent.py
+++ b/TransformationSystem/Agent/TransformationCleaningAgent.py
@@ -206,7 +206,7 @@ class TransformationCleaningAgent(AgentModule):
         if not isinstance(res['Value'], list):
           try:
             transDirectories = ast.literal_eval(res['Value'])
-          except:
+          except Exception as _:
             # It can happen if the res['Value'] is '/a/b/c' instead of '["/a/b/c"]'
             transDirectories.append(res['Value'])
         else:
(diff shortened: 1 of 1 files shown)


TransformationSystem/Agent/TransformationCleaningAgent.py
```diff
@@ -204,7 +204,11 @@ def getTransformationDirectories(self, transID):
       transDirectories = []
       if res['Value']:
         if not isinstance(res['Value'], list):
-          transDirectories = ast.literal_eval(res['Value'])
+          try:
+            transDirectories = ast.literal_eval(res['Value'])
+          except:
+            # It can happen if the res['Value'] is '/a/b/c' instead of '["/a/b/c"]'
+            transDirectories = list(res['Value'])
```
sfayer: Hi,

Won't this result in transDirectories being set to ['/', 'a', '/', 'b', '/', 'c'] for the given example? Maybe it'd be safer to do `transDirectories = [ res['Value'] ]`?

Regards,
Simon

zmathe: Thank you Simon! You are right. It is corrected

atsareg:
Please, fix the formatting error, it is not really just formatting...

PR #3757, opened 2018-07-18
[v6r20] CI: generate self signed certificates TLS compliant

chaen:
This PR changes the way we generate self signed certificate. Things are now done "properly", following TLS standard. This makes them compatible with standard tools, that carefully obey the standard (tornado, to name only one).
Note that I kept the subjects on purpose, so all the CI tests do not need to change

To make it short:
* generate a CA certificate, puts it in the proper location, and generate the hash index 
* generate a host certificate, signed by the CA, which is allowed as a TLS server and client
* generate a client certificate, signed by the CA, which is allowed as a TLS client

By default, host and client certificates are valid for 1 day, but options can be passed 

maybe @sfayer wants to have a look, since he knows a lot about TLS :-) 

I tested it on the integration job of jenkins, and it went happily 

BEGINRELEASENOTES

*Tests
CHANGE: generate self signed certificate TLS compliant

ENDRELEASENOTES


sfayer:
Looks good to me... You can use environment variables in the openssl config files if you wanted to avoid having to re-write the file with sed each time. Something like:
`DNS.1 = ${ENV:fqan}` and then simply add an export on the line where you generate fqan in the parent script: `export fqdn=$((hostname...)`.

PR #3759, opened 2018-07-19
[v6r20] dirac-rms-request: fix check for FTSClient, silence error

andresailer:
In the past the FTSClient instantiation would throw an exception, this is no longer the case, so we ping the service, which fails if it doesn't exist.

Otherwise we always get this error message:
> Failed getFTSJobsForRequest URL for service DataManagement/FTSManager not found

Note that the FTS3Client does not have the equivalent functionality as far as I could see

BEGINRELEASENOTES
*RMS
FIX: dirac-rms-request: silence a warning, when not using the old FTS Services

ENDRELEASENOTES

### DIFF
--- a/RequestManagementSystem/Client/ReqClient.py
+++ b/RequestManagementSystem/Client/ReqClient.py
@@ -456,8 +456,12 @@ def printRequest(request, status=None, full=False, verbose=True, terse=False):
     if request.RequestID:
       from DIRAC.DataManagementSystem.Client.FTSClient import FTSClient
       ftsClient = FTSClient()
-  except Exception as e:
-    gLogger.debug("Could not instantiate FtsClient", repr(e))
+      res = ftsClient.ping()
+      if not res['OK']:
+        gLogger.debug("Could not instantiate FtsClient", res)
+        ftsClient = None
+  except ImportError as err:
+    gLogger.debug("Could not instantiate FtsClient because of Exception", repr(err))
 
   if full:
     output = ''


RequestManagementSystem/Client/ReqClient.py
```diff
@@ -452,12 +452,13 @@ def printRequest(request, status=None, full=False, verbose=True, terse=False):
...
 
   ftsClient = None
-  try:
-    if request.RequestID:
-      from DIRAC.DataManagementSystem.Client.FTSClient import FTSClient
-      ftsClient = FTSClient()
-  except Exception as e:
-    gLogger.debug("Could not instantiate FtsClient", repr(e))
+  if request.RequestID:
+    from DIRAC.DataManagementSystem.Client.FTSClient import FTSClient
```
chaen: The `try` block is needed because the client is not currently capable of instantiating the `fts3` library :(

andresailer: Really?
I thought it was there because of this line, which was dropped:
https://github.com/DIRACGrid/DIRAC/commit/6ffc86a54799c297d8c4fd3c59b16089f4c88b12#diff-ad259475f8d719f6783dc46e7a4503f1L66
I will a try/except block again.

chaen: Sadly not

```
In [1]: from DIRAC.DataManagementSystem.Client.FTSClient import FTSClient
<ipython-input-1-721f79090bc1> in <module>()
----> 1 from DIRAC.DataManagementSystem.Client.FTSClient import FTSClient

[...]
/cvmfs/lhcb.cern.ch/lib/lcg/releases/LCG_88/pytools/2.0/x86_64-slc6-gcc49-opt/lib/python2.7/site-packages/M2Crypto-0.25.1-py2.7-linux-x86_64.egg/M2Crypto/util.py in <module>()
     18 from M2Crypto import m2, six
     19 if py27plus:
---> 20     from typing import AnyStr, Tuple, Union  # noqa
     21     # see https://github.com/python/typeshed/issues/222
     22     AddrType = Union[Tuple[str, int], str]

ImportError: No module named typing
```

andresailer:
```
>>> import M2Crypto, DIRAC.DataManagementSystem.Client.FTSClient, fts3
>>> print "Works for me",  fts3.__file__, M2Crypto.__file__
Works for me /cvmfs/clicdp.cern.ch/DIRAC/Linux_x86_64_glibc-2.12/lib/python2.7/site-packages/fts3/__init__.py /cvmfs/clicdp.cern.ch/DIRAC/Linux_x86_64_glibc-2.12/lib/python2.7/site-packages/M2Crypto/__init__.py
```

chaen: yes, the bundle works, not the cvmfs installation. I tell you, it is misery...
thanks for putting it back

PR #3762, opened 2018-07-23
[v6r20] GetReleaseNotes: add --repo option

andresailer:
BEGINRELEASENOTES

*Docs
NEW: add --repo option, e.g. --repo DiracGrid/DiracOS, or just --repo DiracOS, fixes DIRACGrid/DIRACOS#30

ENDRELEASENOTES

```
$ ./GetReleaseNotes.py  -r DIRACGRID/diracos --branches=master --openPRs
[master]

NEW: (#29) requirements.txt copied from DIRAC
CHANGE: (#29) allow to have pipRequirements a URL or a file
```

### DIFF
--- a/docs/Tools/GetReleaseNotes.py
+++ b/docs/Tools/GetReleaseNotes.py
@@ -5,6 +5,7 @@
 from collections import defaultdict
 from datetime import datetime, timedelta
 import argparse
+from pprint import pformat
 import logging
 import textwrap
 import requests
@@ -45,7 +46,7 @@ def req2Json(url, parameterDict=None, requestType='GET'):
     log.error("Unable to access API: %s", req.text)
     raise RuntimeError("Failed to access API")
 
-  log.debug("Result obtained: %s", req.text)
+  log.debug("Result obtained:\n %s", pformat(req.json()))
   return req.json()
 
 
@@ -139,9 +140,8 @@ def collateReleaseNotes(prs):
           systemChangesDict[system].append(line)
 
     for system, changes in systemChangesDict.iteritems():
-      if not system:
-        continue
-      releaseNotes += "*%s\n\n" % system
+      if system:
+        releaseNotes += "*%s\n\n" % system
       releaseNotes += "\n".join(changes)
       releaseNotes += "\n\n"
     releaseNotes += "\n"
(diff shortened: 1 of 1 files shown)


atsareg:
Very nice. I suggest also to add the repo name to the generated notes to see that this is not from the DIRAC project. e.g.:
    FIX: (DIRACOS #12345) release note text... 

./GetReleaseNotes.py  -r DIRACGRID/diracos  -> will it work ? Is it case insensitive ? Shouldn't be ./GetReleaseNotes.py  -r DIRACGrid/DIRACOS ?

chaen:
Thanks ! 

@atsareg The idea is to use this script against the DIRACOS repo to put release notes in there directly, so I am not sure we want the repo name printed

atsareg:
I see. But I would be happy to include those notes also to the related DIRAC overall release. We do not do it now (not so many reasonable cases but still, WebApp, Pilot, now DIRACOS), which is pity. For the moment these packages are released without notes at all although they are supposed to be release together with the DIRAC  project combo.

chaen:
Hum, it's a bit delicate in the sense that the releases are decoupled. When would you put the release notes regarding VMDIRAC in DIRAC ? When the release of VMDIRAC is done, or when you start using it in DIRAC ? 
In the same way, we do not write the change log of the external libraries.
I would be in favour though of linking the external release notes when we do change version though. 
For example when releasing dirac vXYZ, in the changelog there could be a line saying

`use of DIRACOS UVW: link to changelog`

atsareg:
VMDIRAC, WebAppDIRAC are always released together with DIRAC. Pilot or DIRACOS might be different although not necessarily. So, for the former it makes perfect sense, for the latter it does not harm.

chaen:
No, it does not harm. But when will you put them in ?

andresailer:
Re https://github.com/DIRACGrid/DIRAC/pull/3762#issuecomment-407032000: The group and repository is case insensitive.

We could set up readthedocs pages for all projects, which could be not much more than the release notes, then the pages get automatically created and one can link to them without much problem.

atsareg:
I can merge the release notes by hand in the first place if they exist in the PRs using this tool. In this case, the notes will be also available via an html URL that can be linked from the documentation pages.

I agree with André that setting up readthedocs for all the packages would be very good. The docs can be organized similarly to the DIRAC one. I was thinking how this can be merged into just one docs web page but this is virtually impossible. So, having separate linked readthedocs projects is the only reasonable solution.

atsareg:
In any case, I am merging tis PR now. We can continue discussion further.

PR #3768, opened 2018-07-27
[V6r20] Fixes for Glue2 at CERN and some other inconsistencies

andresailer:
BEGINRELEASENOTES
*Core
FIX: The Glue2 parsing handles some common issues more gracefully
   * handle multiple execution environments at a single computing share (i.e., CERN)
   * handle multiple execution environments with the same ID (e.g., SARA)
   * handle cases where the execution environment just does not exist, use sensible dummy values in this case (many sites)

CHANGE: some print outs are prefixed with "SCHEMA PROBLEM", which seem to point to problems in the published information, i.e. keys pointing to non-existent entries, or non-unique IDs

ENDRELEASENOTES


andresailer:
@fstagni @atsareg: given that the GLUE2 information will not be usable for most sites any time soon (I guess), I can add an option to the Bdii2CSAgent to work only for a subset of sites, then one can use it at least for the sites where only GLUE2 information exists.

andresailer:
Referencing the original issue #3633

PR #3813, opened 2018-08-28
getLocalSitesForSE not to crash with a stray SE

iueda:
Just to avoid to have getLocalSitesForSE crashing when the specified SE does not belong to a site.
Don't ask why we have such a SE...

BEGINRELEASENOTES

*DataManagementSystem
FIX: getLocalSitesForSE not to crash when the specified SE does not belong to a site 

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Utilities/DMSHelpers.py
+++ b/DataManagementSystem/Utilities/DMSHelpers.py
@@ -309,6 +309,8 @@ class DMSHelpers(object):
     sites = [site for site in mapping if se in mapping[site]]
     if len(sites) > 1 and self.__opsHelper.getValue('DataManagement/ForceSingleSitePerSE', True):
       return S_ERROR('SE is at more than one site')
+    if len(sites) == 0:
+      return S_ERROR('SE does not belong to any site')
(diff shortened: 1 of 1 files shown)


chaen:
Ouh.. that's a very delicate change. You are changing the interface here. We also have a few SE that are associated to no sites. Currently, if you ask for the Site of an SE not associated, you get an empty list (kind of make sense). But now you would get an error, which is very different. And I am pretty sure there are places in the code (at least in the extensions) where one relies on the empty list (or `S_OK(None)`) 
But I am not sure I understand how your fix solves the crashing issue ?

iueda:
Well, this _getLocalSitesForSE would not be called directly, but via getLocalSitesForSE, which crashes with an empty list returned by _getLocalSitesForSE...

... ah, wait, I was checking the older release, and getLocalSitesForSE has now a protection since v6r19.
```
    if not sites['Value']:
      return S_OK(None)
```
Then, ok I wouldn't need this change in this pull-request.  Sorry for the noise.

chaen:
No worries, thanks !

PR #3814, opened 2018-08-28
[v6r20] Fix in the replica report

atsareg:
After changing the meaning of the plugin section name, the SE information was not properly
evaluated by the SEManager of the File Catalog. This PR fixes it. It also takes properly into account the possible definitions of VO specific paths when constructing PFNs on the fly in the getReplicas() client method, this fixes the PFNs reported by dls -L or similar commands.

BEGINRELEASENOTES

*DMS
FIX: SEManager - adapt to the new meaning of the SE plugin section name
FIX: SEManager - return also VO specific prefixes for the getReplicas() and similar calls
FIX: FileCatalogClient - take into account VO specific prefixes when constructing PFNs on the fly

ENDRELEASENOTES

### DIFF
--- a/Resources/Catalog/FileCatalogClient.py
+++ b/Resources/Catalog/FileCatalogClient.py
@@ -111,7 +111,7 @@ class FileCatalogClient(FileCatalogClientBase):
     result = rpcClient.getReplicas(lfns, allStatus)
     if not result['OK']:
       return result
-    vo = getVOfromProxyGroup().get( 'Value', None )
+    vo = getVOfromProxyGroup().get('Value', None)
 
     lfnDict = result['Value']
     seDict = result['Value'].get('SEPrefixes', {})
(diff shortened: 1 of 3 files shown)


chaen:
I did not yet do the review of the code per se, but what is this VOPath ? I could find no information whatsoever about it in the documentation (This PR could be the opportunity to add it in the documentation of the storage elements btw). But from what I understand, isn't it something that could be handled in a more graceful way with the SEBase ?

Resources/Catalog/FileCatalogClient.py
```diff
@@ -110,13 +111,18 @@ def getReplicas(self, lfns, allStatus=False, timeout=120):
...
 
     lfnDict = result['Value']
     seDict = result['Value'].get('SEPrefixes', {})
     for lfn in lfnDict['Successful']:
       for se in lfnDict['Successful'][lfn]:
-        if not lfnDict['Successful'][lfn][se] and se in seDict:
-          lfnDict['Successful'][lfn][se] = seDict[se] + lfn
+        if not lfnDict['Successful'][lfn][se]:
+          # For some VO's the prefix can be non-standard
+          if "VOPrefix" in seDict and se in seDict['VOPrefix'] and vo in seDict['VOPrefix'][se]:
```
chaen: I believe this is clearer and more efficient no ? 
```
voPrefix = seDict.get('VOPrefix', {}).get(se,{}_.get(vo)
sePrefix = seDict.get(se)
if voProfix:
   # blabla
elif sePrefix:
  # blibli  
```

atsareg:
Review OK ?

PR #3841, opened 2018-09-25
[v7] Dirac.runLocal: have to define DIRACROOT before expandvars is called

andresailer:
`systemCall` does not expand shell variables ($DIRACROOT), which was only defined after the call to `os.path.expandvars`. In the past `shellCall` was called with `$DIRACROOT/scripts/dirac-jobexec`, which worked because `shellCall` expands the variables.

* No release notes, because this is a fix to PR #3808 in the same release
* This currently breaks our nightly `integration` job tests, so a fast merge would be appreciated.


andresailer:
Fixed the formatting

PR #3855, opened 2018-10-10
[integration] script to collate ConfigTemplate files into the dirac.cfg

andresailer:
@chaen please have a look at this

* The parsing and reading of the existing dirac.cfg broke all the links with targets (`#targetName`) because the comment parsing of the CFG drops the `#`
* comments following values are not recognized as such, commas in those comments also cause problems
* keys existing inside a section are not removed if they become outdated (if instead of merging we replace the sections this could be remedied)
* the `##END` comment after the agent sections get shuffled around

Todo:

- [x] Instead of merging into the Systems section, replace the whole section with the contents of the ConfigTemplate.cfg's to ensure consistency. 
  - [x] Add the anchor back after the parse
- [x] Run the script when the documentation is created
- [x] Add a page for the file in the documentation
- [x] in the dirac.cfg file in the git repository just add a link pointing to the dirac.readthedocs.io page with the complete content

BEGINRELEASENOTES

*Docs
NEW: Added script (`docs/Tools/UpdateDiracCFG.py`) to collate the ConfigTemplate.cfg files into the main dirac.cfg file

ENDRELEASENOTES


andresailer:
@chaen Please review

chaen:
Great !

PR #3861, opened 2018-10-19
[v6r20] RMS: correct the test 

chaen:
This fix the issue @andresailer discovered in https://github.com/DIRACGrid/DIRAC/pull/3842 when importing RequestDB. The issue was that a `File` object cannot be reused. The test was incorrectly doing it... So I fixed the test, and added a warning in the developer doc.

BEGINRELEASENOTES

* RMS
FIX: tests do not re-use File objects

ENDRELEASENOTES

### DIFF
--- a/docs/source/DeveloperGuide/Systems/RequestManagement/index.rst
+++ b/docs/source/DeveloperGuide/Systems/RequestManagement/index.rst
@@ -112,6 +112,11 @@ Once the request is ready, you can insert it to the `ReqDB`::
   >>> rc = ReqClient() # # create client
   >>> rc.putRequest( request ) # # put request to ReqDB
 
+.. DANGER::
+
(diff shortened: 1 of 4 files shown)


docs/source/DeveloperGuide/Systems/RequestManagement/index.rst
```diff
@@ -112,6 +112,11 @@ Once the request is ready, you can insert it to the `ReqDB`::
   >>> rc = ReqClient() # # create client
   >>> rc.putRequest( request ) # # put request to ReqDB
 
+.. DANGER::
```
andresailer:
```suggestion
.. warning::
```
warning gives it a nice read box: e.g. https://dirac.readthedocs.io/en/latest/CodeDocumentation/Core/Utilities/Adler.html

I think this should also go to the Operation.addFile function and maybe the File module or constructor

PR #3876, opened 2018-10-30
[v6r20] Get back discoverInterfaces function

xianghuzhao:
The function "discoverInterfaces" is removed in early v6r20 version. But VMDIRAC still need it in Agent/VirtualMachineMonitorAgent.py
So here it is added again, together with function "getMACFromInterface" which is called by it.

BEGINRELEASENOTES

*Core
FIX: Add function "discoverInterfaces" again which is still needed for VMDIRAC

ENDRELEASENOTES

### DIFF
--- a/Core/Utilities/Network.py
+++ b/Core/Utilities/Network.py
@@ -7,7 +7,42 @@ __RCSID__ = "$Id$"
 import socket
 import urlparse
 import os
+import struct
+import array
+import fcntl
+import platform
+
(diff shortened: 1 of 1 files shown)


chaen:
It would be good to add a comment (well, documenting the functions first) saying that these are used in VMDIRAC. Otherwise, one day or another, they will again be gone during a cleanup.
Alternatively, wouldn't it make more sense to just move them in VMDIRAC ?

PR #3916, opened 2018-12-07
[v6r20] Add isSameSE method to StorageElement 

chaen:
Part of https://github.com/DIRACGrid/DIRAC/issues/3908
@phicharp can you have a look please ? 

BEGINRELEASENOTES

*Resources
NEW: Add isSameSE method to StorageElement which works for all protocols

*TransformationSystem
CHANGE: use SE.isSameSE

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Client/FTS3Operation.py
+++ b/DataManagementSystem/Client/FTS3Operation.py
@@ -1,4 +1,5 @@
 import datetime
+import errno
 import json
 from sqlalchemy import orm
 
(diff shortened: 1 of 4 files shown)


phicharp:
Looks fine with me... 
I only find it strange to have it as a method of StorageElement as it is a binary test. I would have put it maybe in DMSHelpers...

chaen:
Well I implemented there because I related it to the `==` operator. But I may change it.

phicharp:
Good point, yes indeed this is pretty much `==`

phicharp:
We can also put a binary function in `DMSHelpers` that calls that one ;-)

PR #3931, opened 2019-01-14
[v7r0] Modify AuthManager to work with unregistered DN

TaykYoku:
BEGINRELEASENOTES
Modified AuthManager to work with unregistered DN

*Core/DISET/AuthManager.py
FIX: Modified to work with unregistered DN in credDict

*Core/test/Test_AuthManager.py
CHANGE: Added new tests with empty credDict and unregistered DN in credDict

ENDRELEASENOTES


zmathe:
non registered DN should not be able to connect any service, if authentication is required. I think it is related to: https://github.com/DIRACGrid/WebAppDIRAC/pull/311
Probably, this modification will work with WebApp, but it may not work using service. 
I do not understand why you want to allow a DN to connect a DIRAC service, which is not registered in DIRAC. Non registered DN only can connect to the DIRAC service, if authenticated=all 
I let other to comment.

chaen:
Then the clean solution would be to use `dip` instead of `dips` no ?

atsareg:
We have a case when users with certificates in their browsers that is not registered in DIRAC try to access the DIRAC Web Portal. In this case they simply get Not authorized error. They can not access by http because we start to put http -> https redirection to satisfy security requirements. If they access with no certificate at all, they get the visitor access and can have a minimal interaction with the system. So, it is natural that the non-registered certificate case is treated in exactly the same way as no-certificate case. The authorization rules are all collected in the AuthManager for both DIRAC services and the Web Portal. So, it is natural that the non-registered certificate case is managed in the AuthManager as well. Accessing DIRAC services with a non-registered certificate does not make much sense and in most cases will end up in the non-authorized error. Except for methods with ANY access rule. The latter case anyway must be reserved for any access a la dip. So, I do not think that this PR introduces any potential security hole compared with what we have now.

chaen:
I see. Thanks for the detailed explanation. I did not look at the code, but from the logic point of view (no certificate == unknown certificate) it makes sense to me.

PR #3944, opened 2019-01-30
v6r21 -----  fix bashrc creation in case of diracos

zmathe:
BEGINRELEASENOTES

*CORE
FIX: set DIRACOS environment variable before souring diracosrc
ENDRELEASENOTES


zmathe:
replaced by #3945

PR #3956, opened 2019-02-12
[v6r21] change the bucket size

zmathe:
BEGINRELEASENOTES
*MonitoringSystem
CHANGE: Change the bucket size from week to day. 
ENDRELEASENOTES


zmathe:
If I plot all data, which ever entered to the database, I get very strange plot:
823d5d0714f3de090aaa250e474af7a4

It seam if the period was more than one year, one week was not correct. I changed 1w to 7days and it works (Yes, I know it is still a week)
e66b92de241f6f06e4f7954944d19d95

Issue #3955, opened 2019-02-11
[DMS] folder metadata


axlbonnet:
I can't get the metadata of a folder on the DFC (I'm interested in the modification date).
I've tried `dirac-dms-lfn-metadata` on an existing folder but I get this error message :
```
$ dirac-dms-lfn-metadata /biomed/user/c/creatis/diracTests
{'Failed': {'/biomed/user/c/creatis/diracTests': 'No such file or directory'},
 'Successful': {}}
```

The error suggests it should work on directories and it seems to be a bug, but maybe this is just a wrong message.
For information, `dirac-dms-lfn-metadata` works for a file in this folder :
```
$ dirac-dms-lfn-metadata /biomed/user/c/creatis/diracTests/testFile.txt
{'Failed': {},
 'Successful': {'/biomed/user/c/creatis/diracTests/testFile.txt': {'Checksum': '66bd07ba',
                                                                   'ChecksumType': 'Adler32',
                                                                   'CreationDate': datetime.datetime(2019, 1, 23, 13, 59, 54),
                                                                   'FileID': 118L,
                                                                   'GID': 1,
                                                                   'GUID': '09F31262-0A29-F419-8637-36F5577BA137',
                                                                   'Mode': 509,
                                                                   'ModificationDate': datetime.datetime(2019, 1, 23, 13, 59, 54),
                                                                   'Owner': 'creatis',
                                                                   'OwnerGroup': 'biomed_user',
                                                                   'Size': 23L,
                                                                   'Status': 'AprioriGood',
                                                                   'UID': 5}}}

```

chaen:
blind guess: try adding a slash at the end of your directory name.

axlbonnet:
Tried that, it doesn't change anything :
```
$ dirac-dms-lfn-metadata /biomed/user/c/creatis/diracTests/
{'Failed': {'/biomed/user/c/creatis/diracTests/': 'No such file or directory'},
 'Successful': {}}
```

atsareg:
Solved in v6r21 release

PR #3960, opened 2019-02-15
[v6r21] Dirac API returns directory LFN metadata

atsareg:
This is a PR to fix issue #3955 

BEGINRELEASENOTES

*Interfaces
CHANGE: Dirac.py - getLFNMetadata returns result for both file and directory LFNs

ENDRELEASENOTES

### DIFF
--- a/Interfaces/API/Dirac.py
+++ b/Interfaces/API/Dirac.py
@@ -1227,7 +1227,8 @@ class Dirac(API):
     return self.getLfnMetadata(lfns, printOutput=printOutput)
 
   def getLfnMetadata(self, lfns, printOutput=False):
-    """Obtain replica metadata from file catalogue client. Input LFN(s) can be string or list.
+    """Obtain replica metadata from file catalogue client.
+       Input LFN(s) can be string or list. LFN(s) can be either files or directories
 
        Example usage:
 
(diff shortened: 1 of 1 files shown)


Interfaces/API/Dirac.py
```diff
@@ -1249,19 +1250,32 @@ def getLfnMetadata(self, lfns, printOutput=False):
...
       self.log.warn('Failed to retrieve file metadata from the catalogue')
-      self.log.warn(repsResult['Message'])
-      return repsResult
+      self.log.warn(fileResult['Message'])
+      return fileResult
+
+    repsResult = fileResult['Value']
+    if repsResult['Failed']:
+      # Some entries can be directories
+      dirs = repsResult['Failed'].keys()
```
chaen: why do you need to do the copy here ?

Interfaces/API/Dirac.py
```diff
@@ -1249,19 +1250,32 @@ def getLfnMetadata(self, lfns, printOutput=False):
...
+      dirResult = fc.getDirectoryMetadata(dirs)
+      if not dirResult['OK']:
+        self.log.warn('Failed to retrieve directory metadata from the catalogue')
```
chaen: maybe put the message as variable part of the logging  ?

Interfaces/API/Dirac.py
```diff
@@ -1249,19 +1250,32 @@ def getLfnMetadata(self, lfns, printOutput=False):
...
+        self.log.warn(dirResult['Message'])
+        return dirResult
+      for dir in dirResult['Value']['Successful']:
```
chaen: `dir` is a reserved keyword

Interfaces/API/Dirac.py
atsareg: Just to be more readable

PR #4035, opened 2019-04-16
[WIP] Remove race condition from StompMQConnector

wkrzemien:
Add lock mecanism to StomMQConnector
Also, add isActiveCallback field to ReconnectListener and reset it
in disconnect method.

BEGINRELEASENOTES
*Resources/MessageQueue
FIX: With multiple brokers scenario the call of producer close() function can lead to race conditions. This fix adds the lock and also desactivates the reconnect procedure when the disconnect() function is called.
ENDRELEASENOTES

### DIFF
--- a/Resources/MessageQueue/StompMQConnector.py
+++ b/Resources/MessageQueue/StompMQConnector.py
@@ -10,6 +10,7 @@ import ssl
 import time
 import stomp
 
+from DIRAC.Core.Utilities.LockRing import LockRing
 from DIRAC.Resources.MessageQueue.MQConnector import MQConnector
 from DIRAC.Core.Security import Locations
 from DIRAC import S_OK, S_ERROR, gLogger
@@ -38,9 +39,18 @@ class StompMQConnector(MQConnector):
     super(StompMQConnector, self).__init__()
     self.log = gLogger.getSubLogger(self.__class__.__name__)
     self.connections = {}
+    self.__lock = None
 
     if 'DIRAC_DEBUG_STOMP' in os.environ:
       gLogger.enableLogsFromExternalLibs()
+
+  @property
+  def lock(self):
+    """ Lock to assure thread-safe calls of certain methods.
+    """
+    if not self.__lock:
+      self.__lock = LockRing().getLock(self.__class__.__name__, recursive=True)
+    return self.__lock
 
   def setupConnection(self, parameters=None):
     """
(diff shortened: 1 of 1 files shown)


wkrzemien:
@chaen @fstagni please have a look

chaen:
hi @wkrzemien 
I had a quick look at the problem too, and came to the same conclusion as you: the example given on the website cannot work.
First of all, we should target the v6r21 branch, not integration. This is a critical problem that must be fixed urgently and properly, so it has to go in a patch release. 

Second, how did you check that it was a race condition ? I am still not convinced it has to do with multiple brokers.

I attempted a different approach by modifying the `disconnect` method of the StompMQConnector. Basically, before actively call the `disconnect` of the connection object, I remove the `ReconnectListener`.

This is how it was before
https://github.com/DIRACGrid/DIRAC/blob/2cbd221beb4c3613bc7b080af54344375f265759/Resources/MessageQueue/StompMQConnector.py#L183-L192

This is now

```
  def disconnect(self, parameters=None):
    """
    Disconnects from the message queue server
    """
    fail = False
    for connection in self.connections.itervalues():
      try:
        # THESE ARE THE NEW LINES
        if connection.get_listener('ReconnectListener'):
          connection.remove_listener('ReconnectListener')
        ####################
        connection.disconnect()
      except Exception as e:
        self.log.error('Failed to disconnect: %s' % e)
```

By adding this, the test example (which btw should become a standard test if not already there) I gave you works fine.
What do you think of that approach ? 
Do you want that we ask for a second broker on our test instance ?

fstagni:
BTW: this is targeting integration, we need this to target rel-v6r21

chaen:
FYI, I am running what I propose as a hot fix on one machine, and so far so good

wkrzemien:
@chaen Your approach is probably not ok. We have one Listener per connection and the same connection can be reused by N producers in principle. By removing the Listener you remove it from all producers that use the same connection. It is solving the problem, but then in case of the server down, other producer will not try to reconnect. The reconnect operation should be per connection in this scenario.
IAs I told you, I have tests with clos()e called and they pass withoout any problem. So what I did I added the second broker and then the error appears.

wkrzemien:
@fstagni Ok I changed the branch.

chaen:
From what I understood from the code (and the comment btw) bellow is that the disconnect is called only when there are no more producers

https://github.com/DIRACGrid/DIRAC/blob/2cbd221beb4c3613bc7b080af54344375f265759/Resources/MessageQueue/MQConnectionManager.py#L141-L173

but I may have misunderstood ?

wkrzemien:
You are right!
That simplifies everything a lot.
So you approach is much better.

wkrzemien:
Let me check it with my tests.

fstagni:
BTW travis fails because it wrongly uses as a base the `integration` branch.

wkrzemien:
Should I close it and reopen again?

fstagni:
yes, better

wkrzemien:
It will be reopen soon :-)

chaen:
@wkrzemien don't forget that this fix must be in the next patch release so do not wait for too long before opening the new one ;-)

wkrzemien:
Ok. Probably today in the afternoon.

fstagni:
@wkrzemien just bumping this... in case you have the time for making this PR.

wkrzemien:
@fstagni PR done. I wonder where to add tests.

wkrzemien:
@fstagni chris answered me already

PR #4048, opened 2019-04-24
[v6r21] Better third party URL generation and matrix fix

chaen:
BEGINRELEASENOTES
*Resources
CHANGE: SE: give preference to native plugins when generating third party URLS

*DMS
FIX: correct the header of the CSV file generated by dirac-dms-protocol-matrix 
ENDRELEASENOTES


chaen:
rebased

PR #4060, opened 2019-05-02
[v6r21] Fix version of Ubuntu on Travis

petricm:
Travis can run python tests on different versions of Ubuntu: xenial, trusty or precise. At present it switches between them, based on load and this results in slightly different testing behavior. This PR fixes the test environment to trusty 14.04 LTS, we can run only on this version due to fts3/pycurl requirements.

Please merge ASAP to v6r22 and integration.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # From repo
-git+https://:@gitlab.cern.ch:8443/fts/fts-rest.git#egg=fts3
+fts3-rest
 git+https://github.com/DIRACGrid/pyGSI.git#egg=GSI
 
 # From pypi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 python:
   - "2.7"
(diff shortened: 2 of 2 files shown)


chaen:
thanks ! Also v6r21 no ?

petricm:
Yes, the PR is against 21. One should also merge it to 20 if you expect PRs/tests to be run there.

requirements.txt
```diff
@@ -21,6 +21,7 @@ ipython==5.3.0
 numpy>=1.10.1
 pexpect>=4.0.1
 psutil>=4.2.0
+pycurl==7.19.0
```
andresailer: Shouldn't this just be dropped if it anyway comes from fts3?

petricm: I wasn't aware till just now, but on xenial the fts3 requirement of pycurl does not compile and automatically pics a higher version... now I reverted to trusty

petricm: also on trusty you get
```
ERROR: fts3-rest 3.8.4 has requirement pycurl==7.19.0, but you'll have pycurl 7.43.0.2 which is incompatible.
```

petricm: Completely removing the line solves it

PR #4066, opened 2019-05-08
[WIP] [v6r21] Fix for transformation-replication script

andresailer:
BEGINRELEASENOTES

*TS
FIX: The dirac-transformation-replication script will now create valid transformations given only the required arguments. Instead of the 'Broadcast' plugin, the 'Standard' plugin is created if not SourceSE is given. If a value for the plugin argument is given, that will be used.

ENDRELEASENOTES


andresailer:
Of course the 'Standard' plugin is not useful for the ReplicateAndRegister, because it might not do anything (thanks to @chaen for testing the tutorial)...

PR #4069, opened 2019-05-10
[v6r21] update DIRAC installation procedure

zmathe:
BEGINRELEASENOTES

*docs
CHANGE: DIRAC installation procedure is updated taking account DIRACOS
ENDRELEASENOTES


chaen:
Do you eally want to keep the hardcoded values like your DN and such ?

PR #4076, opened 2019-05-22
[v6r21] make sure the module only installed once

zmathe:
BEGINRELEASENOTES

*Core
FIX: A certain module like WebAppDIRAC must be checked out from the code repository once.

ENDRELEASENOTES


Core/scripts/dirac-install.py
```diff
@@ -2505,7 +2505,7 @@ def createSymbolicLink():
       logNOTICE("Writing down the releases files")
       releaseConfig.dumpReleasesToPath()
     logNOTICE("Installing modules...")
-    for modName in modsOrder:
+    for modName in set(modsOrder):
```
andresailer: From the name of the variable, I would assume that the order of entries matters, but a `set` does not preserve the order.

PR #4103, opened 2019-06-03
[integration] Implement Multi-VO File Catalog Metadata Handling

martynia:
BEGINRELEASENOTES
The idea of the update is to allow to store user metadata in a VO dependant way. For indexed
metadata a SQL table name, which normally holds a meta name as a suffix, is now extended to include w VO suffix. This way each VO writes to their own SQL table. 
The [DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py](url) and 
[DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py](url) now contain a "hook" which, when overwritten in a derived class generates a required VO-specific tbale name suffix.
ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
@@ -9,12 +9,14 @@ import six
 import os
 from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Utilities.Time import queryTime
+from DIRAC.DataManagementSystem.DB.FileCatalogComponents.MetaNameMixIn import MetaNameMixIn
 
 
-class DirectoryMetadata:
+class DirectoryMetadata(MetaNameMixIn):
 
   def __init__(self, database=None):
 
+    super(DirectoryMetadata, self).__init__()
     self.db = database
 
   def setDatabase(self, database):
@@ -26,24 +28,33 @@ class DirectoryMetadata:
 #
 
   def addMetadataField(self, pname, ptype, credDict):
-    """ Add a new metadata parameter to the Metadata Database.
-        pname - parameter name, ptype - parameter type in the MySQL notation
+    """
+    Add a new metadata parameter to the Metadata Database.
+    Modified to use fully qualified metadata names.
+
+    :param str pname: parameter name
+    :param str ptype: parameter type in the MySQL notation
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
     """
 
+    # existing pnames are fully qualified, so
+    fqPname = self.getMetaName(pname, credDict)
     result = self.db.fmeta.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
+    if fqPname in result['Value']:
       return S_ERROR('The metadata %s is already defined for Files' % pname)
 
     result = self.getMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      if ptype.lower() == result['Value'][pname].lower():
+
+    if fqPname in result['Value']:
+      if ptype.lower() == result['Value'][fqPname].lower():
         return S_OK('Already exists')
       return S_ERROR('Attempt to add an existing metadata with different type: %s/%s' %
-                     (ptype, result['Value'][pname]))
+                     (ptype, result['Value'][fqPname]))
 
     valueType = ptype
     if ptype.lower()[:3] == 'int':
@@ -58,26 +69,27 @@ class DirectoryMetadata:
       valueType = "VARCHAR(64)"
 
     req = "CREATE TABLE FC_Meta_%s ( DirID INTEGER NOT NULL, Value %s, PRIMARY KEY (DirID), INDEX (Value) )" \
-        % (pname, valueType)
+        % (fqPname, valueType)
     result = self.db._query(req)
     if not result['OK']:
       return result
 
-    result = self.db.insertFields('FC_MetaFields', ['MetaName', 'MetaType'], [pname, ptype])
+    result = self.db.insertFields('FC_MetaFields', ['MetaName', 'MetaType'], [fqPname, ptype])
     if not result['OK']:
       return result
 
     metadataID = result['lastRowId']
-    result = self.__transformMetaParameterToData(pname)
+    result = self.__transformMetaParameterToData(fqPname)
     if not result['OK']:
       return result
 
     return S_OK("Added new metadata: %d" % metadataID)
 
-  def deleteMetadataField(self, pname, credDict):
-    """ Remove metadata field
+  def deleteMetadataField(self, rawPname, credDict):
+    """ Remove metadata field.
+        Table name is now fully qualified
     """
-
+    pname = self.getMetaName(rawPname, credDict)
     req = "DROP TABLE FC_Meta_%s" % pname
     result = self.db._update(req)
     error = ''
@@ -90,7 +102,7 @@ class DirectoryMetadata:
         result["Message"] = error + "; " + result["Message"]
     return result
 
-  def getMetadataFields(self, credDict):
+  def getMetadataFields(self, credDict, stripVO=False):
     """ Get all the defined metadata fields
     """
 
@@ -102,6 +114,9 @@ class DirectoryMetadata:
     metaDict = {}
     for row in result['Value']:
       metaDict[row[0]] = row[1]
+    # strip the VO suffix, if required
+    if stripVO:
+      metaDict = self.stripSuffix(metaDict, credDict)
 
     return S_OK(metaDict)
 
@@ -189,19 +204,20 @@ class DirectoryMetadata:
     if not dirmeta['OK']:
       return dirmeta
 
-    for metaName, metaValue in metadict.iteritems():
-      if metaName not in metaFields:
+    for metaName, metaValue in metadict.items():
+      fqMetaName = self.getMetaName(metaName, credDict)
+      if fqMetaName not in metaFields:
         result = self.setMetaParameter(dpath, metaName, metaValue, credDict)
         if not result['OK']:
           return result
         continue
       # Check that the metadata is not defined for the parent directories
-      if metaName in dirmeta['Value']:
+      if fqMetaName in dirmeta['Value']:
         return S_ERROR('Metadata conflict detected for %s for directory %s' % (metaName, dpath))
-      result = self.db.insertFields('FC_Meta_%s' % metaName, ['DirID', 'Value'], [dirID, metaValue])
+      result = self.db.insertFields('FC_Meta_%s' % fqMetaName, ['DirID', 'Value'], [dirID, metaValue])
       if not result['OK']:
         if result['Message'].find('Duplicate') != -1:
-          req = "UPDATE FC_Meta_%s SET Value='%s' WHERE DirID=%d" % (metaName, metaValue, dirID)
+          req = "UPDATE FC_Meta_%s SET Value='%s' WHERE DirID=%d" % (fqMetaName, metaValue, dirID)
           result = self.db._update(req)
           if not result['OK']:
             return result
@@ -211,8 +227,15 @@ class DirectoryMetadata:
     return S_OK()
 
   def removeMetadata(self, dpath, metadata, credDict):
-    """ Remove the specified metadata for the given directory
     """
+    Remove the specified metadata for the given directory for users own VO.
+
+    :param str dpath: directory path
+    :param dict metadata: metadata dictionary
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
+    """
+
     result = self.getMetadataFields(credDict)
     if not result['OK']:
       return result
@@ -227,6 +250,8 @@ class DirectoryMetadata:
 
     failedMeta = {}
     for meta in metadata:
+      # get fully qualified metadata name
+      meta = self.getMetaName(meta, credDict)
       if meta in metaFields:
         # Indexed meta case
         req = "DELETE FROM FC_Meta_%s WHERE DirID=%d" % (meta, dirID)
@@ -248,9 +273,17 @@ class DirectoryMetadata:
       return S_OK()
 
   def setMetaParameter(self, dpath, metaName, metaValue, credDict):
-    """ Set an meta parameter - metadata which is not used in the the data
-        search operations
     """
+    Set an meta parameter - metadata which is not used in the the data
+    search operations.
+
+    :param str dpath: directory path
+    :param str metaName: metadata name
+    :param str metaValue: metadata value
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
+    """
+
     result = self.db.dtree.findDir(dpath)
     if not result['OK']:
       return result
@@ -260,7 +293,7 @@ class DirectoryMetadata:
 
     result = self.db.insertFields('FC_DirMeta',
                                   ['DirID', 'MetaKey', 'MetaValue'],
-                                  [dirID, metaName, str(metaValue)])
+                                  [dirID, self.getMetaName(metaName, credDict), str(metaValue)])
     return result
 
   def getDirectoryMetaParameters(self, dpath, credDict, inherited=True, owndata=True):
@@ -303,10 +336,19 @@ class DirectoryMetadata:
 
     return S_OK(metaDict)
 
-  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
-    """ Get metadata for the given directory aggregating metadata for the directory itself
-        and for all the parent directories if inherited flag is True. Get also the non-indexed
-        metadata parameters.
+  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True, stripVO=False):
+    """
+    Get metadata for the given directory aggregating metadata for the directory itself
+    and for all the parent directories if inherited flag is True. Get also the non-indexed
+    metadata parameters. If the method is used in a call chain which supplies data back to
+    the client stripVO should  be set to True, otherwise the default should be used.
+
+    :param str path: directory path
+    :param dict credDict: client credential dictionary
+    :param bool inherited: iclude parent directories if True
+    :param bool owndata:
+    :param bool stripVO: if set to True, the VO suffix is stripped.
+    :return: standard Dirac result object
     """
 
     result = self.db.dtree.getPathIDs(path)
@@ -350,6 +392,11 @@ class DirectoryMetadata:
       metaDict.update(result['Value'])
       for meta in result['Value']:
         metaOwnerDict[meta] = 'OwnParameter'
+
+    if stripVO:
+      metaDict = self.stripSuffix(metaDict, credDict)
+      metaOwnerDict = self.stripSuffix(metaOwnerDict, credDict)
+      metaTypeDict = self.stripSuffix(metaTypeDict, credDict)
 
     result = S_OK(metaDict)
     result['MetadataOwner'] = metaOwnerDict
@@ -513,6 +560,8 @@ class DirectoryMetadata:
     if not result['OK']:
       return result
     metaTypeDict = result['Value']
+    metaTypeDict = self.stripSuffix(metaTypeDict, credDict)
+
     resultDict = {}
     extraDict = {}
     for key, value in metaDict.iteritems():
@@ -586,14 +635,14 @@ class DirectoryMetadata:
     # Now check the meta data for the requested directory and its parents
     finalMetaDict = dict(metaDict)
     for meta in metaDict.keys():
-      result = self.__checkDirsForMetadata(meta, metaDict[meta], pathString)
+      fqmeta = self.getMetaName(meta, credDict)
+      result = self.__checkDirsForMetadata(fqmeta, metaDict[meta], pathString)
       if not result['OK']:
         return result
       elif result['Value'] is not None:
         # Some directory in the parent hierarchy is already conforming with the
         # given metadata, no need to check it further
         del finalMetaDict[meta]
-
     if finalMetaDict:
       pathSelection = ''
       if pathDirID:
@@ -603,11 +652,12 @@ class DirectoryMetadata:
         pathSelection = result['Value']
       dirList = []
       first = True
-      for meta, value in finalMetaDict.iteritems():
+      for meta, value in finalMetaDict.items():
+        fqmeta = self.getMetaName(meta, credDict)
         if value == "Missing":
-          result = self.__findSubdirMissingMeta(meta, pathSelection)
+          result = self.__findSubdirMissingMeta(fqmeta, pathSelection)
         else:
-          result = self.__findSubdirByMeta(meta, value, pathSelection)
+          result = self.__findSubdirByMeta(fqmeta, value, pathSelection)
         if not result['OK']:
           return result
         mList = result['Value']
--- a/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
@@ -11,25 +11,31 @@ from DIRAC.Core.Utilities.List import intListToString
 from DIRAC.DataManagementSystem.Client.MetaQuery import FILE_STANDARD_METAKEYS, \
     FILES_TABLE_METAKEYS, \
     FILEINFO_TABLE_METAKEYS
+from DIRAC.DataManagementSystem.DB.FileCatalogComponents.MetaNameMixIn import MetaNameMixIn
 
 
-class FileMetadata:
-
+class FileMetadata(MetaNameMixIn):
   def __init__(self, database=None):
 
+    super(FileMetadata, self).__init__()
     self.db = database
 
   def setDatabase(self, database):
     self.db = database
 
-##############################################################################
-#
-#  Manage Metadata fields
-#
-##############################################################################
+  ##############################################################################
+  #
+  #  Manage Metadata fields
+  #
+  ##############################################################################
   def addMetadataField(self, pname, ptype, credDict):
-    """ Add a new metadata parameter to the Metadata Database.
-        pname - parameter name, ptype - parameter type in the MySQL notation
+    """
+    Add a new metadata parameter to the Metadata Database.
+
+    :param str pname: parameter name
+    :param str ptype: parameter type in the MySQL notation
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
     """
 
     if pname in FILE_STANDARD_METAKEYS:
@@ -38,57 +44,70 @@ class FileMetadata:
     result = self.db.dmeta.getMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      return S_ERROR('The metadata %s is already defined for Directories' % pname)
+    # existing pnames are fully qualified, so
+    fqPname = self.getMetaName(pname, credDict)
+    if fqPname in result['Value']:
+      return S_ERROR('The metadata %s is already defined for Directories' % fqPname)
 
     result = self.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      if ptype.lower() == result['Value'][pname].lower():
+    if fqPname in result['Value']:
+      if ptype.lower() == result['Value'][fqPname].lower():
         return S_OK('Already exists')
       else:
         return S_ERROR('Attempt to add an existing metadata with different type: %s/%s' %
-                       (ptype, result['Value'][pname]))
-
+                       (ptype, result['Value'][fqPname]))
     valueType = ptype
     if ptype == "MetaSet":
       valueType = "VARCHAR(64)"
     req = "CREATE TABLE FC_FileMeta_%s ( FileID INTEGER NOT NULL, Value %s, PRIMARY KEY (FileID), INDEX (Value) )" \
-        % (pname, valueType)
+          % (fqPname, valueType)
     result = self.db._query(req)
     if not result['OK']:
       return result
 
-    result = self.db.insertFields('FC_FileMetaFields', ['MetaName', 'MetaType'], [pname, ptype])
+    result = self.db.insertFields('FC_FileMetaFields', ['MetaName', 'MetaType'], [fqPname, ptype])
     if not result['OK']:
       return result
 
     metadataID = result['lastRowId']
-    result = self.__transformMetaParameterToData(pname)
+    result = self.__transformMetaParameterToData(fqPname)
     if not result['OK']:
       return result
 
     return S_OK("Added new metadata: %d" % metadataID)
 
   def deleteMetadataField(self, pname, credDict):
-    """ Remove metadata field
+    """
+    Remove metadata field (only from user's own VO)
+
+    :param str pname: parameter name
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
     """
 
-    req = "DROP TABLE FC_FileMeta_%s" % pname
+    fqPname = self.getMetaName(pname, credDict)
+    req = "DROP TABLE FC_FileMeta_%s" % fqPname
     result = self.db._update(req)
     error = ''
     if not result['OK']:
       error = result["Message"]
-    req = "DELETE FROM FC_FileMetaFields WHERE MetaName='%s'" % pname
+    req = "DELETE FROM FC_FileMetaFields WHERE MetaName='%s'" % fqPname
     result = self.db._update(req)
     if not result['OK']:
       if error:
         result["Message"] = error + "; " + result["Message"]
     return result
 
-  def getFileMetadataFields(self, credDict):
-    """ Get all the defined metadata fields
+  def getFileMetadataFields(self, credDict, stripVO=False):
+    """
+    Get all the defined metadata fields.
+
+    :param dict credDict: client credential dictionary
+    :param bool stripVO: flag whether to return metadata name with
+     or without a VO suffix
+    :return: standard Dirac result object
     """
 
     req = "SELECT MetaName,MetaType FROM FC_FileMetaFields"
@@ -100,17 +119,28 @@ class FileMetadata:
     for row in result['Value']:
       metaDict[row[0]] = row[1]
 
+    # strip the suffix, if required (for clients)
+    if stripVO:
+      metaDict = self.stripSuffix(metaDict, credDict)
     return S_OK(metaDict)
 
-###########################################################
-#
-# Set and get metadata for files
-#
-###########################################################
+  ###########################################################
+  #
+  # Set and get metadata for files
+  #
+  ###########################################################
 
   def setMetadata(self, path, metadict, credDict):
-    """ Set the value of a given metadata field for the the given directory path
     """
+    Set the value of a given metadata field for the the given directory path.
+    Modified to use fully qualified metadata name.
+
+    :param str path: directory path
+    :param dict metadict: metadata dictionary {name:value}
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
+    """
+
     result = self.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
@@ -124,16 +154,15 @@ class FileMetadata:
     else:
       return S_ERROR('File %s not found' % path)
 
-    for metaName, metaValue in metadict.iteritems():
-      if metaName not in metaFields:
+    for metaName, metaValue in metadict.items():
+      fqMetaName = self.getMetaName(metaName, credDict)
+      if fqMetaName not in metaFields:
         result = self.__setFileMetaParameter(fileID, metaName, metaValue, credDict)
       else:
-        result = self.db.insertFields('FC_FileMeta_%s' % metaName,
-                                      ['FileID', 'Value'],
-                                      [fileID, metaValue])
+        result = self.db.insertFields('FC_FileMeta_%s' % fqMetaName, ['FileID', 'Value'], [fileID, metaValue])
         if not result['OK']:
           if result['Message'].find('Duplicate') != -1:
-            req = "UPDATE FC_FileMeta_%s SET Value='%s' WHERE FileID=%d" % (metaName, metaValue, fileID)
+            req = "UPDATE FC_FileMeta_%s SET Value='%s' WHERE FileID=%d" % (fqMetaName, metaValue, fileID)
             result = self.db._update(req)
             if not result['OK']:
               return result
@@ -143,8 +172,16 @@ class FileMetadata:
     return S_OK()
 
   def removeMetadata(self, path, metadata, credDict):
-    """ Remove the specified metadata for the given file
     """
+    Remove the specified metadata for the given file (for user's own VO only)
+
+    :param str path: file path
+    :param dict metadata: metadata dictionary
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
+    """
+
+    # this would be fully qualified already
     result = self.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
@@ -160,6 +197,8 @@ class FileMetadata:
 
     failedMeta = {}
     for meta in metadata:
+      # get fully qualified metadata name
+      meta = self.getMetaName(meta, credDict)
       if meta in metaFields:
         # Indexed meta case
         req = "DELETE FROM FC_FileMeta_%s WHERE FileID=%d" % (meta, fileID)
@@ -192,12 +231,20 @@ class FileMetadata:
     return S_OK(fileID)
 
   def __setFileMetaParameter(self, fileID, metaName, metaValue, credDict):
-    """ Set an meta parameter - metadata which is not used in the the data
-        search operations
     """
+    Set an meta parameter - metadata which is not used in the the data
+    search operations. metaName is VO aware.
+
+    :param int fileID: file ID
+    :param str metaName: metadata name
+    :param metaValue: metadata value
+    :param dict credDict: client credential dictionary
+    :return: standard Dirac result object
+    """
+
     result = self.db.insertFields('FC_FileMeta',
                                   ['FileID', 'MetaKey', 'MetaValue'],
-                                  [fileID, metaName, str(metaValue)])
+                                  [fileID, self.getMetaName(metaName, credDict), str(metaValue)])
     return result
 
   def setFileMetaParameter(self, path, metaName, metaValue, credDict):
@@ -238,9 +285,16 @@ class FileMetadata:
 
     return S_OK(metaDict)
 
-  def getFileUserMetadata(self, path, credDict):
-    """ Get metadata for the given file
+  def getFileUserMetadata(self, path, credDict, stripVO=False):
     """
+    Get metadata for the given file.
+
+    :param str path:  file path
+    :param dict credDict: client credential dictionary
+    :param bool stripVO: If True, the VO suffix is stripped from metadata name
+    :return: standard Dirac result object
+    """
+
     # First file metadata
     result = self.getFileMetadataFields(credDict)
     if not result['OK']:
@@ -268,6 +322,9 @@ class FileMetadata:
       metaDict.update(result['Value'])
       for meta in result['Value']:
         metaTypeDict[meta] = 'NonSearchable'
+
+    if stripVO:
+      metaDict = self.stripSuffix(metaDict, credDict)
 
     result = S_OK(metaDict)
     result['MetadataType'] = metaTypeDict
@@ -329,11 +386,11 @@ class FileMetadata:
     result = self.db._update(req)
     return result
 
-#########################################################################
-#
-#  Finding files by metadata
-#
-#########################################################################
+  #########################################################################
+  #
+  #  Finding files by metadata
+  #
+  #########################################################################
 
   def __createMetaSelection(self, value):
     ''' Create selection string to be used in the SQL query
@@ -512,7 +569,7 @@ class FileMetadata:
       elif meta in FILE_STANDARD_METAKEYS:
         standardMetaDict[meta] = value
       else:
-        userMetaDict[meta] = value
+        userMetaDict[self.getMetaName(meta, credDict)] = value
 
     tablesAndConditions = []
     leftJoinTables = []
@@ -571,7 +628,7 @@ class FileMetadata:
     if not result['Value']:
       return S_OK([])
 
-#     fileList = [ row[0] for row in result['Value' ] ]
+    #     fileList = [ row[0] for row in result['Value' ] ]
     fileList = []
     for row in result['Value']:
       fileID = row[0]
@@ -600,13 +657,13 @@ class FileMetadata:
     dirFlag = result['Selection']
 
     # 2.- Get known file metadata fields
-#     fileMetaDict = {}
-    result = self.getFileMetadataFields(credDict)
+    #     fileMetaDict = {}
+    result = self.getFileMetadataFields(credDict)  # this returns fully qualified meta name
     if not result['OK']:
       return result
     fileMetaKeys = result['Value'].keys() + FILE_STANDARD_METAKEYS.keys()
-    fileMetaDict = dict(item for item in metaDict.iteritems() if item[0] in fileMetaKeys)
-
+    fileMetaDict = dict(item for item in metaDict.items()
+                        if self.getMetaName(item[0], credDict) in fileMetaKeys)
     fileList = []
     idLfnDict = {}
 
--- a/DataManagementSystem/DB/FileCatalogComponents/MetaNameMixIn.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/MetaNameMixIn.py
@@ -1,0 +1,41 @@
+""" DIRAC base MixIn class to manage file metadata and directory names.
+"""
+from __future__ import division
+
+
+class MetaNameMixIn(object):
+
+  def getMetaName(self, meta, credDict):
+    """
+    Return a metadata name based on client supplied meta name and client credentials
+    For the base class it just returns the name passed in.
+    This method is a pass-through and is meant to be overwritten by derived classes.
+
+    :param meta:  meta name
+    :param credDict: client credentials
+    :return: meta name
+    """
+
+    return meta
+
+  def getMetaNameSuffix(self, credDict):
+    """
+    Get meta name suffix based on client credentials. The method is needed to be able
+    to return metadata w/o a suffix to the client.
+    This method is a pass-through and is meant to be overwritten by derived classes.
+
+    :param credDict: client credentials
+    :return: the suffix. And empty string for a base class.
+    """
+
+    return ''
+
+  def stripSuffix(self, metaDict, credDict):
+    """
+    Strip suffix pass through, just return the metadata dictionary.
+
+    :param metaDict: meta dictionary to modify.
+    :param credDict: credential dictionary.
+    :return: unchanged metaDict
+    """
+    return metaDict
--- a/DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
@@ -33,7 +33,7 @@ class MultiVOMetaNameMixIn(MetaNameMixIn):
     :return: VO specific suffix
     """
     vo = Registry.getGroupOption(credDict['group'], 'VO')
-    return '_' + vo.replace('-', '_')
+    return '_' + vo.replace('-', '_').replace('.','')
 
   def stripSuffix(self, metaDict, credDict):
     """
--- a/DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
@@ -1,0 +1,50 @@
+""" DIRAC Multi VO MixIn class to manage file metadata and directory for multiple VO.
+"""
+from __future__ import division
+
+__RCSID__ = "$Id$"
+
+from DIRAC.DataManagementSystem.DB.FileCatalogComponents.MetaNameMixIn import MetaNameMixIn
+from DIRAC.ConfigurationSystem.Client.Helpers import Registry
+
+
+class MultiVOMetaNameMixIn(MetaNameMixIn):
+  """
+  MULti-VO MetaName MixIn implementation.
+  """
+
+  def getMetaName(self, meta, credDict):
+    """
+    Return a fully-qualified metadata name based on client-suplied metadata name and
+    client credentials. User VO is added to the metadata passed in.
+
+    :param meta: metadata name
+    :param credDict: client credentials
+    :return: fully-qualified metadata name
+    """
+
+    return meta + self.getMetaNameSuffix(credDict)
+
+  def getMetaNameSuffix(self, credDict):
+    """
+    Get a VO specific suffix from user credentials.
+
+    :param credDict: user credentials
+    :return: VO specific suffix
+    """
+    vo = Registry.getGroupOption(credDict['group'], 'VO')
+    return '_' + vo.replace('-', '_')
+
+  def stripSuffix(self, metaDict, credDict):
+    """
+    Strip the suffix from all keys which contain it, removing all other keys.
+
+    :param metaDict: original dict
+    :param credDict: user credential dictionary
+    :return: a new dict with modified keys
+    """
+
+    suffix = self.getMetaNameSuffix(credDict)
+    smetaDict = {key.rsplit(suffix, 1)[0]: value for key, value in metaDict.iteritems()
+                 if key.endswith(suffix)}
+    return smetaDict
--- a/DataManagementSystem/Service/FileCatalogHandler.py
+++ b/DataManagementSystem/Service/FileCatalogHandler.py
@@ -488,12 +488,13 @@ class FileCatalogHandler(RequestHandler):
   types_getMetadataFields = []
 
   def export_getMetadataFields(self):
-    """ Get all the metadata fields
+    """ Get all the metadata fields.
     """
-    resultDir = gFileCatalogDB.dmeta.getMetadataFields(self.getRemoteCredentials())
+    resultDir = gFileCatalogDB.dmeta.getMetadataFields(self.getRemoteCredentials(), stripVO=True)
     if not resultDir['OK']:
       return resultDir
-    resultFile = gFileCatalogDB.fmeta.getFileMetadataFields(self.getRemoteCredentials())
+
+    resultFile = gFileCatalogDB.fmeta.getFileMetadataFields(self.getRemoteCredentials(), stripVO=True)
     if not resultFile['OK']:
       return resultFile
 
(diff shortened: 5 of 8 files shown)


DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -8,6 +8,7 @@
 import os
 from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Utilities.Time import queryTime
+import Utilities
```
fstagni: Use full imports. Here, and later.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -24,25 +25,52 @@ def setDatabase(self, database):
...
+    Return a metadata name based on client supplied meta name and client credentials
+    For the base class it just returns the name passed in.
+    This method is meant to be overwritten by derived classes.
+
+    :param meta:  meta name
+    :param credDict: client credentials
+    :return: meta name
+    """
+
+    return meta
```
fstagni: What about returning `S_ERROR("method to be overwritten")` ?

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -24,25 +25,52 @@ def setDatabase(self, database):
...
+        Modified to use fully qualified metadata names.
     """
-
+    # existing pnames are fully qualified, so
+    fqPname = self.getMetaName(pname, credDict)
     result = self.db.fmeta.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
+    if fqPname in result['Value'].keys():
```
fstagni: Please avoid the `.keys()` methods (not python 3 friendly)

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -585,14 +620,17 @@ def findDirIDsByMetadata(self, queryDict, path, credDict):
...
+      result = self.__checkDirsForMetadata(fqmeta, metaDict[meta], pathString)
+      print "__checkDirsForMetadata result", result
       if not result['OK']:
         return result
       elif result['Value'] is not None:
         # Some directory in the parent hierarchy is already conforming with the
         # given metadata, no need to check it further
         del finalMetaDict[meta]
-
+    print "finalMetaDict ?",finalMetaDict
```
fstagni: No `print`s please.

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -194,13 +205,47 @@ def __getFileID( self, path ):
...
     """
     result = self.db._insert( 'FC_FileMeta',
                           ['FileID', 'MetaKey', 'MetaValue'],
-                          [fileID, metaName, str( metaValue )] )
+                          [fileID, self.getMetaName(metaName, credDict), str( metaValue )] )
     return result
 
+  @staticmethod
+  def fqMetaName(metaName, credDict):
+    return metaName+'_'+credDict['group']
```
fstagni: For the correct format please see pep8 rules (that we adopt with the exclusion of the spacing). Also travis complains (https://travis-ci.org/DIRACGrid/DIRAC/jobs/540768734)

DataManagementSystem/DB/FileCatalogComponents/MultiVODirectoryMetadata.py
```diff
@@ -0,0 +1,39 @@
+""" DIRAC Multi VO FileCatalog plugin class to manage file metadata for multiple VO.
+"""
+
+__RCSID__ = "$Id$"
+
+from DirectoryMetadata import DirectoryMetadata
```
fstagni: No relative imports.

Interfaces/API/Dirac.py
```diff
@@ -1262,6 +1282,119 @@ def getLfnMetadata(self, lfns, printOutput=False):
 
     return repsResult
 
+  #### User Metadata API ######################################################
```
fstagni: What are the advantages of adding these methods?

tests/Integration/Resources/Catalog/Test_UserMetadata.py
```diff
@@ -0,0 +1,127 @@
+"""
+Test of multi-VO user metadata handling. Assumes a running Dirac instance with the (master?) FileCatalog
```
fstagni: Can you run this in Jenkins?

tests/Integration/Resources/Catalog/Test_UserMetadata.py
```diff
@@ -0,0 +1,127 @@
...
+"""
+
```
fstagni: BTW the spacing of this file is 4 instead of 2

Interfaces/API/Dirac.py
```diff
@@ -65,6 +66,25 @@ def parseArguments(args):
     argList += arg.split(',')
   return argList
 
+def fileCatalogAPIWrapper(func):
```
chaen: that's more a debugging function for you, not really something to go to the API

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: This method is a hook, and it works "as is" w/o overwriting it. I added an underscore to the method name to make it clear that it is an internal method.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: changed it to read: 
`if fqPname in result['Value']:`

I don't see why the original code (present all over the places in the same file would be not Python3 friendly. The `in` operator works file in Python 3 even thought `.keys()` does not return a list.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: Removed.

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
martynia: The method is obsolete. Removed.

DataManagementSystem/DB/FileCatalogComponents/MultiVODirectoryMetadata.py
martynia: Fixed.

tests/Integration/Resources/Catalog/Test_UserMetadata.py
martynia: Fixed.

tests/Integration/Resources/Catalog/Test_UserMetadata.py
martynia: Will eventually do, this is the plan.

Interfaces/API/Dirac.py
martynia: I'm not insisting here. I just wanted to add all the client-side API call to already exported server-side calls, assuming that someone you had exported them had a reason to do so.

Interfaces/API/Dirac.py
martynia: Well, this is a wrapper to hide away all the logging code which is identical for most of the methods. Otherwise a lot of code repetition would occur.

martynia:
Your inheritance proposal would work in some cases but it would not work everywhere. Look at the 
`__expandMetaDictionary(self, metaDict, credDict)` method. It uses a line:
`suffix = self._getMetaNameSuffix(credDict)`
to identify the suffix. This cannot be done by changing the derived method parameter. At least not at the first glance. Let me think about it.

DataManagementSystem/Service/FileCatalogHandler.py
```diff
@@ -523,16 +535,41 @@ def export_removeMetadata(self, pathMetadataDict):
...
+    gLogger.debug("Directory metadata RPC",metaDict)
+    # code below breaks client-side checks
+    suffix = Utilities.fqMetaNameSuffix(self.getRemoteCredentials())
+    if metaDict['OK']:
+      if metaDict['Value']:
+        metaDict['Value'] = {key.replace(suffix,'') if key.endswith(suffix)
+                             else key:value for key, value in metaDict['Value'].iteritems()}
+      if metaDict['MetadataOwner']:
+        metaDict['MetadataOwner']= {key.replace(suffix,'') if key.endswith(suffix)
+                             else key:value for key, value in metaDict['Value'].iteritems()}
```
atsareg: Should it be "else key:value for key, value in metaDict['MetadataOwner'].iteritems()}" ?

DataManagementSystem/Service/FileCatalogHandler.py
```diff
@@ -523,16 +535,41 @@ def export_removeMetadata(self, pathMetadataDict):
...
+
+      if metaDict['MetadataType']:
+        metaDict['MetadataType']= {key.replace(suffix,'') if key.endswith(suffix)
+                                    else key:value for key, value in metaDict['Value'].iteritems()}
```
atsareg: idem

Resources/Catalog/FileCatalogClient.py
```diff
@@ -292,7 +292,7 @@ def getFileUserMetadata(self, path, timeout=120):
     """
     directory = "/".join(path.split("/")[:-1])
     rpcClient = self._getRPC(timeout=timeout)
-    result = rpcClient.getFileUserMetadata(path)
+    result = rpcClient.findDirectoriesByMetadata(path)
```
atsareg: Is it intentional ?

atsareg:
You use user group as the metadata specification. While the intention is to use per VO metadata. Imagine the practical case: VO production manager produces data and users access it. The metadata will belong to vo_prod group and the user will look it up with the vo_user group and will not find any data.

DataManagementSystem/Service/FileCatalogHandler.py
martynia: Good spot !

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
atsareg: Why do you need at all getMetaName and getMetaNameSuffix in the FC classes ? Just use methods from the Utilities module. The same for the File metadata, the utility functions are the same.

atsareg: The functions can get argument specifying the naming method (no modifiction, by group, by VO, by ...). This can be defined in the catalog configuration. Then you do not need inherited classes to just implement this function

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -248,7 +278,8 @@ def removeMetadata(self, dpath, metadata, credDict):
 
   def setMetaParameter(self, dpath, metaName, metaValue, credDict):
     """ Set an meta parameter - metadata which is not used in the the data
-        search operations
+        search operations.
+        Modified by JM: make metaName Dirac group aware.
```
atsareg: Comments like that are suitable for release notes but for the in-code documentation

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -508,10 +539,14 @@ def __findSubdirMissingMeta(self, meta, pathSelection):
   def __expandMetaDictionary(self, metaDict, credDict):
     """ Expand the dictionary with metadata query
     """
-    result = self.getMetadataFields(credDict)
+    result = self.getMetadataFields(credDict) # this now returns a fully qualified meta names, unlike the query !
```
atsareg: Again this is a comment reflecting this particular release event, not relevant in the long term

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -585,14 +620,17 @@ def findDirIDsByMetadata(self, queryDict, path, credDict):
     # Now check the meta data for the requested directory and its parents
     finalMetaDict = dict(metaDict)
     for meta in metaDict.keys():
-      result = self.__checkDirsForMetadata(meta, metaDict[meta], pathString)
+      fqmeta = self.getMetaName(meta, credDict) # expand table name
+      # use fqmeta whet it is meant to be a table name
+      result = self.__checkDirsForMetadata(fqmeta, metaDict[meta], pathString)
+      print "__checkDirsForMetadata result", result
```
atsareg: This print is also removed ?

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -143,9 +151,12 @@ def setMetadata( self, path, metadict, credDict ):
 
     return S_OK()
 
-  def removeMetadata( self, path, metadata, credDict ):
-    """ Remove the specified metadata for the given file
+  def removeMetadata( self, path, metadata_r, credDict ):
```
atsareg: metadata_r variable name is not following naming convention (rawMetadat ?)

martynia:
> You use user group as the metadata specification. While the intention is to use per VO metadata. Imagine the practical case: VO production manager produces data and users access it. The metadata will belong to vo_prod group and the user will look it up with the vo_user group and will not find any data.

Hi,

The code I'm working on now has a following statement, when resolving the suffix (we have discussed that during the meeting):

```python 
def _getMetaNameSuffix(self, credDict):
    """
    Get a VO specific suffix from user credentials.

    :param credDict: user credentials
    :return: VO specific suffix
    """
    vo = Registry.getGroupOption(credDict['group'], 'VO')
    return '_' + vo
```

Interfaces/API/Dirac.py
atsareg: Please, do add FC related stuff to Dirac.py. FileCatalog() is itself the API.

Resources/Catalog/FileCatalog.py
```diff
@@ -293,6 +293,7 @@ def w_execute( self, *parms, **kws ):
     else:
       # FIXME: Return just master result here. This is temporary as more detailed
       # per catalog result needs multiple fixes in various client calls
+      # JM non-lfn methods would return {} if not a master...
```
atsareg: Do not make personal comments please

fstagni:
Did you close this to re-open in v6r22?

atsareg:
No, this was closed by mistake

atsareg:
Though we have to decide which release it goes to. The changes are relatively contained to go into a patch. But they should be Jenkins-tested. Also, I do not see the need to make special FC components, everything can be in the original component code. This will not speed up testing or make the release safer because most of the modifications are in the original components anyway.

chaen:
Hint1: v7r1
Hint2: I advocate again that original components should be left untouched as much as possible

andresailer:
Maybe @martynia could push his updates?

martynia:
> Maybe @martynia could push his updates?

I would do it, need to address some comments still. I agree that we should discuss the current status, not the history.

Resources/Catalog/FileCatalogClient.py
martynia: Not sure why this happened. It clearly look like my commit.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: OK

Interfaces/API/Dirac.py
martynia: OK, but there was some FC related stuff there already. I've just added more ;-). What should go and what should stay in Dirac.py ?

atsareg: The only pure FC method exposed through Dirac.py is getAllReplicas() which I think was a mistake to introduce. All the others do something on top of the FC queries. So, I do not think you should add anything at all to Dirac.py

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -57,26 +85,27 @@ def addMetadataField(self, pname, ptype, credDict):
...
-    result = self.__transformMetaParameterToData(pname)
+    result = self.__transformMetaParameterToData(fqPname)
     if not result['OK']:
       return result
 
     return S_OK("Added new metadata: %d" % metadataID)
 
-  def deleteMetadataField(self, pname, credDict):
-    """ Remove metadata field
+  def deleteMetadataField(self, r_pname, credDict):
```
atsareg: r_pname -> rawPname

martynia:
I addressed most of the comments you made, less:

- `Dirac.py` API extension. This is still to be modified and moved elsewhere,

- moved the integration tests about, but still not in Jenkins (work in progress)

I defined both 'hook' methods to be private, keeping the inheritance tree, as before. One of them ('the suffix stripper') is called from the Handler -making it private required some more changes in the code (look for `strip_suffix` default parameter in some methods).

There might still be some minor things to change, but I would appreciate your comments about the concept first.

 Thanks, JM

martynia:
Hi,

I committed my changes last Friday and I was wondering if your have been notified. My other concern is that
Travis seemed to run the old code:

https://travis-ci.org/DIRACGrid/DIRAC/jobs/540768734 <https://travis-ci.org/DIRACGrid/DIRAC/jobs/540768734>

Could you give it (Travis) a kick ? Alternatively I could commit a small change to activate it.

   best, JM

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -57,26 +85,27 @@ def addMetadataField(self, pname, ptype, credDict):
       valueType = "VARCHAR(64)"
 
     req = "CREATE TABLE FC_Meta_%s ( DirID INTEGER NOT NULL, Value %s, PRIMARY KEY (DirID), INDEX (Value) )" \
-        % (pname, valueType)
+        % (fqPname, valueType)
```
andresailer: I think there are VOs with hyphens in their name, so the TableName now needs to be surrounded by backticks, doesn't it?

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -507,102 +550,102 @@ def __findFilesByMetadata( self, metaDict, dirList, credDict ):
...
-    result = self.getFileMetadataFields( credDict )
+    #     fileMetaDict = {}
+    result = self.getFileMetadataFields(credDict)  # this returns fully qualified meta name
     if not result['OK']:
       return result
     fileMetaKeys = result['Value'].keys() + FILE_STANDARD_METAKEYS.keys()
-    fileMetaDict = dict( item for item in metaDict.items() if item[0] in fileMetaKeys )
-
+    fileMetaDict = dict(item for item in metaDict.items()
+                        if item[0] + self._getMetaNameSuffix(credDict) in fileMetaKeys)
```
andresailer:
```suggestion
                        if self._getMetaName(item[0], credDict) in fileMetaKeys)
```

DataManagementSystem/DB/FileCatalogComponents/MultiVODirectoryMetadata.py
```diff
@@ -0,0 +1,39 @@
...
+
+class MultiVODirectoryMetadata(DirectoryMetadata):
+  """
+  MULti-VO FileCatalog plugin implementation.
+  """
+
+  def __init__(self, database=None):
+    DirectoryMetadata.__init__(self, database=database)
+
+  def _getMetaName(self, meta, credDict):
```
andresailer: Maybe this and getMetaNameSuffix can be turned into a Mixin class and added to this and MultiVOFileMetadata

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -101,7 +130,11 @@ def getMetadataFields(self, credDict):
     metaDict = {}
     for row in result['Value']:
       metaDict[row[0]] = row[1]
-
+    # strip the VO suffix, if required
+    if strip_suffix:
+      suffix = self._getMetaNameSuffix(credDict)
+      metaDict = {key.replace(suffix, ''): value for key, value in metaDict.iteritems()
```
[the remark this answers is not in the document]
martynia: I changed the statement to read:
```python
metaDict = {key.rsplit(suffix, 1)[0] if key.endswith(suffix)
                  else key: value for key, value in metaDict.iteritems()}
```
It solves a problem when a user defined meta name ends with the _ and vo name (like __gridpp)_
Addressing your second comment: for file metadata the original dict only uses _tablename_metaname_ keys, all of them are extended by __voname_. For directories also other keys are present which have to be kept. I think it is much safer to use the 'keep other keys' approach, as above for all cases.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -512,6 +557,10 @@ def __expandMetaDictionary(self, metaDict, credDict):
     if not result['OK']:
       return result
     metaTypeDict = result['Value']
+    suffix = self._getMetaNameSuffix(credDict)
+    # remove the suffix - wrong, creates problem with selections down the line.
```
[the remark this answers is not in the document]
martynia: It must have been a comment I left for myself and forgot to remove it.

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -13,442 +13,486 @@
...
     if not result['OK']:
       return result
 
-    return S_OK( "Added new metadata: %d" % metadataID )
+    return S_OK("Added new metadata: %d" % metadataID)
 
-  def deleteMetadataField( self, pname, credDict ):
-    """ Remove metadata field
+  def deleteMetadataField(self, pname, credDict):
+    """ Remove metadata field (only from user's own group)
```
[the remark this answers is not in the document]
martynia: Sure

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -13,442 +13,486 @@
...
               return result
           else:
             return result
 
     return S_OK()
 
-  def removeMetadata( self, path, metadata, credDict ):
-    """ Remove the specified metadata for the given file
+  def removeMetadata(self, path, rawMetadata, credDict):
+    """ Remove the specified metadata for the given file (for user's own group only)
```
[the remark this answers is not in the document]
martynia: ditto

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
martynia: Yes, much cleaner.

DataManagementSystem/DB/FileCatalogComponents/MultiVODirectoryMetadata.py
martynia: Yes, could do that. We would need 2 mixing classes: BaseMixin (pass through) and DerivedMixin one to maintain the suffix adding and removal. Base FileMetadata and DirectoryMetadata classes would both inherit from the BaseMixin and the Multi VO counterparts would inherit from the DerivedMixin. If it is what you meant I can implement this. I'll obviously find suitable names for the mixin classes.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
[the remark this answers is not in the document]
andresailer: If you always use the same approach, then maybe turn this into a function.

DataManagementSystem/DB/FileCatalogComponents/MultiVODirectoryMetadata.py
andresailer: Yes. Reduces code duplication

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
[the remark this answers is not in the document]
martynia: In the brand new mixin ....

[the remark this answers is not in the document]
andresailer: :+1:

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: This is not the only place which suffers from this problem. FileMetadata.py has similar statements all over the place. The problem existed even before, if a user decided to name his metadata using hyphens. I can bactick all the occurrences if you find it  a right way to go, although a really robust solution would be to use sqlalchemy, like in other Dirac modules. This would be a big change, though.

andresailer: At the moment hyphens in metadata names are not supported, which people can find a workaround. It might be harder to rename VOs for that reason.
Maybe instead of adding backticks everywhere one can replace hyphens in the getMetaNameSuffix method?

andresailer:
Side note: At the moment the Transformation InputDataAgent uses the host credentials to find files by metadata, which would have the "hosts" group associated to it, which doesn't belong to any VO.
One would have to make some adjustments to make the multiVO metadata also work nicely with a multiVO InputDataAgent.

martynia:
Thanks,
We will have to take care of that.

   best, JM

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -350,6 +384,17 @@ def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
...
+    if strip_suffix:
+      suffix = self._getMetaNameSuffix(credDict)
+      metaDict = {key.replace(suffix, '') if key.endswith(suffix)
+                  else key: value for key, value in metaDict.iteritems()}
+
+      metaOwnerDict = {key.replace(suffix, '') if key.endswith(suffix)
+                       else key: value for key, value in metaOwnerDict.iteritems()}
+
+      metaTypeDict = {key.replace(suffix, '') if key.endswith(suffix)
+                      else key: value for key, value in metaTypeDict.iteritems()}
```
fstagni: What about
```
def stripSuffix(aDict, suffix):
  return {key.replace(suffix, '') if key.endswith(suffix)
              else key: value for key, value in aDict.iteritems()}

if strip_suffix:
  metaDict = stripSuffix(metaDict, suffix)
  metaOwnerDict = stripSuffix(metaOwnerDict, suffix)
  metaTypeDict = stripSuffix(metaTypeDict, suffix)

```

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -302,7 +336,7 @@ def getDirectoryMetaParameters(self, dpath, credDict, inherited=True, owndata=Tr
 
     return S_OK(metaDict)
 
-  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
+  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True, strip_suffix=False):
```
fstagni: normally we use `camelCase` instead of `pascal_case`

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -13,442 +13,486 @@
...
-    return S_OK( fileID )
+      return S_ERROR('File not found')
+    return S_OK(fileID)
 
-  def __setFileMetaParameter( self, fileID, metaName, metaValue, credDict ):
+  def __setFileMetaParameter(self, fileID, metaName, metaValue, credDict):
     """ Set an meta parameter - metadata which is not used in the the data
-        search operations
+        search operations.
+        Modified by JM: make metaName Dirac group aware.
```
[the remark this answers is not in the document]
fstagni: ...it seems you committed something in your local repo, but it didn't yet reach here.

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
```diff
@@ -13,442 +13,486 @@
...
 
-  def setFileMetaParameter( self, path, metaName, metaValue, credDict ):
+  def _getMetaName(self, meta, credDict):
+    """
+    Return a metadata name based on client supplied meta name and client credentials
+    For the base class it just returns the name passed in.
+    This method is a 'hook' and is meant to be overwritten by derived classes.
+
+    :param meta:  meta name
+    :param credDict: client credentials
```
fstagni:
```
    :param str meta:  meta name
    :param dict credDict: client credentials
```
(also later)

Interfaces/API/Dirac.py
fstagni: Please drop them @martynia

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,128 @@
...
+  def test_ReplicasByMetadata(self):
+    metaDict6 = {'JMMetaInt6': 13}
+    result = self.dirac.getReplicasByMetadata(metaDict6, path='/')
+    self.assertFalse(result['OK'])
+
+
+if __name__ == '__main__':
+  suite = unittest.defaultTestLoader.loadTestsFromTestCase(TestUserMetadataTestCase)
+  suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(testMetadata))
+  testResult = unittest.TextTestRunner(verbosity=2).run(suite)
```
fstagni: Add at the end here:
```
  sys.exit(not testResult.wasSuccessful())
```

DataManagementSystem/DB/FileCatalogComponents/FileMetadata.py
martynia: These have been addressed in the last push. (mix-in)

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: Not there anymore. Added mixing classes as suggested by Andre.

Interfaces/API/Dirac.py
martynia: I understand you don't want to pollute Dirac.py with FC stuff, but it would be very useful to manipulate  user metadata programmatically (and there is server code for that). So is there any suitable place to move the client APi I have written ?

tests/Integration/DataManagementSystem/Test_UserMetadata.py
martynia: Done

Interfaces/API/Dirac.py
martynia: And the tests would be useless since they use the client methods.

fstagni:
Please rebase on `integration` branch.

martynia:
> Please rebase on `integration` branch.
Done. Waiting for tests to complete.

martynia:
Hi,

One topic which was (I hope) covered by my colleagues  during the last Dirac meeting is relocation of the File Catalog  user metadata API. 

Is the  directory:

Resources/Catalog/ 

a suitable place to move the user metadata stuff from Dirac.py ?

best, JM

chaen:
Hi,

yes they mentioned something about it, but I have not looked at it in detail yet.
I am not really sure what you are trying to achieve here. Can you elaborate please ?

chaen:
Maybe a bit more meat: `FileCatalog` is the API when you need to do things from within DIRAC.
If you want users to be exposed to some function, then `Dirac.py` makes sense. At the same time, it should not be too messy. So since there is already a `Job.py`, maybe we can create another file. 
But again, it all depends what you are trying to do

martynia:
Well,

I originally put client calls related to user metadata handling into Dirac.py. The client calls use existing RPC server-side calls, so I was not very innovative here.

The client operations serve both testing purposes (‘within Dirac', in your speak) and should be exposed to the users (or developers creating tools like data movers etc.), so they can manipulate data programmatically. If Dirac.py is too long (and IMHO it is), we could create another file in the same directory, i.e. UserMetadata.py or move it to the FileCatalog API. 

 If we drop the API completely that the only way users could manipulate metadata would be by using the File Catalog CLI. Tool developers would be stuck.

From the discussion on GitHub the opposition to keeping the API in Dirac.py is strong. So where move it to ?

Best, JM

chaen:
For developers the interface to the file catalog *is* `FileCatalog.py`. 
For users, I am not sure, our users do not use the metadata of DFC

martynia:
Removed FC API from Dirac.py + some bug fixes + autopep8

DataManagementSystem/DB/FileCatalogComponents/MultiVOFileMetadata.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class MultiVOFileMetadata(MultiVOMetaNameMixIn, FileMetadata):
+  """
+  MULti-VO FileCatalog plugin implementation.
+  """
+
+  def __init__(self, database=None):
+    FileMetadata.__init__(self, database=database)
+    MultiVOMetaNameMixIn.__init__(self)
```
fstagni: These inheritances are partly broken. You are in this situation right now:

```
In [1]: class MetaNameMixInNoObject():
   ...:   def __init__(self):
   ...:     print("I am MetaNameMixInNoObject")
   ...: 
   ...: 
   ...: class MultiVOMetaNameMixInNoObject(MetaNameMixInNoObject):
   ...:   def __init__(self):
   ...:     print("I am MultiVOMetaNameMixInNoObject")
   ...: 
   ...: 
   ...: class DirectoryMetadataNoObject(MetaNameMixInNoObject):
   ...:  def __init__(self):
   ...:   print("I am DirectoryMetadataNoObject")
   ...: 
   ...: 
   ...: class MultiVODirectoryMetadataNoObject(MultiVOMetaNameMixInNoObject, DirectoryMetadataNoObject):
   ...:   def __init__(self):
   ...:     print("I am MultiVODirectoryMetadataNoObject")
   ...:     DirectoryMetadataNoObject.__init__(self)
   ...:     MultiVOMetaNameMixInNoObject.__init__(self)
   ...: 
   ...: 
   ...: MultiVODirectoryMetadataNoObject()
   ...: 
I am MultiVODirectoryMetadataNoObject
I am DirectoryMetadataNoObject
I am MultiVOMetaNameMixInNoObject
```
As you see what's in `MetaNameMixInNoObject` is not executed. This is because you are using old style classes: http://python-history.blogspot.com/2010/06/method-resolution-order.html

This might not be a problem with the current code because there's nothing really done in  `MetaNameMixInNoObject`, but we shouldn't rely on that.


What should instead be done is the following (inheriting from `object` and using `super`:
```
In [1]: class MetaNameMixIn(object):
   ...:   def __init__(self):
   ...:     print("I am MetaNameMixIn")
   ...: 
   ...: 
   ...: class MultiVOMetaNameMixIn(MetaNameMixIn):
   ...:   def __init__(self):
   ...:     super(MultiVOMetaNameMixIn, self).__init__()
   ...:     print("I am MultiVOMetaNameMix")
   ...: 
   ...: 
   ...: class DirectoryMetadata(MetaNameMixIn):
   ...:   def __init__(self):
   ...:     super(DirectoryMetadata, self).__init__()
   ...:     print("I am DirectoryMetadata")
   ...: 

In [2]: class MultiVODirectoryMetadataWithSuper(MultiVOMetaNameMixIn, DirectoryMetadata):
   ...:   def __init__(self):
   ...:     super(MultiVODirectoryMetadataWithSuper, self).__init__()
   ...:     print("I am MultiVODirectoryMetadataWithSuper")
   ...: 

In [3]: MultiVODirectoryMetadataWithSuper()
I am MetaNameMixIn
I am DirectoryMetadata
I am MultiVOMetaNameMix
I am MultiVODirectoryMetadataWithSuper
```

Also, python 3 by default would use the MRO of new-style classes.

martynia: Sorry for not replying straight away, but I was on holidays and I'm taking some days off next week again ...

I totally agree I should be using new style classes, I'll fix this.  However your example above is not quite accurate since I explicitly call base class inits:
https://github.com/DIRACGrid/DIRAC/blob/0d83eddd103d050edcec6abaf58b2ea32f3ba532/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py#L16-L19

and also:

https://github.com/DIRACGrid/DIRAC/blob/0d83eddd103d050edcec6abaf58b2ea32f3ba532/DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py#L15-L16

Another problem with `super` is that all `init` prototypes have to agree, and in our example we have an additional `database=`  argument in one of them. I could pass it as a default to another one but it would not look nice.

How do you see this ?

   best, JM

fstagni: You can define `def __init__(self, *args, **kwargs):` for absorbing all the arguments.

martynia: Hi, the inheritance tree has been updated.

fstagni:
@martynia please just fix https://travis-ci.org/DIRACGrid/DIRAC/jobs/572821909#L2693 in my view this would be the last change to be done here.

At the same time I am not the most expert in the code written here, so I'd like @atsareg @chaen and @andresailer to give a final review.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -350,6 +362,13 @@ def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
       for meta in result['Value']:
         metaOwnerDict[meta] = 'OwnParameter'
 
+    if strip_suffix:
+      #      metaDict = {key.replace(suffix, '') if key.endswith(suffix)
```
andresailer: It is not necessary to keep these commented lines

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -512,6 +531,9 @@ def __expandMetaDictionary(self, metaDict, credDict):
     if not result['OK']:
       return result
     metaTypeDict = result['Value']
+    # strip the suffix
```
andresailer: That  comment is not adding anything useful

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -302,7 +314,7 @@ def getDirectoryMetaParameters(self, dpath, credDict, inherited=True, owndata=Tr
 
     return S_OK(metaDict)
 
-  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
+  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True, strip_suffix=False):
```
andresailer: It would be helpful to have a description under which circumestances strip_suffix true or false should be used

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -604,9 +626,9 @@ def findDirIDsByMetadata(self, queryDict, path, credDict):
       first = True
       for meta, value in finalMetaDict.items():
         if value == "Missing":
-          result = self.__findSubdirMissingMeta(meta, pathSelection)
+          result = self.__findSubdirMissingMeta(fqmeta, pathSelection)
```
andresailer: Should this still be `meta`, given that this comes from the loop variable two lines above?

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
```diff
@@ -604,9 +626,9 @@ def findDirIDsByMetadata(self, queryDict, path, credDict):
...
         else:
-          result = self.__findSubdirByMeta(meta, value, pathSelection)
+          result = self.__findSubdirByMeta(fqmeta, value, pathSelection)
```
andresailer: Should this still be meta, given that this comes from the loop variable four lines above?

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
```diff
@@ -0,0 +1,51 @@
...
+  MULti-VO MetaName MixIn implementation.
+  """
+
+  def __init__(self, *args, **kwargs):
+    super(MultiVOMetaNameMixIn, self).__init__(*args, **kwargs)
+
+  def getMetaName(self, meta, credDict):
+    """
+    Return a fully-qualified metadata name based on client-suplied metadata name and
+    client credentials. User group is added to the metadata passed in.
```
andresailer: It is the users VO that is added to the metadata name

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
```diff
@@ -0,0 +1,51 @@
...
+    :param credDict: user credentials
+    :return: VO specific suffix
+    """
+    vo = Registry.getGroupOption(credDict['group'], 'VO')
+    return '_' + vo.replace('-', '_')
+
+  def stripSuffix(self, metaDict, credDict):
+    """
+    Strip the suffix from all keys which contain it, leaving other keys intact.
+    :param metaDict: original dict
```
andresailer: Need a newline before the first `:param ...`

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
```diff
@@ -0,0 +1,51 @@
...
+    :param credDict: user credential dictionary
+    :return: a new dict with modified keys
+    """
+
+    suffix = self.getMetaNameSuffix(credDict)
+    smetaDict = {key.rsplit(suffix, 1)[0]: value for key, value in metaDict.iteritems()
```
andresailer: This removes all entries that do not end with suffix, which doesn't sound like what is intended given the docstring

chaen:
I let @andresailer decide since it touches only File/Directory Metadata managers, which we do not use

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
martynia: done

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
martynia: A uses only gets its own VO data, so stripping all the other keys is intended. The comment is wrong.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: I think we need a line:
```python
fqmeta = self.getMetaName(meta, credDict)
```
as a first line in both loops, so ```python __findSubdirByMeta ``` looks into a correct table.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: as above.

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
martynia: Yes.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata.py
martynia: ```strip_suffix=true```  is to be used when a call is meant to be in a call chain which returns data to a user, otherwise it should be ``` false``` to maintain backward compatibility since that method is used internally as well.

martynia:
I addressed the comments a few days ago. Is it anything else which should be done on this ?
Thanks, JM

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -27,22 +29,25 @@ def setDatabase(self, database):
...
+    if fqPname in result['Value']:
       return S_ERROR('The metadata %s is already defined for Files' % pname)
 
     result = self.getMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      if ptype.lower() == result['Value'][pname].lower():
+
+    if fqPname in result['Value'].keys():
```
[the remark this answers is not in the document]
martynia: OK

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -37,56 +39,58 @@ def addMetadataField(self, pname, ptype, credDict):
...
+    fqPname = self.getMetaName(pname, credDict)
+    if fqPname in result['Value'].keys():
+      return S_ERROR('The metadata %s is already defined for Directories' % fqPname)
 
     result = self.getFileMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      if ptype.lower() == result['Value'][pname].lower():
+    if fqPname in result['Value'].keys():
```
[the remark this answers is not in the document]
martynia: OK

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,140 @@
+"""
+Test of multi-VO user metadata handling. Assumes a running Dirac instance with the (master?) FileCatalog
+"""
+
```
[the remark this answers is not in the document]
martynia: OK, added to other new files.

DataManagementSystem/DB/FileCatalogComponents/MultiVOMetaNameMixIn.py
```diff
@@ -0,0 +1,52 @@
+""" DIRAC Multi VO MixIn class to manage file metadata and directory for multiple VO.
+"""
+
```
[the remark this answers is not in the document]
martynia: Done.

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,140 @@
...
+    # meta index -d
+    result = self.fc.addMetadataField('JMTestDirectory6', 'INT', metaType='-d')
+    self.assertTrue(result['OK'])
+    self.assertNotEqual(result['Value'], 'Already exists')
+    self.assertTrue(result['Value'].startswith('Added new metadata:'))
+
+    # meta show
+    result = self.fc.getMetadataFields()
+    self.assertTrue(result['OK'])
+    self.assertDictContainsSubset({'JMMetaInt6': 'INT', }, result['Value']['FileMetaFields'])
```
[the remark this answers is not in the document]
martynia: It must have been a typo.

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,140 @@
...
+    # find
+    metaDirDict = {'JMTestDirectory6': 126}
+    result = self.fc.findDirectoriesByMetadata(metaDirDict, path='/')
+    self.assertTrue(result['OK'])
+    self.assertIn(self.dir5, result['Value'].values())
+
+    # API call only
+    result = self.fc.getFileUserMetadata(self.lfn5)
+    self.assertTrue(result['OK'])
+    # 'Value': {'JMMetaInt6': 13L, 'JMMetaInt5': 12L, 'JMTestDirectory': 124L}
```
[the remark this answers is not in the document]
martynia: No, removed.

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -37,56 +39,58 @@ def addMetadataField(self, pname, ptype, credDict):
     result = self.db.dmeta.getMetadataFields(credDict)
     if not result['OK']:
       return result
-    if pname in result['Value'].keys():
-      return S_ERROR('The metadata %s is already defined for Directories' % pname)
+    # existing pnames are fully qualified, so
+    fqPname = self.getMetaName(pname, credDict)
+    if fqPname in result['Value'].keys():
```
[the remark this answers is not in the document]
martynia: Done.

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,140 @@
...
+    result = self.dirac.removeFile(self.lfn5)
+    self.assertTrue(result['OK'])
+    os.remove(self.fullPath)
+
+
+class testMetadata(TestUserMetadataTestCase):
+  def test_AddQueryRemove(self):
+    result = self.dirac.getLfnMetadata(self.lfn5)
+    self.assertTrue(result['OK'])
+    self.assertTrue(self.lfn5 in result['Value']['Successful'].keys())
```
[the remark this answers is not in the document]
martynia: Done.

tests/Integration/DataManagementSystem/Test_UserMetadata.py
```diff
@@ -0,0 +1,140 @@
...
+from DIRAC.Interfaces.API.Dirac import Dirac
+from DIRAC.Resources.Catalog.FileCatalog import FileCatalog
+from DIRAC import gLogger
+from DIRAC.Core.Utilities.Time import toString
+
+
+def random_dd(outfile, size_mb):
+  import os
+  with open(outfile, 'w') as f:
+    for i in range((int(size_mb) * 2 ** 20) / 512):
```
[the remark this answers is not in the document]
martynia: Fixed.

DataManagementSystem/DB/FileCatalogComponents/MetaNameMixIn.py
```diff
@@ -0,0 +1,43 @@
+""" DIRAC base MixIn class to manage file metadata and directory names.
+"""
+
+
+class MetaNameMixIn(object):
+
+  def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
martynia: Done.

fstagni:
Just rebase or merge the `integration` branch and then I think it can be merged.

martynia:
I have synced  my branch with integration right now. It says it could be merged after review.
Best, JM

DataManagementSystem/Service/FileCatalogHandler.py
```diff
@@ -524,16 +526,26 @@ def export_removeMetadata(self, pathMetadataDict):
   types_getDirectoryUserMetadata = [StringTypes]
 
   def export_getDirectoryUserMetadata(self, path):
-    """ Get all the metadata valid for the given directory path
+    """ Get all the metadata valid for the given directory path.
+        Client gets metadata w/o the VO suffix.
     """
-    return gFileCatalogDB.dmeta.getDirectoryMetadata(path, self.getRemoteCredentials())
+    metaDict = gFileCatalogDB.dmeta.getDirectoryMetadata(path, self.getRemoteCredentials(), enableStripping=False)
```
andresailer: This line (and corresponding debug output) has to be dropped, there is no point querying the database twice only for debug reasons

martynia:
@zmathe could you please review this pull request ? Thanks, JM

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -90,7 +96,7 @@ def deleteMetadataField(self, pname, credDict):
         result["Message"] = error + "; " + result["Message"]
     return result
 
-  def getMetadataFields(self, credDict):
+  def getMetadataFields(self, credDict, enableStripping=False):
```
zmathe: In order to keep backward compatibility I would set enableStripping=True
And probably better to use stripVO ?

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -28,22 +30,25 @@ def setDatabase(self, database):
   def addMetadataField(self, pname, ptype, credDict):
     """ Add a new metadata parameter to the Metadata Database.
         pname - parameter name, ptype - parameter type in the MySQL notation
+        Modified to use fully qualified metadata names.
```
zmathe: probably better to update the doc string. Please use
:param str pname: ....

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -211,7 +221,7 @@ def setMetadata(self, dpath, metadict, credDict):
     return S_OK()
 
   def removeMetadata(self, dpath, metadata, credDict):
-    """ Remove the specified metadata for the given directory
+    """ Remove the specified metadata for the given directory for users own VO
```
zmathe: please add the method variables to the doc string

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -249,7 +261,7 @@ def removeMetadata(self, dpath, metadata, credDict):
 
   def setMetaParameter(self, dpath, metaName, metaValue, credDict):
     """ Set an meta parameter - metadata which is not used in the the data
-        search operations
+        search operations.
```
zmathe: here as well

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
```diff
@@ -303,10 +315,11 @@ def getDirectoryMetaParameters(self, dpath, credDict, inherited=True, owndata=Tr
 
     return S_OK(metaDict)
 
-  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True):
+  def getDirectoryMetadata(self, path, credDict, inherited=True, owndata=True, enableStripping=False):
```
zmathe: I have the same comment regarding to the enableStripping variable as above

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -11,25 +11,27 @@
...
-##############################################################################
+  ##############################################################################
+  #
+  #  Manage Metadata fields
+  #
+  ##############################################################################
   def addMetadataField(self, pname, ptype, credDict):
     """ Add a new metadata parameter to the Metadata Database.
-        pname - parameter name, ptype - parameter type in the MySQL notation
+        pname - parameter name, ptype - parameter type in the MySQL notation.
```
zmathe: please use the proper doc string for the method variable

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -38,56 +40,58 @@ def addMetadataField(self, pname, ptype, credDict):
...
-    result = self.__transformMetaParameterToData(pname)
+    result = self.__transformMetaParameterToData(fqPname)
     if not result['OK']:
       return result
 
     return S_OK("Added new metadata: %d" % metadataID)
 
   def deleteMetadataField(self, pname, credDict):
-    """ Remove metadata field
+    """ Remove metadata field (only from user's own VO)
```
zmathe: please add the method variables to the docscring

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -38,56 +40,58 @@ def addMetadataField(self, pname, ptype, credDict):
...
-    req = "DELETE FROM FC_FileMetaFields WHERE MetaName='%s'" % pname
+    req = "DELETE FROM FC_FileMetaFields WHERE MetaName='%s'" % fqPname
     result = self.db._update(req)
     if not result['OK']:
       if error:
         result["Message"] = error + "; " + result["Message"]
     return result
 
-  def getFileMetadataFields(self, credDict):
+  def getFileMetadataFields(self, credDict, enableStripping=False):
```
zmathe: same comment as above

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -100,16 +104,20 @@ def getFileMetadataFields(self, credDict):
...
+  ###########################################################
+  #
+  # Set and get metadata for files
+  #
+  ###########################################################
 
   def setMetadata(self, path, metadict, credDict):
-    """ Set the value of a given metadata field for the the given directory path
+    """ Set the value of a given metadata field for the the given directory path.
+        Modified to use fully qualified metadata name.
```
zmathe: docstring: same as above

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -143,8 +150,9 @@ def setMetadata(self, path, metadict, credDict):
     return S_OK()
 
   def removeMetadata(self, path, metadata, credDict):
-    """ Remove the specified metadata for the given file
+    """ Remove the specified metadata for the given file (for user's own VO only)
```
zmathe: docstring

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -193,11 +203,11 @@ def __getFileID(self, path):
 
   def __setFileMetaParameter(self, fileID, metaName, metaValue, credDict):
     """ Set an meta parameter - metadata which is not used in the the data
-        search operations
+        search operations. metaName is VO aware.
```
zmathe: docstring

DataManagementSystem/DB/FileCatalogComponents/FileMetadata/FileMetadata.py
```diff
@@ -238,7 +248,7 @@ def _getFileUserMetadataByID(self, fileIDList, credDict, connection=False):
 
     return S_OK(metaDict)
 
-  def getFileUserMetadata(self, path, credDict):
+  def getFileUserMetadata(self, path, credDict, enableStripping=False):
```
zmathe: same as above

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
martynia: This particular method is used in the handler in which case the VO suffix has to be stripped of and multiple times internally in its class where the suffix has to be present. Flipping the default would require more typing but it would not change anything. The backward compatibility with the single VO system is not maintained by this default but by loading a correct plugin - _DirectoryMetadata_ or _MultiVODirectoryMetadata_ class. In a single VO case the suffix is "stripped" as well, but it is defined to be an empty string so stripping has no effect.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
martynia: I updated the docstrigs in the methods I modified to my best knowledge. Since I haven't written the code in the first place, some params description might be incomplete or even incorrect.

DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/DirectoryMetadata.py
fstagni: I think it's fine to keep it `False` as default there.

atsareg:
Returning back to this PR. I agree with Chris and others that the new components should be coded in the way that does not affect the original components. I am sure this can be done.

atsareg:
One question: you have put stripVO flag in some methods with False by default. What is the intended usage of it ? The service is always setting to True. I think you can drop it assuming that the suffix is always stripped off. This will also allow you to maintain the same interface for the new components as for the old ones.

martynia:
Hi,

I have answered this question on some occasions before. Methods in question are used (not my decision) either by a server RPC call (in which case the suffix is to be stripped off, just before returning to a user) and internally as convenience methods, in which case the suffix is not to be stripped, because convenience methods act on fully qualified names (with the suffix). 

Hope it helps, JM

martynia:
@fstagni could you review, please

martynia:
@atsareg  

> One question: you have put stripVO flag in some methods with False by default. What is the intended usage of it ? The service is always setting to True. I think you can drop it assuming that the suffix is always stripped off. This will also allow you to maintain the same interface for the new components as for the old ones.

I replied to this. Are you satisfied with my answer or do you have other comments ?

fstagni:
IIUC this can be closed, as replaced by more recent PR.

PR #4110, opened 2019-06-06
[LATER] Docs: updated basic tutorial for CC7

antolu:
I guess take a look at this when the rest of the software is ready for CentOS 7. 

BEGINRELEASENOTES

*Docs
CHANGE: updated basic tutorial for CC7 instead of SLC6. 

ENDRELEASENOTES

### DIFF
--- a/docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
+++ b/docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
@@ -13,7 +13,7 @@ HOSTNAME=`hostname`
 
 # avoid calling passwd, which might be the cern passwd in /usr/sue/bin
 # START add_dirac
-adduser -s /bin/bash -d /home/dirac dirac
+adduser -s /bin/bash -d /home/dirac dirac || echo "User dirac already exists."
 echo password | /usr/bin/passwd --stdin dirac
 mkdir -p /opt/dirac/sbin
 chown -R dirac:dirac /opt/dirac/
@@ -62,7 +62,7 @@ systemctl start runsvdir-start
 ## SETUP FOR MYSQL
 # remove mysql
 # START mysqlInstall
-yum remove -y $(rpm -qa | grep -i mysql | paste -sd ' ')
+yum remove -y $(rpm -qa | grep -i -e mysql -e mariadb | paste -sd ' ') || echo  "MySQL is not yet installed"
 rm -rf /var/lib/mysql/*
 yum install -y \
     https://dev.mysql.com/get/Downloads/MySQL-5.7/mysql-community-devel-5.7.25-1.el7.x86_64.rpm \
@@ -73,7 +73,7 @@ yum install -y \
 # END mysqlInstall
 
 # START mysqlStart
-systemctl start mysqld
+systemctl start mysqld || systemctl restart mysqld
 # END mysqlStart
 
 cat > mysqlSetup.sql <<EOF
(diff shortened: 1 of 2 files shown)


andresailer:
You also have to change line 32 of the basicTutoSetup.rst file

Does the shell script also work when run twice on the same VM?

docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
```diff
@@ -48,10 +36,28 @@ exec chpst -u dirac $RUNSVDIR -P /opt/dirac/startup 'log:  DIRAC runsv'
 # END runsvdir-start
 EOF
 
+sed -iE "s@ExecStart=/.*@ExecStart=/opt/dirac/sbin/runsvdir-start@g" /usr/lib/systemd/system/runsvdir-start.service
```
andresailer: Why did you chose to just replace the content of one line, instead of replacing the whole file, which is what is shown to the user, which might lead to inconsistent behaviour between the shell script and what the user should get from the tutorial?

docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
```diff
@@ -22,22 +22,10 @@ chown -R dirac:dirac /opt/dirac/
...
-stop on runlevel [^2345]
-normal exit 0 111
-respawn
-exec /opt/dirac/sbin/runsvdir-start
-# END runsvdir.conf
-EOF
-
-
-cat > /opt/dirac/sbin/runsvdir-start <<'EOF'
+cat > /opt/dirac/sbin/runsvdir-start <<EOF
```
andresailer:
```suggestion
cat > /opt/dirac/sbin/runsvdir-start <<'EOF'
```

antolu:
> You also have to change line 32 of the basicTutoSetup.rst file
> 
> Does the shell script also work when run twice on the same VM?

Oh crap I missed that one when I redid the file on the second run. 

The script doesn't work on the first run, because the line that removes old mysql throws an error since no old mysql packages can be found (empty argument).

docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
antolu: I didn't think of that. But yes, that makes more sense

docs/source/AdministratorGuide/Tutorials/basicTutoSetup.sh
```diff
@@ -60,15 +66,15 @@ start runsvdir || restart runsvdir
 yum remove -y $(rpm -qa | grep -i mysql | paste -sd ' ')
```
andresailer:
```suggestion
 yum remove -y $(rpm -qa | grep -i -e mysql -e mariadb | paste -sd ' ') || echo "nothing to remove?"
```
This ignores errors in this line, and also removed mariadb packages, which is what I had to do on my cc7 VM

andresailer:
> The script doesn't work on the first run, because the line that removes old mysql throws an error since no old mysql packages can be found (empty argument).

I also forgot that the users already exist, so they need to be removed before running again.
or the error is explicitly ignored like
```
adduser .... dirac || echo "Ignore error, user already exists"
```

antolu:
> > The script doesn't work on the first run, because the line that removes old mysql throws an error since no old mysql packages can be found (empty argument).
> 
> I also forgot that the users already exist on the second run, so they need to be removed before running again, or the error is explicitly ignored like
> 
> ```
> adduser .... dirac || echo "Ignore error, user already exists"
> ```

There is another issue with the users. `/opt/dirac` is not removed, so the script to generate new certificates fail to overwrite them. In the same manner the client installation fails because the user directory still exists (and contains files).

antolu:
Script still fails when generating certificates. There is an error in the code that generates the certificates that makes it unable to overwrite existing keys. Also openssl throws error `ERROR:Serial number 1000 has already been issued, check the database/serial_file for corruption`, which supposedly has been fixed in a later version of openssl.

chaen:
You also get this error when you try regenerating the same certificate.

andresailer:
I think we can give up on re-running the script without manual intervention (removing users, folders).
As long as the script runs on a new VM, I would be satisfied

chaen:
So would I

PR #4112, opened 2019-06-11
[v6r21] Fix RequestTaskManager

phicharp:
BEGINRELEASENOTES
*TransformationSystem
FIX: Testing if the request ID is correct was not done properly, test the numerical value

ENDRELEASENOTES


phicharp:
To be clearer: bad `ExternalID` are `'0'` as a string which is not spotted by `if reqID`

PR #4120, opened 2019-06-13
[integration] Add ES monitoring support to service and agent.

pujanm:
This will add ES support to the DISET Service and will extend the monitoring system.


Core/DISET/private/Service.py
```diff
@@ -53,14 +55,20 @@ def __init__( self, serviceData ):
         Standalone is true if there is only one service started
         If it's false, every service is linked to a different MonitoringClient
     """
+    self.activityMonitoring = gConfig.getValue("/DIRAC/ActivityMonitoring") == "yes"
```
zmathe: we should be to accept Yes as well:
```python
self.activityMonitoring = gConfig.getValue("/DIRAC/ActivityMonitoring", "false").lower in ("yes","true")
```

Core/DISET/private/Service.py
```diff
@@ -236,40 +245,51 @@ def _loadActions( self ):
...
-        gLogger.error( "Missing property", prop[0] )
-        value = 'unset'
-      self._monitor.setComponentExtraParam( prop[1], value )
-    for secondaryName in self._cfg.registerAlsoAs():
-      gLogger.info( "Registering %s also as %s" % ( self._name, secondaryName ) )
-      self._validNames.append( secondaryName )
+    if self.activityMonitoring:
+      self.log["site"] = self._cfg.getHostname()
+      self.log["componentType"] = "service"
+      self.log["componentName"] = self._name
```
zmathe: you should add the ConponentLocation as well: 
```python
self._cfg.getURL()
```

MonitoringSystem/Client/Types/ComponentsActivityMonitoring.py
```diff
@@ -0,0 +1,42 @@
...
+
+  def __init__(self):
+    super(ComponentsActivityMonitoring, self).__init__()
+
+    """ c'tor
+
+    :param self: self reference
+    """
+
+    self.keyFields = ["site", "componentType", "componentName"]
```
zmathe: add componentLocaction as well

tests/Integration/Monitoring/Test_MonitoringReporter.py
```diff
@@ -250,6 +251,12 @@ def test_addComponentRecords(self):
     self.assertTrue(result['OK'])
     self.assertEqual(result['Value'], len(self.data))
 
+  def test_addComponentsActivityRecords(self):
```
zmathe: You should add different data. self.data is containing WMSMonitoring informations.
Please create a new dictionary. In addition it is probably better to create  a new test (TestComponentsActivityMonitoringTestCase), because you will have to add more test...

zmathe:
@pujanm please check why tarvis is failing

pujanm:
@zmathe I will do the changes
And travis is failing for these two reasons:
1. `TransformationSystem/Agent/test/Test_Plugins.py` (No changes made by me here.)
2. Some lines are too long and there are whitespaces between `[ ]` and `( )` in the code of 
    `Core/DISET/private/Service.py`

And one more doubt why the changes I made are not being reflected like the data is not being gathered in the ES backend. What might be the reason?

zmathe:
you should rebase your branch.
You should write the test case (integration test) with the correct data in order to make sure that the new monitoring type is correctly defined. After you can debug what is the problem in the Service.py file.

pujanm:
The tests have passed. So would there be issues in `Service.py` ?

pujanm:
@zmathe Should I improve the code of `TransformationSystem/Agent/test/Test_Plugins.py` to `py3k` standards?

zmathe:
You have not modified this file and you do not need to improve. You can try to rebase, most of the case it fixes the problem

pujanm:
I tried rebasing but again one check is failing because of this file `TransformationSystem/Agent/test/Test_Plugins.py`.

fstagni:
Fix `Test_Plugins.py` too please

Core/DISET/private/Service.py
```diff
@@ -60,7 +61,16 @@ def __init__( self, serviceData ):
     if serviceData[ 'loadName' ] not in self._validNames:
       self._validNames.append( serviceData[ 'loadName' ] )
     self._cfg = ServiceConfiguration( list( self._validNames ) )
-    if serviceData[ 'standalone' ]:
+    # ElasticDB
+    self.activityMonitoring = gConfig.getValue("/DIRAC/ActivityMonitoring", "false").lower() in ("yes","true")
```
chaen: You cannot have this logic in the `__init__` because the remote configuration is not loaded yet. You have to do all this in the `initialize` method

pujanm: Sure, I will be making changes as we discussed yesterday and will push all the changes for further review.

Core/DISET/private/Service.py
```diff
@@ -60,7 +61,16 @@ def __init__( self, serviceData ):
     if serviceData[ 'loadName' ] not in self._validNames:
       self._validNames.append( serviceData[ 'loadName' ] )
     self._cfg = ServiceConfiguration( list( self._validNames ) )
-    if serviceData[ 'standalone' ]:
+    # ElasticDB
+    self.activityMonitoring = gConfig.getValue("/DIRAC/ActivityMonitoring", "false").lower() in ("yes","true") 
+    gLogger.info("The value fo activitMonitoring flag is " + str(self.activityMonitoring))
+    if self.activityMonitoring:
+      self.activityMonitoringReporter = MonitoringReporter(monitoringType="ComponentsActivityMonitoring")
+      self.log = {}
```
chaen: I am assuming this is a dictionary which contains the entries you want to send to the monitoring ? `log` is not really an appropriate name :)
Moreover, `self.log` is a instance attribute which is modified in multiple threads if I understand correctly -> not thread safe

pujanm: So how does this look if we transfer all the ES monitoring logic to a method named `_activityMonitoring` and inside that we have a dictionary which is only accessible within this method and we call it periodically with `gThreadScheduler.addPeriodicTask(300, _activityMonitoring)` this way ? 
And can you please suggest a name instead of `log` ?

Core/DISET/private/Service.py
```diff
@@ -110,7 +121,13 @@ def initialize( self ):
                             }
     #Call static initialization function
     try:
-      self._handler[ 'class' ]._rh__initializeClass( dict( self._serviceInfoDict ),
+      if self.activityMonitoring:
+        self._handler[ 'class' ]._rh__initializeClass( dict( self._serviceInfoDict ),
+                                                     self._lockManager,
+                                                     self._msgBroker,
+                                                     self.log )
```
chaen: Did you check how this last parameter is used in the `Requesthandler` class ? since your `self.log` and `self._monitor` do not have the same interface, I doubt the `RequestHandler` will be happy

Core/DISET/private/Service.py
```diff
@@ -577,4 +632,16 @@ def __endReportToMonitoring( self, initialWallTime, initialCPUTime ):
...
     percentage = cpuTime / wallTime * 100.
     if percentage > 0:
-      self._monitor.addMark( 'CPU', percentage )
+      if self.activityMonitoring:
+        # ElasticDB
+        self.log["CPU"] = percentage
+        gLogger.info("Added CPU")
+        self.log["timestamp"] = time.time()
+
+        if "RunningThreads" in self.log:
```
chaen: I do not understand why do you only add the record if `RunningThreads` is in it ?

pujanm: I observed that there were many logs of `Configuration Server` which were being added without `ThreadPoolContents` so that's why I did this.

Core/DISET/private/Service.py
pujanm: Yes I check the method inside `RequestHandler` it was being used here `srv_getMonitor` just to return `cls.__monitor` so didn't understand what to pass there as we are not initializing `self._monitor` when `activityMonitoring` flag is on.

Core/DISET/private/Service.py
chaen: For the name, something more describing, I don't know, like `monitoringData` or, `monitoringRecord` or anything like that. 

I am not sure I understand your logic of period task. For example, we want to increase the query counter everytime there is a connection. How would you do that with a periodic task ?

Core/DISET/private/Service.py
chaen: Sorry I do not understand what you mean. 
If there are logs (and they are not *logs* ) without the `ThreadPoolContents`, there must be a reason. But probably not a reason to ignore these records. And of course, if you check how this `ThreadPoolContents` is filled, you'll see that it's a period task every 30 seconds

Core/DISET/private/Service.py
chaen: Before we were sending the `MonitoringClient` object, so I assume that it would make more sense to send the `MonitoringReporter` object, don't you think ? 
Also, the `RequestHandler` is inherited by all DIRAC services, so it might be used there

Core/DISET/private/Service.py
pujanm: So can we update it with the help of this `self._stats[ 'connections' ]` which is accessible globally?
With the help of which we can update both connections and query counter.

pujanm:
I guess there is some issue with this file: `DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py` that's why Travis build is failing.

fstagni:
Fixed in https://github.com/DIRACGrid/DIRAC/pull/4153 (will merge just after review OK)

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'threads',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotRunningThreads(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: please do not use /opt/dirac/ I already said you can install DIRAC to different directory ...

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'percentage',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotCpuUsage(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same here

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'MB',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotMemoryUsage(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same here

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'seconds',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotRunningTime(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same here

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'Connections',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotConnections(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same here

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'ActiveQueries',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotActiveQueries(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'PendingQueries',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotPendingQueries(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'MaxFD',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotMaxFD(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,404 @@
...
+                         },
+                'unit': 'RunningThreads',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotActivityRunningThreads(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/opt/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
zmathe: same

tests/Integration/Monitoring/Test_Plotter.py
pujanm: Yes I changed that but there was a travis issue which has to be merged so will push the changes.

pujanm:
@zmathe I did the changes. I guess the travis issues are fixed in #4163.

FrameworkSystem/Client/MonitoringClient.py
```diff
@@ -3,6 +3,8 @@
     Uses RPC Framework/Monitoring service. Calls registerActivities exposed function
 """
 
+from __future__ import absolute_import
+from six.moves import range
```
chaen: why ? `xrange` does the trick no ?

FrameworkSystem/Client/MonitoringClient.py
```diff
@@ -90,88 +121,96 @@ def __init__( self ):
...
-    if name in ( 'version', 'DIRACVersion', 'description', 'startTime', 'platform' ):
-      self.__compRegistrationExtraDict[ name ] = str( value )
-    elif name in ( 'cycles', 'queries' ):
-      self.__compCommitExtraDict[ name ] = str( value )
+  def setComponentExtraParam(self, name, value):
+    """
+    Sets the extra parameters of the component reporting.
+
+    :type  name: string
+    :param name: It should belong to one of these 'version', 'DIRACVersion', 'description', 'startTime', 'platform'.
```
chaen: what about `cycles` and `queries` ?

FrameworkSystem/Client/MonitoringClient.py
```diff
@@ -224,64 +263,68 @@ def addMark( self, name, value = 1 ):
...
+            totalValue /= len(consolidatedMarks[key][markTime])
+          consolidatedMarks[key][markTime] = totalValue
+      if len(consolidatedMarks[key]) == 0:
+        del(consolidatedMarks[key])
     self.activitiesMarks = remainderMarks
     return consolidatedMarks
 
-  def flush( self, allData = False ):
+  def flush(self, allData=False):
+    """ This is the method which is called by the gMonitoringFlusher object in order to flush all the data
```
chaen: That's a bit too detailed. You do not need to explain the internals of each and every object that is used inside the function. 
You need to explain what it does, and what are the params. 

Inside the code itself, if needs to be, you can detail what is going on

FrameworkSystem/private/monitoring/ServiceInterface.py
```diff
@@ -18,6 +18,15 @@
 from DIRAC.FrameworkSystem.private.monitoring.MonitoringCatalog import MonitoringCatalog
 
 class ServiceInterface( object ):
+  """
+  The gServiceInterface singleton object which is used by the MonitoringHandle to register activities and
+  commit the data is inhertied from this class.
+  When the gMonitor object calls the registerActivities method through the rpcClient at that time the ServiceInterface
```
chaen: again, too verbose.

FrameworkSystem/Client/MonitoringClient.py
pujanm: Nope, I guess this was done by `python-modernize`. I will remove it.

FrameworkSystem/Client/MonitoringClient.py
```diff
@@ -68,12 +63,18 @@ def __init__(self):
     time.sleep(0.1)
 
   def flush(self, allData=False):
-    """ This flush method calls the flush method from the gMonitor object.
+    """ This method is used to periodically flush the data
+        :type allData: bool
```
chaen: You need a line between the space and the params (true everywhere, I won't repeat it)

Core/DISET/private/Service.py
```diff
@@ -599,7 +598,17 @@ def _mbConnect(self, trid, handlerObj=None):
 
   def _executeAction(self, trid, proposalTuple, handlerObj):
     try:
-      return handlerObj._rh_executeAction(proposalTuple)
+      response = handlerObj._rh_executeAction(proposalTuple)
```
zmathe: I am not sure it will work as it is now. response can be S_ERROR and you will not have response[1]

Core/DISET/private/Service.py
```diff
@@ -614,7 +623,17 @@ def _mbReceivedMsg(self, trid, msgObj):
     if not result['OK']:
       return result
     handlerObj = result['Value']
-    return handlerObj._rh_executeMessageCallback(msgObj)
+    response = handlerObj._rh_executeMessageCallback(msgObj)
```
zmathe: here as well in case of error only response[0] is exists

Core/DISET/private/Service.py
```diff
@@ -599,7 +598,17 @@ def _mbConnect(self, trid, handlerObj=None):
 
   def _executeAction(self, trid, proposalTuple, handlerObj):
     try:
-      return handlerObj._rh_executeAction(proposalTuple)
+      response = handlerObj._rh_executeAction(proposalTuple)
+      if self.activityMonitoring:
```
zmathe: so:
```python
if self.activityMonitoring and response[1]:
```

Core/DISET/RequestHandler.py
```diff
@@ -134,10 +134,11 @@ def _rh_executeAction(self, proposalTuple):
       message = "Method %s for action %s does not return a S_OK/S_ERROR!" % (actionTuple[1], actionTuple[0])
       gLogger.error(message)
       retVal = S_ERROR(message)
-    self.__logRemoteQueryResponse(retVal, time.time() - startTime)
+    elapsedTime = time.time() - startTime
+    self.__logRemoteQueryResponse(retVal, elapsedTime)
     result = self.__trPool.send(self.__trid, retVal)  # this will delete the value from the S_OK(value)
     del retVal
-    return result
+    return result, elapsedTime
```
chaen: You can't do that. The return has to be `S_OK` or `S_ERROR`

Core/DISET/private/Service.py
```diff
@@ -317,264 +385,293 @@ def _processInThread( self, clientTransport ):
...
+    return handlerObj._rh_executeConnectionCallback('connected')
 
-  def _executeAction( self, trid, proposalTuple, handlerObj ):
+  def _executeAction(self, trid, proposalTuple, handlerObj):
     try:
-      return handlerObj._rh_executeAction( proposalTuple )
+      response = handlerObj._rh_executeAction(proposalTuple)
+      if self.activityMonitoring and len(response) == 2:
+        self.activityMonitoringReporter.addRecord({
+            'timestamp': time.time(),
```
chaen: We use `utctime`

FrameworkSystem/Client/MonitoringClient.py
andresailer: python3 doesn't have `xrange`

chaen: YesI know, but `2to3` does the migration nicely

pujanm:
Okay thanks will work on these changes.

Core/Base/AgentModule.py
```diff
@@ -22,6 +23,8 @@
 from DIRAC.FrameworkSystem.Client.MonitoringClient import gMonitor
 from DIRAC.ConfigurationSystem.Client import PathFinder
 from DIRAC.FrameworkSystem.Client.MonitoringClient import MonitoringClient
+from DIRAC.Core.Utilities.ThreadScheduler import gThreadScheduler
```
chaen: Actually we never used a gThreadScheduler in an Agent as far as I know. @fstagni @atsareg any reason why this would not work ?

Core/Base/AgentModule.py
```diff
@@ -381,31 +389,44 @@ def am_go(self):
     self.log.notice(" Average execution/polling time: %.2f%%" % elapsedPollingRate)
     if cycleResult['OK']:
       self.log.notice(" Cycle was successful")
+      if self.activityMonitoring:
+        self.activityMonitoringReporter.addRecord({
+            'timestamp': datetime.datetime.utcnow(),
```
chaen: I am wondering whether I did not tell you something wrong before. Please wait for @zmathe comment as whether the timestamp should be epoch or utc

zmathe: We store epoch in ES, but the Monitoring system accept epoch(int) time and utc datetime as well. If the timestamp is datetime, it will be converted to epoch. For performance reason I prefer epoch, because it will be easier to serialize and send through the network ...

pujanm: Okay so I will go with epoch.

Core/DISET/private/Service.py
```diff
@@ -317,264 +385,298 @@ def _processInThread( self, clientTransport ):
...
+      handlerObj = result['Value']
+    return handlerObj._rh_executeConnectionCallback('connected')
 
-  def _executeAction( self, trid, proposalTuple, handlerObj ):
+  def _executeAction(self, trid, proposalTuple, handlerObj):
     try:
-      return handlerObj._rh_executeAction( proposalTuple )
+      response = handlerObj._rh_executeAction(proposalTuple)
+      if self.activityMonitoring and response["OK"]:
+        self.activityMonitoringReporter.addRecord({
```
chaen: Here, you are back in a thread, which we already concluded is not really thread safe. 
Not to waste more time, I'll ask @zmathe to make it thread safe (it should be quite straight forward)

tests/Integration/Monitoring/Test_Plotter.py
```diff
@@ -0,0 +1,408 @@
...
+                         },
+                'unit': 'seconds',
+                'granularity': 900}
+
+    obj = ComponentMonitoringPlotter(None, None)
+    res = obj._plotRunningTime(reportRequest, plotInfo, plotName)
+    self.assertEqual(res['OK'], True)
+    self.assertEqual(res['Value'], {'plot': True, 'thumbnail': False})
+
+    res = compare('%s.png' % plotName, '/home/dirac/DIRAC/tests/Integration/Monitoring/png/%s.png' % plotName)
```
chaen: I am pretty sure @zmathe told you multiple times not to hardcode path like that

pujanm: Like we had discussed on slack he asked me to keep the path as `/home/dirac`.

chaen: No, he told you that you cannot have hardcoded path, because it can be anywhere, for example `/home/dirac` instead of `/opt/dirac`

zmathe: @pujanm I told you many times do not hardcode  the path. You can see in the screen shoot:
Screenshot from 2019-07-31 14-54-51

pujanm: Okay sorry @zmathe @chaen will do the change.

dirac.cfg
```diff
@@ -357,6 +357,8 @@ Operations
   #Any value here can be overwriten in the setup specific section
   Defaults
   {
+    # This will globally enable ES based monitoring for Service and AgentModule.
+    EnableActivityMonitoring = yes
```
chaen: Also add one at the level of a service, for the example

Core/Base/AgentModule.py
```diff
@@ -390,8 +392,10 @@ def am_go(self):
     if cycleResult['OK']:
       self.log.notice(" Cycle was successful")
       if self.activityMonitoring:
+        # Here we record the data about the cycle duration along with some basic details about the
+        # component and right now it isn't committed to the ES backend.
         self.activityMonitoringReporter.addRecord({
-            'timestamp': datetime.datetime.utcnow(),
+            'timestamp': time.time(),
```
zmathe: this is not UTC epoch

Core/DISET/private/Service.py
```diff
@@ -313,7 +312,7 @@ def __reportThreadPoolContents(self):
     if self.activityMonitoring:
       # As ES accepts raw data these monitoring fields are being sent here because they are time dependant.
       self.activityMonitoringReporter.addRecord({
-          'timestamp': datetime.datetime.utcnow(),
+          'timestamp': time.time(),
```
zmathe: here as well

Core/DISET/private/Service.py
```diff
@@ -347,7 +346,7 @@ def handleConnection(self, clientTransport):
     if self.activityMonitoring:
       #  As ES accepts raw data these monitoring fields are being sent here because they are action dependant.
       self.activityMonitoringReporter.addRecord({
-          'timestamp': datetime.datetime.utcnow(),
+          'timestamp': time.time(),
```
zmathe: here as well

pujanm:
@zmathe Can you approve this as one code owner review will be required I think to approve this PR.

chaen:
Why wasn't that one merged yet ?

fstagni:
I'll merge this one tomorrow if there are no objections.

PR #4129, opened 2019-06-20
[v6r21] Freeze typing to 3.6.6 because it causes an exception in hypothesis

andresailer:
Fix for tests:

maybe review merge and propagate?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,6 +37,8 @@ sqlalchemy>=1.0.9
 stomp.py>=3.1.5
 suds-jurko>=0.6
 sphinx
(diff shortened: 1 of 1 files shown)


chaen:
Maybe just add a comment in the requirements.txt such that we know in the future why the version is fixed ? Otherwise I'm all up for merging and propagating straight away. Can you do the same PR agaisnt DIRACOS please ?

andresailer:
When I wanted to make the PR against diracOS this morning, github was not working.
I already made a test on gitlab: https://gitlab.cern.ch/CLICdp/iLCDirac/diracos-test/-/jobs/4572275

PR #4144, opened 2019-07-01
[v6r21] a logic bug in counting numberOfTasks

zhangxiaomei:
MCExtension has a logic bug in the code when counting "numberOfTasks = maxTasks - ( total - failed )", total should be "totalcreated" instead of just "created"
original: total = statusDict.get( 'Created', 0 )
modified: total = statusDict.get( 'TotalCreated', 0 )
After changed, the "MaxNumberOfTasks" can correctly limit the number of jobs 

created.

BEGINRELEASENOTES

*Transformation
FIX: fixed a logic bug in counting numberOfTasks in MCExtension which is expected 

to limit the total number of tasks for MC transformations

ENDRELEASENOTES

### DIFF
--- a/TransformationSystem/Agent/MCExtensionAgent.py
+++ b/TransformationSystem/Agent/MCExtensionAgent.py
@@ -93,7 +93,7 @@ class MCExtensionAgent( AgentModule ):
     done = statusDict.get( 'Done', 0 )
     failed = statusDict.get( 'Failed', 0 )
     waiting = statusDict.get( 'Waiting', 0 )
-    total = statusDict.get( 'Created', 0 )
+    total = statusDict.get( 'TotalCreated', 0 )
     # If the failure rate is higher than acceptable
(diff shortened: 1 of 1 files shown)


TransformationSystem/Agent/MCExtensionAgent.py
```diff
@@ -93,7 +93,7 @@ def _calculateTaskNumber( self, maxTasks, statusDict ):
     done = statusDict.get( 'Done', 0 )
     failed = statusDict.get( 'Failed', 0 )
     waiting = statusDict.get( 'Waiting', 0 )
-    total = statusDict.get( 'Created', 0 )
+    total = statusDict.get( 'TotalCreated', 0 )
```
andresailer:
```suggestion
    total = statusDict.get('TotalCreated', 0)
```
Correct formatting

PR #4152, opened 2019-07-04
[v6r21] codecov configuration

andresailer:
This sets the target coverage diff for the "patch" check to 0 and adds a threshold of 0.03 for the "project" coverage to allow random fluctuations in the diff (caused by calls to "random" in functions that are tested?).

(Also fixes a syntax error in the yaml compared to that of the integration branch.)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,10 +1,9 @@
 comment: off
 
 ignore:
-  - ".*/[Tt]ests/.*"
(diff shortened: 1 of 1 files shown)


andresailer:
Should all tests be ignored (:+1:) or should they be kept (:-1:) when calculating coverage?
Vote now:

PR #4181, opened 2019-07-25
[v6r22] allow to instans an extension of diracos

zmathe:
BEGINRELEASENOTES

*Core
NEW: Allow to install an extension of DIRACOS
ENDRELEASENOTES

### DIFF
--- a/Core/scripts/dirac-install.py
+++ b/Core/scripts/dirac-install.py
@@ -72,6 +72,12 @@ After providing the default configuration files, DIRAC or your extension can be 
 
    It will install DIRAC v6r20-pre16
 
+   You can install an extension of diracos.
+
+   for example::
+
+     dirac-install -r v9r4-pre2 -l LHCb --dirac-os --dirac-os-version=LHCb:master
+
 3. You have possibility to install a not-yet-released DIRAC, module or extension using -m or --tag options.
    The non release version can be specified.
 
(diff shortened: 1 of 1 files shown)


Core/scripts/dirac-install.py
```diff
@@ -1113,18 +1113,18 @@ def getExtenalsVersion(self, release=None):
 
   def getDiracOSExtensionAndVersion(self, diracOSVersion):
     """
-    This method return the diracos and version taking into 
+    This method return the diracos and version taking into
```
chaen: Maybe just explain that the tar file looked for will be `<Extension<diracos-<version>`. And you may want to add that option at the very top of the file, where you give the examples ?

zmathe: right.

Core/scripts/dirac-install.py
```diff
@@ -72,6 +72,12 @@
 
    It will install DIRAC v6r20-pre16
 
+   You can install an extension of diracos.
+
+   for example::
+
+     dirac-install -r v9r4-pre2 -l LHCb --dirac-os --dirac-os-version=LHCb:master
```
atsareg: I guess that the example should read (no equals sign):

  dirac-install -r v9r4-pre2 -l LHCb --dirac-os --dirac-os-version LHCb:master

PR #4206, opened 2019-08-07
[v6r22] improved treatment for the case without SpaceToken in SpaceOccupancy

iueda:
BEGINRELEASENOTES

*Resources/Storage
CHANGE: improved treatment for the case without SpaceToken in SpaceOccupancy 

ENDRELEASENOTES


chaen:
Fine with me, please just check the formating error

PR #4209, opened 2019-08-08
[integration] Add ES Monitoring support to RequestManagementSystem and DataManagementSystem/Agent/RequestOperations/

pujanm:
BEGINRELEASENOTES

NEW: Extends the MonitoringSystem to support RMS and DMS/Agents/RequestOperations.

ENDRELEASENOTES


DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,325 @@
...
-    gMonitor.registerActivity( "RemoveReplicaAtt", "Replica removals attempted",
-                               "RequestExecutingAgent", "Files/min", gMonitor.OP_SUM )
-    gMonitor.registerActivity( "RemoveReplicaOK", "Successful replica removals",
-                               "RequestExecutingAgent", "Files/min", gMonitor.OP_SUM )
-    gMonitor.registerActivity( "RemoveReplicaFail", "Failed replica removals",
-                               "RequestExecutingAgent", "Files/min", gMonitor.OP_SUM )
+    super(MoveReplica, self).__init__(operation, csPath)
+
+    # Check whether the ES flag is enabled so we can send the data accordingly.
+    self.rmsMonitoring = Operations().getValue("EnableActivityMonitoring", False)
```
zmathe: This does not reflect if we this flag is True but we set it False in the agent option. But you do not have access to the agent options.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,325 @@
...
     if sourceSE:
       # # check source se for read
-      bannedSource = self.checkSEsRSS( sourceSE, 'ReadAccess' )
+      bannedSource = self.checkSEsRSS(sourceSE, 'ReadAccess')
       if not bannedSource["OK"]:
-        gMonitor.addMark( "ReplicateAndRegisterAtt", len( self.operation ) )
-        gMonitor.addMark( "ReplicateFail", len( self.operation ) )
+        if self.rmsMonitoring:
+          for opFile in self.operation:
+            for status in ["FileAttempted", "FileFailed"]:
```
zmathe: I do not get what you want to do here? You set all files FileAttempted and FileFailed This is what you want?

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,325 @@
...
+        return S_OK(self.operation.Error)
 
     # # check targetSEs for write
     bannedTargets = self.checkSEsRSS()
     if not bannedTargets['OK']:
-      gMonitor.addMark( "ReplicateAndRegisterAtt", len( self.operation ) )
-      gMonitor.addMark( "ReplicateFail", len( self.operation ) )
+      if self.rmsMonitoring:
+        for opFile in self.operation:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: here as well.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,325 @@
...
     # # for removal the targetSEs are the sourceSEs of the replication
     targetSEs = self.operation.sourceSEList
-    bannedTargets = self.checkSEsRSS( targetSEs, access = 'RemoveAccess' )
+    bannedTargets = self.checkSEsRSS(targetSEs, access='RemoveAccess')
     if not bannedTargets['OK']:
-      gMonitor.addMark( "RemoveReplicaAtt" )
-      gMonitor.addMark( "RemoveReplicaFail" )
+      if self.rmsMonitoring:
+        for opFile in self.operation:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: same here as well

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -247,122 +388,173 @@ def dmTransfer( self, opFile ):
...
+    if opFile.Error:
+      if self.rmsMonitoring:
+        self.rmsMonitoringReporter.addRecord({
+            "timestamp": time.mktime(datetime.datetime.utcnow().timetuple()),
+            "host": socket.getfqdn(),
+            "objectType": "File",
+            "operationType": self.operation.Type,
+            "objectID": opFile.FileID,
+            "parentID": self.operation.OperationID,
+            "status": "FileFailed",
```
zmathe: I think the "status": "ReplicateFailed" and this is true for other places. @chaen 
What do you think? My opinion we should be able do distinguish between different activities.

DataManagementSystem/Agent/RequestOperations/PhysicalRemoval.py
```diff
@@ -27,121 +27,207 @@
...
+  def __call__(self):
     """ perform physical removal operation """
-    bannedTargets = self.checkSEsRSS( access = 'RemoveAccess' )
+    bannedTargets = self.checkSEsRSS(access='RemoveAccess')
     if not bannedTargets['OK']:
-      gMonitor.addMark( "PhysicalRemovalAtt" )
-      gMonitor.addMark( "PhysicalRemovalFail" )
+      if self.rmsMonitoring:
+        for opFile in self.operation:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: same here as well

DataManagementSystem/Agent/RequestOperations/PutAndRegister.py
```diff
@@ -43,57 +51,93 @@ class PutAndRegister( DMSRequestOperationsBase ):
...
 
         opFile.Status = "Failed"
         opFile.Error = "Wrong parameters: TargetSE should contain only one targetSE"
 
-        gMonitor.addMark( "PutAtt", 1 )
-        gMonitor.addMark( "PutFail", 1 )
-
-      return S_ERROR( "TargetSE should contain only one target, got %s" % targetSEs )
+        if self.rmsMonitoring:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: here as well.

DataManagementSystem/Agent/RequestOperations/PutAndRegister.py
```diff
@@ -43,57 +51,93 @@ class PutAndRegister( DMSRequestOperationsBase ):
...
 
     targetSE = targetSEs[0]
-    bannedTargets = self.checkSEsRSS( targetSE )
+    bannedTargets = self.checkSEsRSS(targetSE)
     if not bannedTargets['OK']:
-      gMonitor.addMark( "PutAtt" )
-      gMonitor.addMark( "PutFail" )
+      if self.rmsMonitoring:
+        for opFile in self.operation:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: here as well

DataManagementSystem/Agent/RequestOperations/ReTransfer.py
```diff
@@ -47,67 +54,152 @@ class ReTransfer( DMSRequestOperationsBase ):
...
     # # check targetSEs for removal
     targetSE = targetSEs[0]
-    bannedTargets = self.checkSEsRSS( targetSE )
+    bannedTargets = self.checkSEsRSS(targetSE)
     if not bannedTargets['OK']:
-      gMonitor.addMark( "FileReTransferAtt" )
-      gMonitor.addMark( "FileReTransferFail" )
+      if self.rmsMonitoring:
+        for opFile in self.operation:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: here as well

DataManagementSystem/Agent/RequestOperations/RemoveFile.py
```diff
@@ -29,124 +29,239 @@
...
+    fc = FileCatalog(self.operation.catalogList)
 
-    res = fc.getReplicas( [wf.LFN for wf in waitingFiles] )
+    res = fc.getReplicas([wf.LFN for wf in waitingFiles])
     if not res['OK']:
-      gMonitor.addMark( "RemoveFileAtt" )
-      gMonitor.addMark( "RemoveFileFail" )
+      if self.rmsMonitoring:
+        for opFile in waitingFiles:
+          for status in ["FileAttempted", "FileFailed"]:
```
zmathe: here as well

RequestManagementSystem/Agent/RequestExecutingAgent.py
```diff
@@ -155,13 +160,21 @@ def __init__(self, *args, **kwargs):
...
-                              "RequestExecutingAgent", "Loops/min", gMonitor.OP_SUM)
-    gMonitor.registerActivity("Processed", "Request Processed",
-                              "RequestExecutingAgent", "Requests/min", gMonitor.OP_SUM)
-    gMonitor.registerActivity("Done", "Request Completed",
-                              "RequestExecutingAgent", "Requests/min", gMonitor.OP_SUM)
+    # Check whether the ES flag is enabled so we can send the data accordingly.
+    self.rmsMonitoring = Operations().getValue("EnableActivityMonitoring", False)
+
+    if self.rmsMonitoring:
+      self.rmsMonitoringReporter = MonitoringReporter(monitoringType="RMSMonitoring")
```
zmathe: we should be able to disable monitoring in the agent option

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: Like I did this similar to `gMonitor` object as there the source is banned so we send status Attempted and Failed both as from here nothing will move further.

pujanm: @zmathe can I resolve this if we are clear on it?

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: I think we won't be able to access the service or agent flag from here as they are not instances of any of the two so that's why I choose to use only Operations here but we can do it for `RequestExecutingAgent` as its an agent instance. I will think about this then too.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
zmathe: yes, you can resolve all of them

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: @zmathe can we pass this flag from RequestExecutingAgent to the `RequestOperations` this way we will be able to keep the flag configurable from both Operations and Agent?

pujanm: Solved this problem.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: I think we can do this with the help of our plotter where we select first the `Operation Type` and then the `Object Type` i.e. `Operation or File`.

pujanm: For example this plot:
moveReplicaFileStatus

This is particular to the `MoveReplica` operation and for object type `File`.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
     sourceSE = self.operation.SourceSE if self.operation.SourceSE else None
     if sourceSE:
       # # check source se for read
-      bannedSource = self.checkSEsRSS( sourceSE, 'ReadAccess' )
+      bannedSource = self.checkSEsRSS(sourceSE, 'ReadAccess')
       if not bannedSource["OK"]:
-        gMonitor.addMark( "ReplicateAndRegisterAtt", len( self.operation ) )
-        gMonitor.addMark( "ReplicateFail", len( self.operation ) )
+        if self.rmsMonitoring:
+          for opFile in self.operation:
```
chaen: I do not like these status. The status is not `FileFailed` or `FileAttempted`. The status is `Attempted` or `Failed`. You have the `objectType` to specify `File`

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
+            for status in ["FileAttempted", "FileFailed"]:
+              self.rmsMonitoringReporter.addRecord({
+                  "timestamp": time.mktime(datetime.datetime.utcnow().timetuple()),
```
chaen: According to the various docs (`time` and `datetime`, this does not give utc)

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
+                  "host": socket.getfqdn(),
```
chaen: Please favour the DIRAC module `Network.getFQDN` for consistency

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
+                  "objectType": "File",
+                  "operationType": self.operation.Type,
+                  "objectID": opFile.FileID,
+                  "parentID": self.operation.OperationID,
+                  "status": status,
+                  "nbObject": 1
```
chaen: I think the idea for File was to not use the `objectID`, thus allowing you to send only one report with `nbObject` being the total number of files. @zmathe can you confirm ?

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
-        return S_OK( self.operation.Error )
+        self.log.info(self.operation.Error)
+        return S_OK(self.operation.Error)
 
     # # check targetSEs for write
     bannedTargets = self.checkSEsRSS()
     if not bannedTargets['OK']:
-      gMonitor.addMark( "ReplicateAndRegisterAtt", len( self.operation ) )
-      gMonitor.addMark( "ReplicateFail", len( self.operation ) )
+      if self.rmsMonitoring:
```
chaen: all the same remarks as before

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -7,201 +7,324 @@
...
 
     # # check sourceSEs for removal
     # # for removal the targetSEs are the sourceSEs of the replication
     targetSEs = self.operation.sourceSEList
-    bannedTargets = self.checkSEsRSS( targetSEs, access = 'RemoveAccess' )
+    bannedTargets = self.checkSEsRSS(targetSEs, access='RemoveAccess')
     if not bannedTargets['OK']:
-      gMonitor.addMark( "RemoveReplicaAtt" )
-      gMonitor.addMark( "RemoveReplicaFail" )
+      if self.rmsMonitoring:
```
chaen: same as before ( I won't repeat, I let you find and check the other occurrences)

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
chaen: No, we decided that the status is `Failed`, full stop. The type of activiy is already reflected in the Data Management accounting

RequestManagementSystem/Client/Request.py
```diff
@@ -82,6 +82,7 @@ def __init__( self, fromDict = None ):
     self.RequestName = None
     self.OwnerGroup = None
     self.SourceComponent = None
+    self.RequestID = None
```
chaen: No no no no. That will break everything. Read the comment: https://github.com/DIRACGrid/DIRAC/blob/integration/RequestManagementSystem/Client/Request.py#L50

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: Okay, noted. I will work on these changes. And will change at every similar occurrence.

RequestManagementSystem/Client/Request.py
pujanm: Sorry, I did this because Travis was failing will remove it.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
pujanm: I had verified this and it gives timestamp in accordance with UTC time.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
zmathe: yes, right. We want to send one report with total number of files

RequestManagementSystem/Client/Request.py
chaen: Travis can't be failing because of that, there must be something else

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
chaen: I am not sure 
https://docs.python.org/2/library/datetime.html#datetime.datetime.timetuple
https://docs.python.org/2/library/datetime.html#datetime.datetime.utctimetuple
https://docs.python.org/2/library/time.html#time.gmtime


```
In [9]: map(int, (time.mktime(datetime.datetime.utcnow().utctimetuple()), time.mktime(datetime.datetime.utcnow().timetuple()), time.mktime(datetime.datetime.now().timetuple()), time.mktime(d
   ...: atetime.datetime.now().utctimetuple()),  time.mktime(time.gmtime())))
Out[9]: [1566199416, 1566195816, 1566203016, 1566206616, 1566199416]
```

chaen: the trick is often the day light saving factor

zmathe: Maybe better to use the time method from DIRAC?
```
from DIRAC.Core.Utilities import Time
Time.toEpoch()
```
or please check how it is done in that module...

pujanm: Okay will check this.

pujanm: @chaen @zmathe I checked this `Time.toEpoch()` it uses this `datetime.datetime(2000, 1, 1).utcnow().timetuple()` and I think in our case this will fit `datetime.datetime(2000, 1, 1).utcnow().utctimetuple()` because in this case `tm_isdst=0` so it confirms that the time is in `UTC` otherwise if we use `timetuple` then `tm_isdst=-1` so it is assumed that it is an unkown factor.
So should we go with this `datetime.datetime(2000, 1, 1).utcnow().utctimetuple()`?

chaen: Hum, tricky. But Zoltan is right, use `Time.toEpoch()`. maybe it is not correct (maybe it is), but at least it is consistent everywhere

pujanm: Okay fine I will go with `Time.toEpoch()`.

DataManagementSystem/Agent/RequestOperations/PhysicalRemoval.py
```diff
@@ -30,118 +30,194 @@
...
+
+    if self.rmsMonitoring:
+      self.rmsMonitoringReporter.addRecord({
+          "timestamp": int(Time.toEpoch()),
+          "host": Network.getFQDN(),
+          "objectType": "File",
+          "operationType": self.operation.Type,
+          "parentID": self.operation.OperationID,
+          "status": "Attempted",
+          "nbObject": len(toRemoveDict) * len(targetSEs)
```
chaen: targetSEs may be None. Anyway, don't bother with that, we are interested only in the number of File objects to treat

DataManagementSystem/Agent/RequestOperations/PhysicalRemoval.py
```diff
@@ -30,118 +30,194 @@
...
+
+        if self.rmsMonitoring:
+          self.rmsMonitoringReporter.addRecord({
+              "timestamp": int(Time.toEpoch()),
+              "host": Network.getFQDN(),
+              "objectType": "File",
+              "operationType": self.operation.Type,
+              "parentID": self.operation.OperationID,
+              "status": "Successful",
+              "nbObject": len(targetSEs)
```
chaen: No, you have 1 operation

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -247,122 +373,170 @@ def dmTransfer( self, opFile ):
...
+
+              if self.rmsMonitoring:
+                self.rmsMonitoringReporter.addRecord({
+                    "timestamp": int(Time.toEpoch()),
+                    "host": Network.getFQDN(),
+                    "objectType": "File",
+                    "operationType": self.operation.Type,
+                    "parentID": self.operation.OperationID,
+                    "status": "Successful",
+                    "nbObject": 1
```
chaen: You can also do that outside of the loop

DataManagementSystem/Agent/RequestOperations/PutAndRegister.py
```diff
@@ -112,73 +161,113 @@ def __call__( self ):
...
+            "operationType": self.operation.Type,
+            "parentID": self.operation.OperationID,
+            "status": "Failed",
+            "nbObject": 1
+        }
       if not putAndRegister["OK"]:
-        gMonitor.addMark( "PutFail", 1 )
+        if self.rmsMonitoring:
+          failFileObj["timestamp"] = int(Time.toEpoch())
+          self.rmsMonitoringReporter.addRecord(failFileObj)
```
chaen: that can't possibly work.
You are constantly adding the same dictionary and updating the field, so in practice you are sending multiple time the exact same info.

DataManagementSystem/Agent/RequestOperations/ReTransfer.py
```diff
@@ -47,67 +52,144 @@ class ReTransfer( DMSRequestOperationsBase ):
...
+      if self.rmsMonitoring:
+        for status in ["Attempted", "Failed"]:
+          self.rmsMonitoringReporter.addRecord({
+              "timestamp": int(Time.toEpoch()),
+              "host": Network.getFQDN(),
+              "objectType": "File",
+              "operationType": self.operation.Type,
+              "parentID": self.operation.OperationID,
+              "status": status,
+              "nbObject": 1
```
chaen: that's wrong (it was wrong). Either you consider the operation to be attempted (and possibly failed) and then nbObject is 1. Or/and you count each files

DataManagementSystem/Agent/RequestOperations/ReTransfer.py
```diff
@@ -47,67 +52,144 @@ class ReTransfer( DMSRequestOperationsBase ):
...
+      reTransfer = se.retransferOnlineFile(opFile.LFN)
       if not reTransfer["OK"]:
         opFile.Error = reTransfer["Message"]
-        self.log.error( "Retransfer failed", opFile.Error )
-        gMonitor.addMark( "FileReTransferFail", 1 )
+        self.log.error("Retransfer failed", opFile.Error)
+
+        if self.rmsMonitoring:
+          fileObj["timestamp"] = int(Time.toEpoch())
+          fileObj["status"] = "Failed"
```
chaen: same as before. Always the same dict.

DataManagementSystem/Agent/RequestOperations/RegisterFile.py
```diff
@@ -43,75 +49,115 @@ class RegisterFile( OperationHandlerBase ):
...
-    dm = DataManager( catalogs = catalogs )
+      catalogs = [cat.strip() for cat in catalogs.split(',')]
+    dm = DataManager(catalogs=catalogs)
     # # get waiting files
     waitingFiles = self.getWaitingFilesList()
     # # loop over files
     for opFile in waitingFiles:
 
-      gMonitor.addMark( "RegisterAtt", 1 )
+      if self.rmsMonitoring:
```
chaen: here as well, why don't you take it out the loop ?

DataManagementSystem/Agent/RequestOperations/RegisterReplica.py
```diff
@@ -27,101 +33,149 @@ class RegisterReplica( DMSRequestOperationsBase ):
...
-      catalogs = [ cat.strip() for cat in catalogs.split( ',' ) ]
+      catalogs = [cat.strip() for cat in catalogs.split(',')]
     # # get waiting files
     waitingFiles = self.getWaitingFilesList()
     # # loop over files
     registerOperations = {}
     for opFile in waitingFiles:
 
-      gMonitor.addMark( "RegisterReplicaAtt", 1 )
+      if self.rmsMonitoring:
```
chaen: why not outside the loop ?

DataManagementSystem/Agent/RequestOperations/RemoveFile.py
```diff
@@ -35,118 +35,216 @@
...
+    res = fc.getReplicas([wf.LFN for wf in waitingFiles])
     if not res['OK']:
-      gMonitor.addMark( "RemoveFileAtt" )
-      gMonitor.addMark( "RemoveFileFail" )
+      if self.rmsMonitoring:
+        for status in ["Attempted", "Failed"]:
+          self.rmsMonitoringReporter.addRecord({
+              "timestamp": int(Time.toEpoch()),
+              "host": Network.getFQDN(),
+              "objectType": "File",
```
chaen: same as before. Either 1 operation and/or all files

DataManagementSystem/Agent/RequestOperations/RemoveReplica.py
```diff
@@ -31,90 +31,164 @@
...
-    bannedTargets = self.checkSEsRSS( targetSEs, access = 'RemoveAccess' )
+    bannedTargets = self.checkSEsRSS(targetSEs, access='RemoveAccess')
     if not bannedTargets['OK']:
-      gMonitor.addMark( "RemoveReplicaAtt" )
-      gMonitor.addMark( "RemoveReplicaFail" )
+      if self.rmsMonitoring:
+        for status in ["Attempted", "Failed"]:
+          self.rmsMonitoringReporter.addRecord({
+              "timestamp": int(Time.toEpoch()),
+              "host": Network.getFQDN(),
```
chaen: same, all files or 1 ops

DataManagementSystem/Agent/RequestOperations/RemoveReplica.py
```diff
@@ -31,90 +31,164 @@
...
+
+    if self.rmsMonitoring:
+      self.rmsMonitoringReporter.addRecord({
+          "timestamp": int(Time.toEpoch()),
+          "host": Network.getFQDN(),
+          "objectType": "File",
+          "operationType": self.operation.Type,
+          "parentID": self.operation.OperationID,
+          "status": "Attempted",
+          "nbObject": len(toRemoveDict) * len(targetSEs)
```
chaen: just number of files

DataManagementSystem/Agent/RequestOperations/RemoveReplica.py
chaen: I won't repeat it further, check and update

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -212,7 +222,18 @@ def __checkReplicas(self):
...
-        gMonitor.addMark("ReplicateFail", len(targetSESet))
+        if self.rmsMonitoring:
+          self.rmsMonitoringReporter.addRecord({
+              "timestamp": int(Time.toEpoch()),
+              "host": Network.getFQDN(),
+              "objectType": "File",
+              "operationType": self.operation.Type,
+              "parentID": self.operation.OperationID,
+              "status": "Failed",
+              "nbObject": len(targetSESet)
```
chaen: No,should be number of files

DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -430,12 +476,26 @@ def fts3Transfer(self):
 
       for ftsFile in fts3Files:
         opFile = rmsFilesIds[ftsFile.rmsFileID]
-        gMonitor.addMark("FTSScheduleOK", 1)
+        if self.rmsMonitoring:
```
chaen: why not outside the loop ? I won't repeat it either, check and update

RequestManagementSystem/Client/Request.py
pujanm: @chaen This what I get when I do the `pylint` check `E:477,28: Instance of 'Request' has no 'RequestID' member (no-member)` since this is an error that's why Travis fails.

chaen: Look how it is done somewhere else, for example https://github.com/DIRACGrid/DIRAC/blob/integration/RequestManagementSystem/Service/ReqManagerHandler.py#L120

pujanm: Thanks this worked.

MonitoringSystem/Client/Types/RMSMonitoring.py
```diff
@@ -0,0 +1,40 @@
+"""
```
chaen: Some more useful comments would be nice, for example what are the various fields

pujanm: Yes, I will add documentation now.

DataManagementSystem/Agent/RequestOperations/MoveReplica.py
```diff
@@ -10,198 +10,306 @@
...
+            self.rmsMonitoringReporter.addRecord({
+                "timestamp": int(Time.toEpoch()),
+                "host": Network.getFQDN(),
+                "objectType": "File",
+                "operationType": self.operation.Type,
+                "parentID": self.operation.OperationID,
+                "status": status,
+                "nbObject": len(self.operation)
+            })
+          self.rmsMonitoringReporter.commit()
```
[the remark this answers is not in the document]
pujanm: No, it is not singleton should I make it?

chaen:
Seems OK with me, I let @zmathe do the final review

zmathe:
@chaen review okay

pujanm:
Hi,
Do I need to update this PR?

fstagni:
Hi, well thanks, yes, updating would actually be helpful.

pujanm:
@fstagni I have updated the PR. Please tell me if any other changes are required.

pujanm:
@fstagni I have updated the PR. Can we move with it further?

pujanm:
@fstagni Hi, can we go ahead with this?

pujanm:
@atsareg Any updates I should make to this PR?

atsareg:
No. I have closed/reopened the PR in order to retrigger the tests

PR #4225, opened 2019-08-18
[v7r1] Fix bugs and conflicts with working with OAuthDIRAC

TaykYoku:
BEGINRELEASENOTES
Fix some bugs and conflicts

*ConfigurationSystem:
FIX: Add getUserDict and fix getInfoAboutProviders methods in Registry and Resources

*FrameworkSystem:
FIX: fix some bugs in ProxyDB

*Resources:
FIX: Edit DIRACCAProxyProvider to use M2Crypto

ENDRELEASENOTES

### DIFF
--- a/ConfigurationSystem/Client/Helpers/Resources.py
+++ b/ConfigurationSystem/Client/Helpers/Resources.py
@@ -304,34 +304,42 @@ def getFilterConfig(filterID):
   return gConfig.getOptionsDict('Resources/LogFilters/%s' % filterID)
 
 
-def getInfoAboutProviders(ofWhat=None, providerName=None, option='', section=''):
+def getInfoAboutProviders(of=None, providerName=None, option='', section=''):
   """ Get the information about providers
 
-      :param basestring ofWhat: provider of what(Id, Proxy or etc.) need to look,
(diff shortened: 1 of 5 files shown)


ConfigurationSystem/Client/Helpers/Resources.py
```diff
@@ -304,34 +304,42 @@ def getFilterConfig(filterID):
   return gConfig.getOptionsDict('Resources/LogFilters/%s' % filterID)
 
 
-def getInfoAboutProviders(ofWhat=None, providerName=None, option='', section=''):
+def getInfoAboutProviders(of=None, providerName=None, option='', section=''):
```
chaen: Maybe `providerType` rather than `of`?

TaykYoku: Definition of provider type already exist as OAuth, CAS, DRIACCA, Certificate and etc. Here means provider of what instance? Provider provide what? proxy, identity and etc.

chaen: ah ok. Thanks for clarifying. I still think that `of` is not a good name though, especially in `if` conditions :D

Issue #4088, opened 2019-05-24
[v7] StageMonitorAgent: replace plugin selection by protocol selection


andresailer:
As discussed in #4086 the StoragePlugins selection should be replaced by a ProtocolSelection, once that is working:

https://github.com/DIRACGrid/DIRAC/blob/1e234f7737a553238facc44160c0c99a345326e4/StorageManagementSystem/Agent/StageMonitorAgent.py#L76

Then
```python
res = StorageElement(storageElement).getFileMetadata(lfnRepIDs, protocols=self.protocols) 
```
and respective changes.

PR #4231, opened 2019-08-19
[v7r0] only pass protocols parameter to getTransportURL in executeMethod

chaen:
This should address https://github.com/DIRACGrid/DIRAC/issues/4088

Before:
```
  File "/tmp/testSE.py", line 6, in <module>
    print StorageElement('CERN-USER').getFileMetadata('/lhcb/user/c/chaen/zozo.xml', protocols='root')
  File "/opt/dirac/DIRAC/Resources/Storage/StorageElement.py", line 1234, in __executeMethod
    res = fcn(urlsToUse, *args, **kwargs)
TypeError: getFileMetadata() got an unexpected keyword argument 'protocols'
```
And after, it just returns the metadata

BEGINRELEASENOTES
*Resources
FIX: SE __executeMethod: only pass protocols parameter to getTransportURL

ENDRELEASENOTES

### DIFF
--- a/Resources/Storage/StorageElement.py
+++ b/Resources/Storage/StorageElement.py
@@ -1125,12 +1125,17 @@ class StorageElementItem(object):
         :param *args: variable amount of non-keyword arguments. SHOULD BE EMPTY
         :param **kwargs: keyword arguments
 
+        A special argument is 'protocols', which will be used by the StorageElement to filter
+        the usable plugins. Unless the method being executed is getTransportURL, this parameter
+        is removed from kwargs.
+        A special kwargs is 'inputProtocol', which can be specified for putFile. It describes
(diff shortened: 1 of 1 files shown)


andresailer:
Do you also want to apply this in the StageMonitorAgent?

atsareg:
I merge it now but please answer André's question

chaen:
I think this was the idea yes. At the same time. But since it is working now, and we anyway plan to get ride of it, is it worth it ?

andresailer:
Do you mean getting rid of the StageMonitorAgent?

It is find by me to keep the configuration as it is for now. Then we can close #4088 ?

chaen:
Yes, and replace it with FTS. 

I guess yes if you agree

PR #4255, opened 2019-08-27
[v7r0] dirac-resource-info command

atsareg:
This PR adds dirac-resource-info command to show available resource for a given VO. For the moment CEs and SEs are shown, other resources types can be added later.

BEGINRELEASENOTES

*Configuration
NEW: Resources - added getStorageElements() method

*Resources
NEW: Added dirac-resource-info command

*docs
NEW: dirac-resource-info added to CommandReference

ENDRELEASENOTES


andresailer:
You don't have to add the `dirac-resource-command.rst` file, except if you want to add a usage example.
The other parts are automatically generated from the commands `--help` message.
But you do have to add to this line:
https://github.com/DIRACGrid/DIRAC/blob/7fb81e743c88e511c1c0a654b3c23f9434749c5c/docs/docs.conf#L101
```
 pattern = dirac-proxy, dirac-info, myproxy, -resource-
```
for the doc generation to pick up commands with "-resource-" in their name in the "Others" section

Resources/scripts/dirac-resource-info.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+
+__RCSID__ = "$Id$"
+
+from DIRAC.Core.Base import Script
+from DIRAC import S_OK, gLogger, exit as DIRACExit
+
+Script.setUsageMessage('\n'.join(['Get information on resources',
+                                  'Usage:',
+                                  '%s [option]...' % Script.scriptName]))
```
andresailer:
```suggestion
                                  '  %s [option]...' % Script.scriptName]))
```

Resources/scripts/dirac-resource-info.py
```diff
@@ -0,0 +1,151 @@
...
+    records.append((se, status,
+                    ",".join(seDict[se]["Protocols"]),
+                    ",".join(seDict[se]["Aliases"])))
+
+  gLogger.notice(printTable(fields, records, printOut=False, columnSeparator='  '))
+  return S_OK()
+
+
+if __name__ == '__main__':
+  if not voName:
```
andresailer: Could you add a check if the voName is set by hand that the VO exists?

Resources/scripts/dirac-resource-info.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def setVOName(args):
+  global voName
+  voName = args
+
+
+Script.registerSwitch("C", "ce", "Get CE info", setCEFlag)
+Script.registerSwitch("S", "se", "Get SE info", setSEFlag)
+Script.registerSwitch("V:", "vo=", "Get resources for the given VO", setVOName)
```
andresailer:
```suggestion
Script.registerSwitch("V:", "vo=", "Get resources for the given VO. If not set taken from the proxy", setVOName)
```

ConfigurationSystem/Client/Helpers/Resources.py
```diff
@@ -194,6 +194,51 @@ def getQueues(siteList=None, ceList=None, ceTypeList=None, community=None, mode=
...
+  baseSEs = {}
+  for se in storageElements:
+    alias = False
+    for option in ('BaseSE', 'Alias'):
+      originalSE = gConfig.getValue('Resources/StorageElements/%s/%s' % (se, option))
+      if originalSE:
+        baseSEs.setdefault(originalSE, []).append(se)
+        alias = True
+        break
+    if not alias:
```
chrisburr: The `alias` variable can be removed:
```suggestion
    else:
```

andresailer:
Please also add the indentation in this line for this other command, which makes the docs fail
https://github.com/DIRACGrid/DIRAC/blob/integration/Resources/scripts/dirac-resource-get-parameters.py#L15

ConfigurationSystem/Client/Helpers/Resources.py
atsareg: ???

chrisburr: I guess you're questioning what a `for..else` does?

The `else` block after a `for` loop is ran if the loop ends without `break` being explicitly used. For example:

```python
for known_user in known_users:
    if user.lower() == known_user.lower():
        print('User is already known')
        break
else:
    print('ERROR: User is not known')
```

atsareg: Wow, I have learnt something today ! Thanks

atsareg:
Review OK ?

PR #4257, opened 2019-08-28
[v7r1] ProxyManager compatibility, add method to get e-mails in Registry, resolve issues

TaykYoku:
This PR resolve some issues with compatibility, bugs, added helper method.

BEGINRELEASENOTES

*Configuration
NEW: Add search emails for group method in Registry

*Framework
NEW: Filter of duplicate mails in Notification
FIX: Delete not used comments
FIX: Optimize test in NotificationHandler
FIX: Add old method storeProxy to proxyDB, fix compatibility for ProxyManager

*Resources
FIX: Rename 'EMail' to 'Email' in ProxyProviders

ENDRELEASENOTES

### DIFF
--- a/ConfigurationSystem/Client/Helpers/Registry.py
+++ b/ConfigurationSystem/Client/Helpers/Registry.py
@@ -502,7 +502,5 @@ def getEmailsForGroup(groupName):
   emails = []
   for username in getUsersInGroup(groupName, defaultValue=[]):
     email = gConfig.getValue("%s/Users/%s/Email" % (gBaseRegistrySection, username), [])
-    if not isinstance(email, list):
-      email = email.replace(' ', '').split(',')
     emails.append(email)
   return emails
--- a/ConfigurationSystem/Client/Helpers/Registry.py
+++ b/ConfigurationSystem/Client/Helpers/Registry.py
@@ -501,8 +501,8 @@ def getEmailsForGroup(groupName):
   """
   emails = []
   for username in getUsersInGroup(groupName, defaultValue=[]):
-    email = gConfig.getValue("%s/Users/%s/EMail" % (gBaseRegistrySection, username), [])
+    email = gConfig.getValue("%s/Users/%s/Email" % (gBaseRegistrySection, username), [])
     if not isinstance(email, list):
       email = email.replace(' ','').split(',')
-    emails += email
-  return list(set(emails))
+    emails.append(email)
+  return emails
--- a/ConfigurationSystem/Client/Helpers/Registry.py
+++ b/ConfigurationSystem/Client/Helpers/Registry.py
@@ -497,7 +497,7 @@ def getEmailsForGroup(groupName):
 
       :param basestring groupName: DIRAC group name
 
-      :return: basestring -- emails listed by ', '
+      :return: list
   """
   emails = []
   for username in getUsersInGroup(groupName, defaultValue=[]):
--- a/ConfigurationSystem/Client/Helpers/Registry.py
+++ b/ConfigurationSystem/Client/Helpers/Registry.py
@@ -497,7 +497,7 @@ def getEmailsForGroup(groupName):
 
       :param basestring groupName: DIRAC group name
 
-      :return: list
+      :return: list(list) -- inner list contains emails for a user
   """
   emails = []
   for username in getUsersInGroup(groupName, defaultValue=[]):
--- a/ConfigurationSystem/Client/Helpers/Registry.py
+++ b/ConfigurationSystem/Client/Helpers/Registry.py
@@ -490,3 +490,19 @@ def getUserDict(username):
     if value:
       resDict[key.replace(relPath, '')] = value
   return S_OK(resDict)
+
+
+def getEmailsForGroup(groupName):
+  """ Get email list of users in group
+
+      :param basestring groupName: DIRAC group name
+
+      :return: basestring -- emails listed by ', '
+  """
+  emails = []
+  for username in getUsersInGroup(groupName, defaultValue=[]):
+    email = gConfig.getValue("%s/Users/%s/EMail" % (gBaseRegistrySection, username), [])
+    if not isinstance(email, list):
+      email = email.replace(' ','').split(',')
+    emails += email
+  return list(set(emails))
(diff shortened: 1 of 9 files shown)


ConfigurationSystem/Client/Helpers/Registry.py
```diff
@@ -490,3 +490,19 @@ def getUserDict(username):
...
+  """ Get email list of users in group
+
+      :param basestring groupName: DIRAC group name
+
+      :return: list
+  """
+  emails = []
+  for username in getUsersInGroup(groupName, defaultValue=[]):
+    email = gConfig.getValue("%s/Users/%s/Email" % (gBaseRegistrySection, username), [])
+    if not isinstance(email, list):
```
chaen: shouldn't the `getValue` call already return a list ?

FrameworkSystem/Service/ProxyManagerHandler.py
```diff
@@ -84,13 +84,15 @@ def export_getUserProxiesInfo(self):
...
-  def export_requestDelegationUpload(self, requestedUploadTime):
+  def export_requestDelegationUpload(self, requestedUploadTime, diracGroup=None):
     """ Request a delegation. Send a delegation request to client
 
         :param int,long requestedUploadTime: requested live time
 
         :return: S_OK(dict)/S_ERROR() -- dict contain id and proxy as string of the request
     """
+    if diracGroup:
+      return S_ERROR("Cannot upload proxy with DIRAC group or VOMS extensions; Please update your DIRAC client.")
```
chaen: This is a completely breaking change. Running pilots and jobs will fail because of that. So this is probably not good.

ConfigurationSystem/Client/Helpers/Registry.py
```diff
@@ -490,3 +490,19 @@ def getUserDict(username):
...
+      email = email.replace(' ', '').split(',')
+    emails.append(email)
```
chrisburr: Is this function supposed to return a list of lists? From the docstring I guess this should be:
```suggestion
    emails.extend(email)
```

ConfigurationSystem/Client/Helpers/Registry.py
chrisburr: I thought it always returned a string? If so, would it be better to simplify:
```python
     email = gConfig.getValue("%s/Users/%s/Email" % (gBaseRegistrySection, username), '')
     email = email.replace(' ', '').split(',')
```

chrisburr: I thought it always returned a string? If so, would it be better to simplify:
```python
     email = gConfig.getValue("%s/Users/%s/Email" % (gBaseRegistrySection, username), '')
     email = email.replace(' ', '').split(',')
```

FrameworkSystem/Service/ProxyManagerHandler.py
TaykYoku: In v7 DIRAC group option not used in this method, all requests from old client will fail. But pilots and other "server side actions" will be from new DIRAC installation version, so its must work fine, how I understand.
This change for some old users Cli installation to notify "Client is old, need to update"

ConfigurationSystem/Client/Helpers/Registry.py
TaykYoku: Yes,` list( list(<user1 emails>), list(<user2 emails>).. )`
For example, Notification.sendMail() method can take list of user emails, but send only to one email - first that return success, so in this case:

```
for addresses in Registry.getEmailsForGroup('dirac_admin'):
  result = NotificationClient().sendMail(addresses, 'subject', 'body')

```

ConfigurationSystem/Client/Helpers/Registry.py
TaykYoku: Yes, how I understand returned value type is that set in the last argument. I didn't know about that.

TaykYoku: Fixed

chaen: @chrisburr yeah, in `getValue` `defaultValue` is actually `defaultType`... and if `defaultType` is an iterable, it does the split for you...

FrameworkSystem/Service/ProxyManagerHandler.py
chaen: all requests from old client will fail -> that is unacceptable. When we update version on teh servers, we still have many thousands of pilots and clients using the old version. There MUST be a backward compatibility of one version.

ConfigurationSystem/Client/Helpers/Registry.py
```diff
@@ -490,3 +490,17 @@ def getUserDict(username):
...
       resDict[key.replace(relPath, '')] = value
   return S_OK(resDict)
+
+
+def getEmailsForGroup(groupName):
+  """ Get email list of users in group
+
+      :param basestring groupName: DIRAC group name
+
+      :return: list
```
chrisburr: Following on from below, in that case:
```suggestion
      :return: list of lists -- inner list contains alternative emails for a user
```

FrameworkSystem/Service/ProxyManagerHandler.py
atsareg: The only thing that will fail is dirac-proxy-init from old clients. The user will receive a meaningful error message in this case suggesting upgrading the client. Old pilots are not affected. This is a compromise but a reasonable compromise I think. This should have been part of the major release (as was intended) v7r0 where we do not promise to keep interfaces compatible. Should we do v8r0 then ?

chaen: Sorry my bad, I thought it was just requesting a delegation (also for downloading the proxy).
So indeed, it is only uploading a proxy that would fail. Moreover, the user would still get it locally available. So the issue is quite limited. But instead of returning an error, why not just ignoring it ? 
We do not keep the backward compatibility for major version ?! That's new to me. Up to which point can we break it ?

atsareg: Andrii is looking at ensuring backward compatibility also for the proxy upload from an old client. Looks like this is feasible. 
I did not say that we do not keep interface compatibility for new major release, but this is not guaranteed. Actually, changing interfaces is the main reason for a major release.

ConfigurationSystem/Client/Helpers/Registry.py
TaykYoku: Done

FrameworkSystem/Service/ProxyManagerHandler.py
TaykYoku: It now seems compatible

TaykYoku:
OK for merging?

PR #4259, opened 2019-08-29
[v6r22] Ensure lcgBundle is installed in singularity container

sfayer:
This adds the lcgBundle version to the dirac-install flags within the SingularityCE container to make the behaviour like the normal grid pilot.

BEGINRELEASENOTES
*Resources
FIX: Ensure lcgBundle is installed in container if LCGBundleVersion is set.
ENDRELEASENOTES


atsareg:
This is fine. But we have to move to using DiracOS now.

PR #4260, opened 2019-08-29
[v7r0] New dirac-wms-match command

atsareg:
This PR adds dirac-wms-command which allows to test a given job description for matching with computing resources capabilities. Example of usage:

> dirac-wms-match -S EGI.SARA.nl,EGI.BARI.it -F test.jdl 
     Site         CE                     Queue               Status  Match  Reason
====================================================================================================================================
  1  EGI.BARI.it  cream-ce-2.ba.infn.it  cream-pbs-infinite  Active  Yes 
  2  EGI.SARA.nl  creamce3.gina.sara.nl  cream-pbs-medium    Active  No     Resource operational JobType requirement not satisfied

BEGINRELEASENOTES

*WMS
NEW: Added dirac-wms-match command to test a given job for matching computing resources
NEW: QueueUtilities - utilities relevent to queues manipultation

ENDRELEASENOTES

### DIFF
--- a/WorkloadManagementSystem/Utilities/QueueUtilities.py
+++ b/WorkloadManagementSystem/Utilities/QueueUtilities.py
@@ -173,7 +173,8 @@ def matchQueue(jobJDL, queueDict, fullMatch=False):
 
   # 5. RAM
   ram = job.getAttributeInt('RAM')
-  if ram and ram > int(queueDict['MaxRAM']):
+  # If MaxRAM is not specified in the queue description, assume 2GB
+  if ram and ram > int(queueDict.get('MaxRAM', 2048)) / 1024:
     noMatchReasons.append('Job RAM %d requirement not satisfied' % ram)
     if not fullMatch:
       return S_OK({'Match': False, 'Reason': noMatchReasons[0]})
--- a/WorkloadManagementSystem/Utilities/QueueUtilities.py
+++ b/WorkloadManagementSystem/Utilities/QueueUtilities.py
@@ -14,6 +14,7 @@ from DIRAC.Core.Security.ProxyInfo import getProxyInfo
 
 __RCSID__ = '$Id$'
 
+
 def getQueuesResolved(siteDict):
   """
   Get the list of queue descriptions merging site/ce/queue parameters and adding some
@@ -30,7 +31,7 @@ def getQueuesResolved(siteDict):
     for ce, ceDict in siteDict[site].items():
       qDict = ceDict.pop('Queues')
       for queue in qDict:
-        
+
         queueName = '%s_%s' % (ce, queue)
         queueDict = qDict[queue]
         queueDict['Queue'] = queue
@@ -83,7 +84,7 @@ def getQueuesResolved(siteDict):
           result = getDIRACPlatform(platform)
           if result['OK']:
             queueDict['Platform'] = result['Value'][0]
-        
+
         queueFinalDict[queueName] = queueDict
 
   return S_OK(queueFinalDict)
--- a/WorkloadManagementSystem/Utilities/QueueUtilities.py
+++ b/WorkloadManagementSystem/Utilities/QueueUtilities.py
@@ -10,6 +10,7 @@ from DIRAC.ConfigurationSystem.Client.Helpers.Resources import getDIRACPlatform
 from DIRAC.ConfigurationSystem.Client.Helpers.Operations import Operations
 from DIRAC.Core.Security.ProxyInfo import getProxyInfo
 
+
 def getQueuesResolved(siteDict):
   """
   Get the list of queue descriptions merging site/ce/queue parameters and adding some
@@ -70,7 +71,7 @@ def getQueuesResolved(siteDict):
                 else queueParameter
 
         # If we have a multi-core queue add MultiProcessor tag
-        if queueDict[queueName].get('NumberOfProcessors',1) > 1:
+        if queueDict[queueName].get('NumberOfProcessors', 1) > 1:
           queueDict[queueName].setdefault('Tag', []).append('MultiProcessor')
 
         queueDict[queueName]['CEName'] = ce
@@ -96,7 +97,8 @@ def getQueuesResolved(siteDict):
 
   return S_OK(queueDict)
 
-def matchQueue(jobJDL, queueDict, fullMatch = False):
+
+def matchQueue(jobJDL, queueDict, fullMatch=False):
   """
   Match the job description to the queue definition
 
@@ -143,7 +145,7 @@ def matchQueue(jobJDL, queueDict, fullMatch = False):
 
   # 3. Banned multi-value match requirements
   for par in ['Site', 'GridCE', 'Platform', 'GridMiddleware',
-                    'PilotType', 'SubmitPool', 'JobType']:
+              'PilotType', 'SubmitPool', 'JobType']:
     parameter = "Banned%s" % par
     if par in queueDict:
       valueSet = set(job.getListFromExpression(parameter))
@@ -166,7 +168,7 @@ def matchQueue(jobJDL, queueDict, fullMatch = False):
   wholeNode = job.getAttributeString('WholeNode')
   if wholeNode:
     tags.add('WholeNode')
-  queueTags = set(queueDict.get('Tags',[]))
+  queueTags = set(queueDict.get('Tags', []))
   if not tags.issubset(queueTags):
     noMatchReasons.append('Job Tag %s not satisfied' % ','.join(tags))
     if not fullMatch:
@@ -198,7 +200,7 @@ def matchQueue(jobJDL, queueDict, fullMatch = False):
         return S_OK({'Match': False, 'Reason': noMatchReasons[0]})
 
   # 2. Required tags
-  requiredTags = set(queueDict.get('RequiredTags',[]))
+  requiredTags = set(queueDict.get('RequiredTags', []))
   if not requiredTags.issubset(tags):
     noMatchReasons.append('Resource RequiredTags %s not satisfied' % ','.join(requiredTags))
     if not fullMatch:
--- a/WorkloadManagementSystem/Utilities/QueueUtilities.py
+++ b/WorkloadManagementSystem/Utilities/QueueUtilities.py
@@ -1,7 +1,8 @@
 """Utilities to help Computing Element Queues manipulation
 """
 
-__RCSID__ = '$Id$'
+from __future__ import absolute_import
+from __future__ import division
 
 from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Utilities.List import fromChar
@@ -10,6 +11,7 @@ from DIRAC.ConfigurationSystem.Client.Helpers.Resources import getDIRACPlatform
 from DIRAC.ConfigurationSystem.Client.Helpers.Operations import Operations
 from DIRAC.Core.Security.ProxyInfo import getProxyInfo
 
+__RCSID__ = '$Id$'
 
 def getQueuesResolved(siteDict):
   """
@@ -21,29 +23,27 @@ def getQueuesResolved(siteDict):
   :return: S_OK/S_ERROR, Value dictionary per queue with configuration data updated, e.g. for SiteDirector
   """
 
-  queueDict = {}
+  queueFinalDict = {}
 
   for site in siteDict:
-    for ce in siteDict[site]:
-      ceDict = siteDict[site][ce]
+    for ce, ceDict in siteDict[site].items():
       qDict = ceDict.pop('Queues')
       for queue in qDict:
+        
         queueName = '%s_%s' % (ce, queue)
-        queueDict[queueName] = qDict[queue]
-        queueDict[queueName] = qDict[queue]
-        queueDict[queueName]['Queue'] = queue
-        queueDict[queueName]['Site'] = site
+        queueDict = qDict[queue]
+        queueDict['Queue'] = queue
+        queueDict['Site'] = site
         # Evaluate the CPU limit of the queue according to the Glue convention
         # To Do: should be a utility
-        if "maxCPUTime" in queueDict[queueName] and \
-           "SI00" in queueDict[queueName]:
-          maxCPUTime = float(queueDict[queueName]['maxCPUTime'])
+        if "maxCPUTime" in queueDict and "SI00" in queueDict:
+          maxCPUTime = float(queueDict['maxCPUTime'])
           # For some sites there are crazy values in the CS
           maxCPUTime = max(maxCPUTime, 0)
           maxCPUTime = min(maxCPUTime, 86400 * 12.5)
-          si00 = float(queueDict[queueName]['SI00'])
-          queueCPUTime = 60. / 250. * maxCPUTime * si00
-          queueDict[queueName]['CPUTime'] = int(queueCPUTime)
+          si00 = float(queueDict['SI00'])
+          queueCPUTime = 60 / 250 * maxCPUTime * si00
+          queueDict['CPUTime'] = int(queueCPUTime)
 
         # Tags & RequiredTags defined on the Queue level and on the CE level are concatenated
         # This also converts them from a string to a list if required.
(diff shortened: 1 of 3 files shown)


WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
+"""Utilities to help Computing Element Queues manipulation
+"""
+
```
chrisburr: I think any new files should start with:
```suggestion
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```
These imports will help avoid new Python 3 incompatibilities being introduced and, unlike `unicode_literals`, are contained to a single file.

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
+"""
+  Get computing resources capable to execute the given job
+"""
```
chrisburr:
```suggestion
"""
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+        # Evaluate the CPU limit of the queue according to the Glue convention
+        # To Do: should be a utility
+        if "maxCPUTime" in queueDict[queueName] and \
+           "SI00" in queueDict[queueName]:
+          maxCPUTime = float(queueDict[queueName]['maxCPUTime'])
+          # For some sites there are crazy values in the CS
+          maxCPUTime = max(maxCPUTime, 0)
+          maxCPUTime = min(maxCPUTime, 86400 * 12.5)
+          si00 = float(queueDict[queueName]['SI00'])
+          queueCPUTime = 60. / 250. * maxCPUTime * si00
```
chrisburr: If future division is imported:
```suggestion
          queueCPUTime = 60 / 250 * maxCPUTime * si00
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+    return S_ERROR('Invalid job description')
+
+  noMatchReasons = []
+
+  # Check job requirements to resource
+  # 1. CPUTime
+  cpuTime = job.getAttributeInt('CPUTime')
+  if not cpuTime:
+    cpuTime = 84600
+  if cpuTime and cpuTime > queueDict.get('CPUTime', 0.):
```
chrisburr: `cpuTime` will never be falsey
```suggestion
  if cpuTime > queueDict.get('CPUTime', 0):
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+  queueDict = {}
+
+  for site in siteDict:
+    for ce in siteDict[site]:
+      ceDict = siteDict[site][ce]
+      qDict = ceDict.pop('Queues')
+      for queue in qDict:
+        queueName = '%s_%s' % (ce, queue)
+        queueDict[queueName] = qDict[queue]
+        queueDict[queueName] = qDict[queue]
```
chrisburr: Duplicated line
```suggestion
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+
+  queueDict = {}
+
+  for site in siteDict:
+    for ce in siteDict[site]:
+      ceDict = siteDict[site][ce]
+      qDict = ceDict.pop('Queues')
+      for queue in qDict:
+        queueName = '%s_%s' % (ce, queue)
+        queueDict[queueName] = qDict[queue]
```
chrisburr: Instead of writing `[queueName]` 30 times I think this it would be cleaner to use
```suggestion
        queueConf = qDict[queue]
```
then end the loop with
```python
        queueDict[queueName] = queueConf
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+        queueDict[queueName] = qDict[queue]
+        queueDict[queueName]['Queue'] = queue
+        queueDict[queueName]['Site'] = site
+        # Evaluate the CPU limit of the queue according to the Glue convention
+        # To Do: should be a utility
+        if "maxCPUTime" in queueDict[queueName] and \
```
chrisburr: There's no need to wrap this line
```suggestion
        if "maxCPUTime" in queueDict[queueName] and "SI00" in queueDict[queueName]:
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+          maxCPUTime = min(maxCPUTime, 86400 * 12.5)
+          si00 = float(queueDict[queueName]['SI00'])
+          queueCPUTime = 60. / 250. * maxCPUTime * si00
+          queueDict[queueName]['CPUTime'] = int(queueCPUTime)
+
+        # Tags & RequiredTags defined on the Queue level and on the CE level are concatenated
+        # This also converts them from a string to a list if required.
+        for tagFieldName in ('Tag', 'RequiredTag'):
+          ceTags = ceDict.get(tagFieldName, [])
+          if isinstance(ceTags, basestring):
```
chrisburr:
```suggestion
          if isinstance(ceTags, six.string_types):
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+            ceTags = fromChar(ceTags)
+          queueTags = queueDict[queueName].get(tagFieldName)
+          if queueTags and isinstance(queueTags, basestring):
```
chrisburr: I don't see a reason to check the truthiness of `queueTags` here:
```suggestion
          if isinstance(queueTags, six.string_types):
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+            queueTags = fromChar(queueTags)
+            queueDict[queueName][tagFieldName] = queueTags
```
chrisburr: These 7 lines can be replaced with:
```suggestion
          queueDict[queueName][tagFieldName] = list(set(ceTags + queueTags))
```
(Also the indentation was wrong)

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+        if "Platform" in queueDict[queueName]:
+          platform = queueDict[queueName]['Platform']
+        elif "Platform" in ceDict:
+          platform = ceDict['Platform']
+        elif "OS" in ceDict:
+          architecture = ceDict.get('architecture', 'x86_64')
+          platform = '_'.join([architecture, ceDict['OS']])
+
+        queueDict[queueName]['Platform'] = platform
+        if "Platform" not in queueDict[queueName] and platform:
```
chrisburr: The first half of this condition will never be true as it's set on the line above

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+      for parameter in result['Value']:
+        value = job.getAttributeString(parameter)
+        if value and opsHelper.getValue('JobScheduling/RunningLimit/%s/%s/%s' % (site, parameter, value), 1) == 0:
+          noMatchReasons.append('Resource operational %s requirement not satisfied' % parameter)
+          if not fullMatch:
+            return S_OK({'Match': False, 'Reason': noMatchReasons[0]})
+
+  if noMatchReasons:
+    return S_OK({'Match': False, 'Reason': noMatchReasons})
+  return S_OK({'Match': True, 'Reason': noMatchReasons})
```
chrisburr: I think this is better rather than having two branches to test:
```suggestion
  return S_OK({'Match': not bool(noMatchReasons), 'Reason': noMatchReasons})
```

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
...
+from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
+from DIRAC.ConfigurationSystem.Client.Helpers import Resources
+from DIRAC.Core.Utilities.PrettyPrint import printTable
+from DIRAC.ResourceStatusSystem.Client.ResourceStatus import ResourceStatus
+from DIRAC.ResourceStatusSystem.Client.SiteStatus import SiteStatus
+from DIRAC.WorkloadManagementSystem.Utilities.QueueUtilities import getQueuesResolved, matchQueue
+
+
+if __name__ == '__main__':
+  with file(args[0]) as f:
```
chrisburr: `file` doesn't exist in Python 3:
```suggestion
  with open(args[0]) as f:
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+            queueDict[queueName][tagFieldName] = queueTags
+          if ceTags:
+            if queueTags:
+              allTags = list(set(ceTags + queueTags))
+              queueDict[queueName][tagFieldName] = allTags
+            else:
+              queueDict[queueName][tagFieldName] = ceTags
+
+        # Some parameters can be defined on the CE level and are inherited by all Queues
+        for parameter in ['MaxRAM', 'NumberOfProcessors', 'WholeNode']:
```
chrisburr: I think this would be clearer if written as:
```suggestion
        for parameter in ['MaxRAM', 'NumberOfProcessors', 'WholeNode']:
          queueParameter = queueDict[queueName].get(parameter, ceDict.get(parameter))
          if queueParameter:
            queueDict[queueName][parameter] = queueParameter
```

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
...
+    if rssClient.rssFlag:
+      result = rssClient.getElementStatus(ce, "ComputingElement")
+      if result['OK']:
+        ceStatus = result['Value'][ce]['all']
+
+    result = matchQueue(jdl, queueDict[queue], fullMatch=fullMatch)
+    if not result['OK']:
+      gLogger.error('Failed in getting match data', result['Message'])
+      DIRACExit(-1)
+    status = "Active" if siteStatus is "Active" and ceStatus is "Active" else "InActive"
```
chrisburr: Should the `A` really be capitalised? Without I read it as `in active state`
```suggestion
    status = "Active" if siteStatus is "Active" and ceStatus is "Active" else "Inactive"
```

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
...
+    gLogger.error('Failed to get Site mask information')
+    DIRACExit(-1)
+  siteMaskList = resultMask.get('Value', [])
+
+  rssClient = ResourceStatus()
+
+  fields = ('Site', 'CE', 'Queue', 'Status', 'Match', 'Reason')
+  records = []
+
+  for queue in queueDict:
```
chrisburr: I think the value is used enough times to justify assigning it to a variable:
```suggestion
  for queue, queueInfo in queueDict.items():
```

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
...
+
+from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
+from DIRAC.ConfigurationSystem.Client.Helpers import Resources
+from DIRAC.Core.Utilities.PrettyPrint import printTable
+from DIRAC.ResourceStatusSystem.Client.ResourceStatus import ResourceStatus
+from DIRAC.ResourceStatusSystem.Client.SiteStatus import SiteStatus
+from DIRAC.WorkloadManagementSystem.Utilities.QueueUtilities import getQueuesResolved, matchQueue
+
+
+if __name__ == '__main__':
```
chrisburr: What is the reason for putting this in the middle here? If it's to make the file importable, it should also capture the `Script` setup.

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,225 @@
...
+
+  :param dict siteDict: dictionary with configuration data as returned by Resources.getQueues() method
+
+  :return: S_OK/S_ERROR, Value dictionary per queue with configuration data updated, e.g. for SiteDirector
+  """
+
+  queueDict = {}
+
+  for site in siteDict:
+    for ce in siteDict[site]:
```
chrisburr:
```suggestion
    for ce, ceDict in siteDict[site].items():
```
and drop the next line

WorkloadManagementSystem/scripts/dirac-wms-match.py
```diff
@@ -0,0 +1,110 @@
...
+
+description = """Get computing resources capable to execute a job with the given description.
+
+Note that only statically defined computing resource parameters are considered although sites
+can fail matching due to their dynamic state, e.g. occupancy by other jobs.
+"""
+
+Script.setUsageMessage('\n'.join([description,
+                                  'Usage:',
+                                  '%s [option]... <job_JDL>' % Script.scriptName]))
```
andresailer:
```suggestion
                                  '  %s [option]... <job_JDL>' % Script.scriptName]))
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: If we put it into the DIRAC.__init__ - will it be by definition applied in all the modules ? Or it should be put in each module ?

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: If future division is imported, these two forms will give the same result I guess. So, is there any reason that the second form is better ?

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: OK

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: Agreed

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: Again, should we convert all basestring's into six.string_types ? I think we use basestring only for type checking

chrisburr: It's exactly equivalent and `six` is already a dependency so there is no reason not to.

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: Lines like that are certainly smarter. But when reading the code you will certainly spend more time to read and understand this one line than reading the three lines above. Squeezing several lines of code in just one is not always better (although smarter :) )

WorkloadManagementSystem/Utilities/QueueUtilities.py
chrisburr: Yeah the future import makes them equivalent. I would favour removing the trailing periods as it's easier to read and better matches with what almost every Python package uses.

WorkloadManagementSystem/scripts/dirac-wms-match.py
atsareg: Well, the rest is just an executable code and not some function definition. So, in this case the part under "if __name__ == '__main__':"  can not be made shorter

WorkloadManagementSystem/Utilities/QueueUtilities.py
chrisburr: It's applied on a per file basis which is what makes it safe to use here and for any other new files in DIRAC. Applying the division import everywhere would probably cause (and fix) bugs. Hopefully #4240 will help smooth the transition.

WorkloadManagementSystem/Utilities/QueueUtilities.py
atsareg: Can you remind me - in Python 3 six.string_types will have to be replaced by something also ? Why we should bother with this replacement now if it is exactly equivalent ?

WorkloadManagementSystem/Utilities/QueueUtilities.py
chrisburr: Would you find something like this better?
```suggestion
  hasMatch = len(noMatchReasons) == 0
  return S_OK({'Match': hasMatch, 'Reason': noMatchReasons})
```
The reason I suggest avoiding the `if` is to avoid accidentally making mistakes like below. In this case I'm almost it's correct but reducing the number of "paths" through the code makes it much easier to test and maintain.
```python
  if noMatchReasons:
    return S_OK({'Match': False, 'Reeson': noMatchReasons})
  return S_OK({'Match': True, 'Reason': noMatchReasons})
```

WorkloadManagementSystem/Utilities/QueueUtilities.py
chrisburr: So we don't have to replace it in the [near!](https://pythonclock.org/) future.

Python 2: `six.string_types == (basestring,)`
Python 3: `six.string_types == (str,)`

WorkloadManagementSystem/Utilities/QueueUtilities.py
```diff
@@ -0,0 +1,214 @@
...
+
+  # 4. MultiProcessor requirements
+  if nProc and nProc > int(queueDict.get('NumberOfProcessors', 1)):
+    noMatchReasons.append('Job NumberOfProcessors %d requirement not satisfied' % nProc)
+    if not fullMatch:
+      return S_OK({'Match': False, 'Reason': noMatchReasons[0]})
+
+  # 5. RAM
+  ram = job.getAttributeInt('RAM')
+  if ram and ram > int(queueDict['MaxRAM']):
```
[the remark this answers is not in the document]
atsareg: True, corrected

atsareg:
Review OK ?

PR #4279, opened 2019-09-18
[v7r1] VOMSSecurityManager replacing VOMSPolicy

atsareg:
The new mecanism for dynamic loading of the DFC plug-ins makes SecurityPolicy plugins obsoleted. The PolicyBasedSecurityManager/VOMSPolicy combo is replaced by the VOMSSecurityManager class with the same instantiation method as for other SecurityManagers.

BEGINRELEASENOTES

*DMS

NEW: FileCatalog - added VOMSSecurityManager class
CHANGE: FileCatalog - dropped PolicyBasedSecurityManager and VOMSPolicy classes

*tests
CHANGE: use VOMSSecurityManager for integration instead of PolicyBasedSecurityManager/VOMSPolicy

ENDRELEASENOTES


chaen:
From the migration point of view, how will that work ? Can we just install a v7r1 DFC service with the existing configuration and it will work, or how will that happen ?

atsareg:
The only case where intervention will be needed is LHCb - changing in the configuration the SecurityManager from PolicyBasedSecurityManager to VOMSSecurityManager. VOMSSecurityManager is actually VOMSPolicy made as a manager

chaen:
Then can you keep the backward compatibility hack to avoid that problem ?

chaen:
ah well in fact it's ok if we change the config before installing, it should not harm then, instructions just have to be damn clear in the wiki

atsareg:
I guess that only LHCb is concerned, I do not know anybody else using VOMSPolicy. So, indeed in the case of LHCb this option must be changed. But I think this is OK, and does not need any backward compatibility hack

atsareg:
I have added instructions to the v7r1 release notes

Issue #4262, opened 2019-08-29
Error in SiteDirector from remote site.


yujikato:
Following line cause error for SiteDirector executed from remote
https://github.com/DIRACGrid/DIRAC/blob/61ef8e6adad2917498a73bf09daccc19bd85f5df/WorkloadManagementSystem/Agent/SiteDirector.py#L1352

- RPC call does not allow keyword argument.
- https://github.com/DIRACGrid/DIRAC/blob/61ef8e6adad2917498a73bf09daccc19bd85f5df/WorkloadManagementSystem/Service/WMSAdministratorHandler.py#L639-L645
  WMSAdministrator assumes "reason", not "statusReason"

Following works in my environment.
`        result = pilotAgentsDB.setPilotStatus(pRef, newStatus, '', 'Updated by SiteDirector')
`

Potentially, following line also cause issue (no problem in our environment as it is executed at local):
https://github.com/DIRACGrid/DIRAC/blob/61ef8e6adad2917498a73bf09daccc19bd85f5df/WorkloadManagementSystem/Agent/PilotStatusAgent.py#L119

Should be modified as follows:
 result = self.pilotDB.setPilotStatus(pilotRef, 'Stalled', '', 'Exceeded max waiting time')

Issue #4263, opened 2019-08-29
Need to set X509_USER_PROXY for pilot job submission to gsissh sites


yujikato:
Job submissions to gsissh sites are failed because X509_USER_PROXY is not set.
Overriding  setProxy in SSHComputingElement as follows works.
```
  def setProxy(self, proxy, valid=0):
    ComputingElement.setProxy(self, proxy, valid)
    if self.ceParameters.get('SSHType','ssh') == 'gsissh':
        result = self._prepareProxy()
        if not result['OK']:
          gLogger.error('SSHComputingElement: failed to set up proxy', result['Message'])
          return result
```

Do you think this is proper modification or any other way to set X509_USER_PROXY
for gsissh sites?

PR #4292, opened 2019-09-30
[v6r21] Several WMS/Resources fixes

atsareg:
This PR fixes issues #4262 and #4263

BEGINRELEASENOTES

*Resources
FIX: SSHComputingElement - define X509_USER_PROXY in case of gsissh access

*WMS
FIX: SiteDirector - do not use keyword arguments when making setPilotStatus call 

ENDRELEASENOTES

### DIFF
--- a/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -1349,7 +1349,7 @@ class SiteDirector(AgentModule):
 
       if newStatus:
         self.log.info('Updating status to %s for pilot %s' % (newStatus, pRef))
-        result = pilotAgentsDB.setPilotStatus(pRef, newStatus, statusReason='Updated by SiteDirector')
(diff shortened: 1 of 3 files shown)


WorkloadManagementSystem/Agent/SiteDirector.py
```diff
@@ -1349,7 +1349,7 @@ def _updatePilotStatus(self, pilotRefs, pilotDict, pilotCEDict):
 
       if newStatus:
         self.log.info('Updating status to %s for pilot %s' % (newStatus, pRef))
-        result = pilotAgentsDB.setPilotStatus(pRef, newStatus, statusReason='Updated by SiteDirector')
+        result = pilotAgentsDB.setPilotStatus(pRef, newStatus, '', 'Updated by SiteDirector')
```
zmathe: is the setPilotStatus method has changed? Why you need this modification?

atsareg: This is a usual case where either service or direct db access is used depending on the db accessibility. The changed call was working with the database but not with the service which is not allowing keyword arguments. This fix makes both kinds of the call working

PR #4327, opened 2019-10-28
[v6r22] FTS: workaround for FTS job/files inconsistencies

chaen:
There are cases when the FTS job is declared in a final state, while the files are not. This creates problems on our side, because the files are then never monitored nor retransferred. 
This PR aims at preventing that by not updating job/files in an inconsistent state. 
The FTS servers have a cleaning thread running every hour

BEGINRELEASENOTES
*DMS
FIX: avoid introducing inconsistencies in the DB due to the FTS inconsistencies

ENDRELEASENOTES


phicharp:
Did you also open a ticket to FTS as this should be fixed on their side?

chaen:
I've asked Andrea to open a ticket, and I've added the ticket to the comment.

PR #4332, opened 2019-10-29
Rel v7r0 rss command exception handling fix.

martynia:
The aim of the change is to:

- remove a call to shifterProxy configuration which would not work for multi VO Dirac

- handle a possible exception which might be thrown in a SE loop, so the SE is skipped and the loop continues to process remaining SEs. This might happen, for example where an SE configuration does not define _localAccessProtocolList_ configuration option.

### DIFF
--- a/ResourceStatusSystem/Agent/CacheFeederAgent.py
+++ b/ResourceStatusSystem/Agent/CacheFeederAgent.py
@@ -45,8 +45,6 @@ class CacheFeederAgent(AgentModule):
     """ Define the commands to be executed, and instantiate the clients that will be used.
     """
 
-    self.am_setOption('shifterProxy', 'DataManager')
-
     res = ObjectLoader().loadObject('DIRAC.ResourceStatusSystem.Client.ResourceStatusClient',
                                     'ResourceStatusClient')
     if not res['OK']:
--- a/ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
+++ b/ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
@@ -147,11 +147,15 @@ class FreeDiskSpaceCommand(Command):
     """
 
     for name in DMSHelpers().getStorageElements():
-      # keeping TB as default
-      diskSpace = self.doNew((name, 'MB'))
-      if not diskSpace['OK']:
-        gLogger.warn("Unable to calculate free/total disk space", "name: %s" % name)
-        gLogger.warn(diskSpace['Message'])
-        continue
+      try:
+        # keeping TB as default
+        diskSpace = self.doNew((name, 'MB'))
+        if not diskSpace['OK']:
+          gLogger.warn("Unable to calculate free/total disk space", "name: %s" % name)
+          gLogger.warn(diskSpace['Message'])
+          continue
+      except Exception as excp:  # pylint: disable=broad-except                                                                                                                                                                            
+          gLogger.error("Failed to get SE %s FreeDiskSpace information (SE skipped) " % name)
+          gLogger.exception("Operation finished  with exception: ", lException=excp)
 
     return S_OK()
--- a/ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
+++ b/ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
@@ -155,7 +155,7 @@ class FreeDiskSpaceCommand(Command):
           gLogger.warn(diskSpace['Message'])
           continue
       except Exception as excp:  # pylint: disable=broad-except
-          gLogger.error("Failed to get SE %s FreeDiskSpace information (SE skipped) " % name)
-          gLogger.exception("Operation finished  with exception: ", lException=excp)
+        gLogger.error("Failed to get SE %s FreeDiskSpace information (SE skipped) " % name)
+        gLogger.exception("Operation finished  with exception: ", lException=excp)
 
     return S_OK()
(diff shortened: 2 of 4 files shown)


ResourceStatusSystem/Agent/CacheFeederAgent.py
```diff
@@ -49,8 +49,6 @@ def initialize(self):
     """ Define the commands to be executed, and instantiate the clients that will be used.
     """
 
-    self.am_setOption('shifterProxy', 'DataManager')
```
chaen: I believe this agent also gets the accounting information from the storage. In that case, you must have a proxy

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
```diff
@@ -147,11 +147,15 @@ def doMaster(self):
...
+      try:
+        # keeping TB as default
+        diskSpace = self.doNew((name, 'MB'))
+        if not diskSpace['OK']:
+          gLogger.warn("Unable to calculate free/total disk space", "name: %s" % name)
+          gLogger.warn(diskSpace['Message'])
+          continue
+      except Exception as excp:  # pylint: disable=broad-except
+        gLogger.error("Failed to get SE %s FreeDiskSpace information (SE skipped) " % name)
+        gLogger.exception("Operation finished  with exception: ", lException=excp)
```
chaen: `continue` ?

chaen:
Can you elaborate on the comment with `localAccessProtocolList` ? I don't really understand how can this throw an exception.

ResourceStatusSystem/Agent/CacheFeederAgent.py
martynia: This call can be replaced by a relevant configuration option, so it is not required, is it ?

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
martynia: Well, I don't know why someone has put continue at the end of a for loop (within the if). I retained it.

martynia:
> Can you elaborate on the comment with `localAccessProtocolList` ? I don't really understand how can this throw an exception.

If the option is missing from the SE options and Dirac tries to use this attribute, this is what happens:

`2019-10-29 15:53:16 UTC ResourceStatus/CacheFeederAgent ERROR: Failed to get SE RAL-LCG29-tape information (SE skipped) 
2019-10-29 15:53:16 UTC ResourceStatus/CacheFeederAgent ERROR: Operation finished  with exception: 
  File "/opt/dirac/DIRAC/Core/Utilities/SiteSEMapping.py", line 72, in getStorageElementsHosts
    seHost = getSEHosts(seName)
  File "/opt/dirac/DIRAC/Core/Utilities/SiteSEMapping.py", line 47, in getSEHosts
    seParameters = getSEParameters(seName)
  File "/opt/dirac/DIRAC/Core/Utilities/SiteSEMapping.py", line 26, in getSEParameters
    protocolsSet = set(se.localAccessProtocolList) | set(se.localWriteProtocolList)
  File "/opt/dirac/DIRAC/Resources/Storage/StorageElement.py", line 1275, in __getattr__
    raise AttributeError("StorageElement does not have a method '%s'" % name)
AttributeError: StorageElement does not have a method 'localAccessProtocolList'`

This is true both for Dowmntime and FreeDiskSpace command execution workflows.
If you don't catch it early enough it leaks to the main agent code and remaining SEs in the loops are not considered.

ResourceStatusSystem/Agent/CacheFeederAgent.py
chaen: Sure, but then it should be:
* In the doc
* not for a patch release

ResourceStatusSystem/Command/FreeDiskSpaceCommand.py
chaen: Hum, probably code moved with a copy paste :-)

chaen:
Your error is because you don't have a credential and the CS is not loaded, so there is an exception in the instantiation of the SE.
You don't see it because:
* DIRAC does not propagate exception (sigh....)
* the SE is instanciated with `hideExceptions=True` 

But `localAccessProtocolList` is not an entry in the CS ever

martynia:
Some SEs don't define a _Protocol_ option in their _AccessProtocol.1_ section (RAL-LCG21-disk being an example) which throws an exception which would normally be caught by CacheFeederAgent (line 152) . My correction prevents this. _localAccessProtocol_ list is then empty (it is not an option sorry my bad)
This exception is only thrown for 'badly' defined SE, like one above. All other SE's are loaded correctly from the CS and appear in the downtime table if reported.  Also free disk space is correctly reported for the sanbox SE.

chaen:
I am not sure it's the responsibility of the end user code (i.e. RSS) to try to correct the mistakes of the badly defined SE. They should just be fixed, don't you think ?

martynia:
Well,
Probably. But the SEs are defined automatically by the BDII agent, if the information is missing then it causes the CacheFeederAgent to skip other SEs, so mistake on 1 SE causes other, well defined SEs to be ignored. We should avoid this dependency, I believe.

ResourceStatusSystem/Agent/CacheFeederAgent.py
martynia: I could re-base it wrt v7r0, would it be a better approach ?

chaen:
True, I agree with you. But how about fixing the agent ? You apparently dug quite a bit already :-)

martynia:
Which agent, BDII or CacheFeeder ? The BDII agent can only supply access protocols if it can get hold of one. There is no default I'm told. @sfayer could you comment on this, please ?

martynia:
Replaced by #4336

PR #4335, opened 2019-10-30
[v7r0] Docs: add words about GRIDFTP SESSION_REUSE

andresailer:
This belongs to #4334 but to avoid conflicts because of the admin doc restructuring this is a separate PR.

### DIFF
--- a/docs/source/AdministratorGuide/HowTo/multiVO.rst
+++ b/docs/source/AdministratorGuide/HowTo/multiVO.rst
@@ -1,3 +1,5 @@
+.. _multi_vo_dirac:
+
 ==============
 Multi-VO DIRAC
 ==============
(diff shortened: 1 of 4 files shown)


andresailer:
Added a page for environment variables:
Selected by
```
$ git grep "os.environ" | grep -w "DIRAC_[A-Za-z0-9_]*" -o | sort -u
DIRAC_DEBUG_DENCODE_CALLSTACK
DIRAC_DEBUG_STOMP
DIRAC_DEPRECATED_FAIL
DIRAC_PROCESSORS
DIRAC_USE_M2CRYPTO
DIRAC_VOMSES
DIRAC_WHOLENODE
```
I think WHOLENODE, and PROCESSORS are not relevant and just set internally by jobwrapper, pilot etc.

PR #4360, opened 2019-12-03
[v7r0] HTCondorCE: modify the PilotJobReferences to determine the paths of the job outputs

aldbr:
To avoid to have too many of them in a single directory, job outputs are randomly sorted into multiple directories on two levels on the server: `workingDirectory/<random_dir1>/<random_dir2>/<job_output>`

The information about their locations is not kept in the pilot job reference which is only composed of the job id. Therefore, the only way to retrieve an output is to perform a `find` command in the `workingDirectory` with the job id.

This PR aims at adding the name of the directories in the pilot job reference to quickly retrieve the associated job output.

Here is the current structure of the pilot job reference:
`htcondorce://<ce name>/<clusterid>.<jobid>`
Here is the proposed one:
`htcondorce://<ce name>/<random dir1><random dir2>-<clusterid>.<jobid>`

The previous pilot job references are still supported in this PR, so that the change is progressive. 

BEGINRELEASENOTES

*Resources
CHANGE: in HTCondor CEs, the pilot stamps are now used to retrieve outputs and logging info.  

ENDRELEASENOTES

### DIFF
--- a/Resources/Computing/HTCondorCEComputingElement.py
+++ b/Resources/Computing/HTCondorCEComputingElement.py
@@ -43,6 +43,8 @@ from DIRAC.Core.Utilities.Subprocess import Subprocess
 
 from DIRAC.Resources.Computing.BatchSystems.Condor import parseCondorStatus, treatCondorHistory
 
+from DIRAC.Core.Utilities.Decorators import deprecated
+
 __RCSID__ = "$Id$"
 
 CE_NAME = 'HTCondorCE'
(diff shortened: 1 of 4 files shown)


andresailer:
If I remember correctly, the pilot reference is also used to associated the payload job to the pilot.
So the pilot reference stored in the pilot database has to be the same that is created on the worker node in this piece of code

https://github.com/DIRACGrid/DIRAC/blob/91e8925639658a3848327e11a83f25614d690f18/WorkloadManagementSystem/PilotAgent/pilotCommands.py#L666-L669

chaen:
The question is probably stupid since I know nothing of this part of the code, but why not using a deterministic location instead of a random one ? I seem to remember it is because when submitting we do not have enough info, but how about creating a tmp directory, submitting from there, and then move to a fixed location based on the pilotID or something like that  ?

chaen:
though I do confirm that the `find` command is a bottlenecked, I've measured it in prod in the past

andresailer:
As far as I remember the location of the log, stdout and stderr file on the submission machine cannot change after the condor job (the pilot) is submitted (I could be wrong, misread the documentation, or this is now possible...). So we need to fix this location at submission time. And then one has to do something to avoid having tens of thousands of files in one folder, if one ever has to do things like `ls` on that folder.

andresailer:
If these pilotStamps don't have to be unique (at least in the DB their column is not required to be unique)

https://github.com/DIRACGrid/DIRAC/blob/6526a04158d1aec7515676ce0f3dcf85c275f8b6/Resources/Computing/HTCondorCEComputingElement.py#L266-L268

One could use them (one stamp per submitted chunk of condor jobs) to spread out the location, and get the location back
https://github.com/DIRACGrid/DIRAC/blob/6526a04158d1aec7515676ce0f3dcf85c275f8b6/Resources/Computing/HTCondorCEComputingElement.py#L79
as the part behind `:::` (or the first 6 characters split in threes).

aldbr:
According to @atsareg:

> The purpose of the pilot stamp is to have a unique pilot identifier assigned by DIRAC itself before the pilot is submitted. This is needed for some CE types, for example, to specify a unique task name (or output file name) in order to later retrieve the output with this name. For some CE types it is not needed indeed. But for generality it is always generated. So, it should be unique.

The pilot stamp doesn't seem to be used in the HTCondorCEs but it has to be unique. 
In this PR, I'm changing the way we generate this stamp:
- The first 3 characters of the stamp are common to a given chunk of pilots to generate. This will be helpful to retrieve the path of the result which will be located in: `<working dir>/<ce name>/<stamp[0]>/<stamp[1:3]>`
- The last 5 characters are unique for every pilots

Therefore, the partitioning was better before but we are now able to retrieve the output of a submission using a deterministic path rather than executing `find`. 

I also modified the `PilotManagerHandler` to add the pilot stamp to the pilot reference to retrieve the logging info in the same way.

Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -51,15 +53,63 @@
...
 
+@deprecated("Please use condorIDAndPathToResultFromJobRef")
 def condorIDFromJobRef(jobRef):
-  """return tuple of "jobURL" and condorID from the jobRef string"""
-  jobURL = jobRef.split(":::")[0]
-  condorID = jobURL.split("/")[-1]
+  """
+  Extract tuple of jobURL and jobID from the jobRef string.
+
+  :params jobRef: PilotJobReference of the following form: htcondorce://<ceName>/<pathToResult>-<condorID>
```
andresailer: This has to be `:param ` instead of `:params ` (and below)
Then simple types can be put as the second argument for `:param` instead of an extra line of `:type`
And it should be `str`, which is the type, instead of `string` which is a module.

aldbr:
Thanks for the review.
I made the changes, is that better now?

Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -193,11 +246,17 @@ def submitJob(self, executableFile, proxy, numberOfJobs=1):
...
+    # Then they also have an individual part to make them unique
     jobStamps = []
+    commonJobStampPart = makeGuid()[:3]
     for _i in range(numberOfJobs):
-      jobStamps.append(makeGuid()[:8])
+      jobStamp = commonJobStampPart + makeGuid()[:5]
+      jobStamps.append(jobStamp)
+
+    # We randomize the location of the pilotoutput and log, because there are just too many of them
+    location = __logDir(self.ceName, commonJobStampPart)
```
atsareg: self.__logDir(...

PR #4361, opened 2019-12-04
[v7r0] Core: restore the possibility to set a global timeout for a Client

chaen:
This fix https://github.com/DIRACGrid/DIRAC/pull/4340/commits/6b7732eb14abbef6b1be88b52658405f748b1f58 prevented setting a global timeout, while this is needed in the LHCb extension

This PR restore the possibility to do so.

Now, looking in detail about it, I realized something quite nasty: the `self.__kwargs` attribute created in the `Client` object ripples down to the `BaseClient.kwargs` where it is actually modified. So it does not matter because:
* it only changes basic types, and not inner containers
* we plan on moving to HTTP anyway 
But still, I wanted to note it

BEGINRELEASENOTES

*Core
FIX: restore the possibility to set a global timeout for a Client
ENDRELEASENOTES

### DIFF
--- a/Core/Base/Client.py
+++ b/Core/Base/Client.py
@@ -71,18 +71,18 @@ class Client(object):
         parameters of BaseClient are exposed.
 
         :param rpc: if an RPC client is passed, use that one
-        :param timeout: we can change the timeout on a per call bases. Default 120 s
+        :param timeout: we can change the timeout on a per call bases. Default is self.timeout
         :param url: We can specify which url to use
     """
     toExecute = self.call
     # Check whether 'rpc' keyword is specified
     rpc = kws.pop('rpc', False)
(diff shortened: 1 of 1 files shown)


andresailer:
When service functions are called through
`executeRPC`, that function's `timeout` value is used to call `_getRPC`
https://github.com/DIRACGrid/DIRAC/blob/46bd152530f2c0d24a249397f08bae028eeac8f4/Core/Base/Client.py#L81

(This path is used for all functions that are not explicitly created on the client, and also for the functions added by the `createClient` decorator)

So adding this would give a real global timeout setting
```diff
-timeout = kws.pop('timeout', 120) 
+timeout = kws.pop('timeout', self.timeout)
```
though that changes these timeouts from 120 to 600 by default

PR #4364, opened 2019-12-10
[v7r1] DFC: remove SEPrefix, VOPath, VOPrefix, ResolvePFN and SEDefinition from the DFC

chaen:
The SE management has nothing to do in the DFC, as discussed already a long time.
So I remove it, and do what I said in the ticket https://github.com/DIRACGrid/DIRAC/issues/4227

BEGINRELEASENOTES

*DMS
CHANGE: remove all the SE mangling in the DFC (SEPrefix, resolvePFN, SEDefinition)

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/DB/FileCatalogComponents/DirectoryManager/DirectoryTreeBase.py
+++ b/DataManagementSystem/DB/FileCatalogComponents/DirectoryManager/DirectoryTreeBase.py
@@ -718,16 +718,7 @@ class DirectoryTreeBase(object):
       for fileName in fileDict:
         successful[path][fileName] = fileDict[fileName]
 
-    result = S_OK({'Successful': successful, 'Failed': failed})
-
-    if self.db.lfnPfnConvention:
-      sePrefixDict = {}
-      resSE = self.db.seManager.getSEPrefixes()
-      if resSE['OK']:
-        sePrefixDict = resSE['Value']
-      result['Value']['SEPrefixes'] = sePrefixDict
-
-    return result
+    return S_OK({'Successful': successful, 'Failed': failed})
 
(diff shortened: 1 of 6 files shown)


atsareg:
This is fine. Just to be checked that we are nowhere relying on the availability of the PFN in the output of the FC.getReplicas() call

chaen:
the PFN stored in the catalog will be left untouched so it's still there if needed

marianne013:
I'm trying to understand the logic, but can't quite work it out yet.
How is the VOPath handled now ?

chaen:
It is still handled in the `StorageElement` as it was https://github.com/DIRACGrid/DIRAC/blob/integration/Resources/Storage/StorageFactory.py#L367
That means if you instantiate a `StorageElement` where the `VOPath` is configured, the `VOPath` will be used in place of `Path`.
What changes in this PR is that none of this mangling is done anymore in the `FileCatalog`, because it should never have been done in a first place.

marianne013:
It's probably OK, though I don't have any time to test it right now.

atsareg:
The output of the FileCatalog CLI "replicas" command will not print out URLs of the file replicas, it will just print out the SE name and LFN instead of a URL. Is it acceptable ? In principle in most cases it is enough to know in which SE the file resides. And also just one possible URL was printed out. So, it looks OK, may be others have a different idea ?

chaen:
I think it will print the URL that the catalog stores. At least that what it should be doing. 
Again, if mangling with path is needed, the client should use the `StorageElement` object, or even better, the `DataManager` class. 
This btw is inline with https://github.com/DIRACGrid/DIRAC/issues/1335

PR #4402, opened 2020-01-10
[v7r0] Docs: GITLABTOKEN not mandatory to generate changelog

chaen:
BEGINRELEASENOTES
*Docs
CHANGE: dirac-docs-get-release-notes.py does not require a GITLABTOKEN
ENDRELEASENOTES


andresailer:
Now that you mention it:
Actually, the token is also not needed for github public repositories, but the rate of API requests is limited, and that limit is much smaller for unauthenticated users.

Though, at least the cern gitlab does not do rate limiting at the moment, as far as I can tell by not seeing the relevant headers in the API returns.

To give a hint, please add
```python
else:
  LOGGER.info('Accessing GitLab unauthenticated')
```

One could also think about adding a "NoAuth" option and even skip the need for the GitTokens.py file.

PR #4413, opened 2020-01-17
[v6r22] FTS3: change proxy lifetime from 2 to 12 hours

chaen:
Operationally, we noticed that when having a low FTS activity, the proxy was expiring too soon, resulting in failed transfers.

BEGINRELEASENOTES

*DMS
CHANGE: change proxy lifetime of FTS3Agent from 2 to 12 hours

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Agent/FTS3Agent.py
+++ b/DataManagementSystem/Agent/FTS3Agent.py
@@ -44,6 +44,9 @@ from DIRAC.DataManagementSystem.Utilities.DMSHelpers import DMSHelpers
 
 AGENT_NAME = "DataManagement/FTS3Agent"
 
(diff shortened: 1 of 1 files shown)


DataManagementSystem/Agent/FTS3Agent.py
```diff
@@ -44,6 +44,9 @@
 
 AGENT_NAME = "DataManagement/FTS3Agent"
 
+# Lifetime in seconds of the proxy we download for submission
+PROXY_LIFETIME = 432000  # 12 hours
```
andresailer: This is 120 hours, isn't it?

chaen: Oups indeed ! Well spotted, I'll change

PR #4443, opened 2020-02-10
[v6r22] Read ES IndexPrefix from CS, if none given use setup name

petricm:
BEGINRELEASENOTES
*MonitoringSystem
CHANGE: Read the IndexPrefix for the CS and use this index prefix when creating ES indices, if not given use the name of the setup

ENDRELEASENOTES

This is needed or us, since we have an ES instance es-ilcdirac with IT, but our setup name in DIRAC is ilc. The problem is that IT wants that the index prefix matches instance name.

### DIFF
--- a/MonitoringSystem/DB/MonitoringDB.py
+++ b/MonitoringSystem/DB/MonitoringDB.py
@@ -10,6 +10,8 @@ from DIRAC import S_OK, S_ERROR, gLogger
 from DIRAC.Core.Base.ElasticDB import ElasticDB
 from DIRAC.Core.Utilities.ElasticSearchDB import generateFullIndexName
 from DIRAC.ConfigurationSystem.Client.Helpers import CSGlobals
+from DIRAC.ConfigurationSystem.Client.Config import gConfig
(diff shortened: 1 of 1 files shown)


MonitoringSystem/DB/MonitoringDB.py
```diff
@@ -20,7 +22,9 @@ class MonitoringDB(ElasticDB):
   """
 
   def __init__(self, name='Monitoring/MonitoringDB', readOnly=False):
-    super(MonitoringDB, self).__init__('MonitoringDB', name, CSGlobals.getSetup().lower())
+    section = getDatabaseSection("Monitoring/MonitoringDB")
+    indexPrefix = gConfig.getValue("%s/IndexPrefix" % section, CSGlobals.getSetup()).lower()
```
andresailer: Maybe mention the option in the module docstring. Similar to: https://github.com/DIRACGrid/DIRAC/pull/4367/files#diff-7ebce825272731fbe4f524930a420d7eR6

petricm:
Added the text the [wiki](https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-v7r0) and added the docs to the file.

PR #4466, opened 2020-02-26
[v7r0]Stomp: fixes the wild reconnections

chaen:
Addresses https://github.com/DIRACGrid/DIRAC/issues/4422

I am wondering whether I should add a few more layers of abstractions for the MQ....

BEGINRELEASENOTES

*Resources
FIX: Stomp reconnection only reconnect the connection that dropped

ENDRELEASENOTES


chaen:
@wkrzemien please check that.
I've run it in prod for a few days in LHCb

wkrzemien:
ok.

PR #4504, opened 2020-03-20
[v7r0] Tests: force the DataManager test setup to use the DFC

chaen:
probably not worth it

### DIFF
--- a/tests/Integration/all_integration_client_tests.sh
+++ b/tests/Integration/all_integration_client_tests.sh
@@ -83,7 +83,9 @@ exit
 
 EOF
 
-dirac-dms-filecatalog-cli < dataManager_create_folders
+# the filecatalog-cli script sorts alphabetically all the defined catalog and takes the first one....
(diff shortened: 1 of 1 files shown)


tests/Integration/all_integration_client_tests.sh
```diff
@@ -83,7 +83,9 @@ exit
 
 EOF
 
-dirac-dms-filecatalog-cli < dataManager_create_folders
+# the filecatalog-cli script sorts alphabetically all the defined catalog and takes the first one....
+# which of course does not work if your catalog is called Bokkkeeping... so force to use the real DFC
```
andresailer:
```suggestion
# which of course does not work if your catalog is called Bookkeeping... so force to use the real DFC
```

chaen: Thanks, applied

andresailer: too many K's still

PR #4521, opened 2020-04-01
[integration] Add Python 3 migration docs

chrisburr:


docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+
+
+Recommendations for code
+------------------------
+
+The "The Conservative Python 3 Porting Guide" linked above is an excellent source of information.
+This sections contains some details that are particularly relevant to DIRAC.
+
+**bytes vs str**
+  The most difficult change change when moving Python 3 is the splitting of the `str` type one for text and one for true binary data.
```
atsareg, line 70: change change -> change

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+  This exposes subtle issues in Python 2 that were likely never noticed and an automatic conversion to fix the is inherently impossible.
```
atsareg, line 71: the -> this

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+RedHat will continue to provide support of their CentOS 8 RPM until June 2024 however the maintenance burden of using Python 2.7 is rapidly increasing as libraries drop support.
+This also complicates any migration to Python 3 as it becomes necessary to make major updates to dependencies at the same time.
+In addition, DIRACOS has two dependencies which might cause issues as a result of being deprecated:
+
+ - CentOS 6 source RPMs are used and these will stop being supported in November 2020.
+ - The maintainers of the Python Package Index and pip have said they will drop support if either:
+     - bugs in Python 2.7 itself make this necessary (which is unlikely)
+     - Python 2 usage reduces to a level where pip maintainers feel it is OK to drop support
+
+While neither of these are likely to affect DIRAC in the immediate future, DIRAC has a relatively slow release cycle and therefore Python 2 will still be used for several years after the migration begins.
```
chaen, line 17: relatively -> extremely ? :-p

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+
+In the case of DIRAC, it is also necessary to move to using JSON for serialising the messages sent between between servers.
+This will be available as a technology preview in v7r2.
+
+DIRACOS
+^^^^^^^
+
+Once it becomes possible to run some DIRAC services with a client that uses Python 3 a suitable DIRACOS release will be needed.
+This will be DIRACOS version 2 and the ``--dirac-os-version=v2rX`` flag to `dirac-install` becomes the way to create a Python 3 based DIRAC installation.
+The will be based upon `conda-forge <https://conda-forge.org/>`_ and `conda-pack <https://conda.github.io/conda-pack/>`_ and provides several benefits over DIRACOS version 1 will maintaining bit-for-bit reproducibility:
```
chaen, line 37: The will -> this will
will maintaining -> while maintaining ?

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+
+    # Good: Prefix the "\n" to make it a bytes object
+    with open("my_file.txt", "rb") as fp:
+       data = fp.read().split(b"\n")
+
+**Dictionaries**
+  In Python 3 ``my_dict.keys()``, ``my_dict.values()`` and ``my_dict.items()`` now return an iterator instead of a list.
+  This is equivalent to ``my_dict.iterkeys()``, ``my_dict.itervalues()`` and ``my_dict.iteritems()`` in Python 2 and these methods have been removed.
+
+  In almost all cases ``my_dict``, ``my_dict.values()`` and ``my_dict.items()`` should be preferred.
```
chaen, line 131: `my_dict.keys()` instead of `my_dict` ?

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+  The is a small overhead in Python 2 when using ``items()`` instead of ``iteritems()`` however this is only applicable when dealing with large dictionaries in tight loops and such code can likely be written as a faster alternative (``six`` provides functions like ``six.iteritems(my_dict)`` if absolutely necessary).
+
+  In rare cases the list object returned might be desirable, if so ``list(my_dict.items())`` can be used.
```
chaen, line 134: so basically you are saying we have to bare the cost of a double copy as long as we use python 2 ?

chrisburr: Does the cost actually matter? If it does, then fair enough and we can litter the code with `six.iterxxxx` but I doubt the cost of making the code less readable is worth it in most places.

docs/source/DeveloperGuide/Python3Migration/index.rst
chrisburr: When should `my_dict.keys()` be used? It's normally better to just use `my_dict`:

```python
lfns = list(lfns_with_metadata)
if lfn in lfns_with_metadata:
    pass
```

instead of:

```python
lfns = list(lfns_with_metadata.keys())
if lfn in lfns_with_metadata.keys():
    pass
```

docs/source/DeveloperGuide/Python3Migration/index.rst
atsareg: I thought that the solution was to use iteritems() everywhere in Python2 and list(obj.iteritems()) if the list is needed. When moving to Python3 we will globally substitute iteritems() by items().

chrisburr: There is no benefit to using `list(obj.iteritems())` instead of `six.iteritems(obj)`.

As I state in the text of this PR, that makes it impossible to have a code base that supports both so we can gradually test services with Python 3 and fix bugs as we find them. I've never read about a project successfully migrating in one shot, there are just too many possible bugs.

docs/source/DeveloperGuide/Python3Migration/index.rst
chaen: Sorry, went too fast, obviously :-) disregard

docs/source/DeveloperGuide/Python3Migration/index.rst
chaen: I think our original idea was to make things explicit when we want a copy. So the only way to achieve it which would be py2 and py3 compatible would be `items` when we do not need a copy, and `list(dict.items())` when we need one.

chrisburr: Sorry I've misread your original comment as "bare the cost of a copy" instead of "bare the cost of a **double** copy".

I've tried to benchmark it and of course there is a cost but it's less than I would have naively expected. If there are a few hot spots where we have particularly big objects we could use `six.iteritems(list(obj))`.

Python 3 `list(x.items())` is notably faster than Python 2 `x.items()` so if there is a service where it matters it's a reason to move that service to Python 3 (or even PyPy which is a LOT faster).

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,227 @@
...
+  .. code-block:: python
+
+    # Bad: Types should be check using isinstance
+    if type(my_variable) == str:
+
+    # Bad: basestring does not exist in Python 3
+    if instance(my_variable, basestring):
+
+    # Good: Supports both Python 2 and 3
+    if instance(my_variable, six.string_types):
```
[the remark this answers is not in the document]
chrisburr: It doesn't:

```
In [1]: a = u'hello world'
In [2]: isinstance(a, str)
Out[2]: False
```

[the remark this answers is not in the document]
zmathe: Yes, in python2 does not work in some cases. But when we know a is not unicode we should not use six.string_types. No? I had no problem to make Graph package work with python2 and python3 https://github.com/DIRACGrid/DIRAC/pull/4497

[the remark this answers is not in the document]
chrisburr: I think the recommendation should be to use `six.string_types` in all situations to avoid any possible mistakes, especially as it doesn't have an obvious failure mode.

I don't see any downside and it's easy to refactor away once we drop Python 2 for good.

fstagni:
I have only 2 questions/considerations:

- does this need to be part of the DIRAC documentation? Personally, I think it shouldn't. IMHO this is the perfect case for a "Wiki" style doc.
- The excellent [Recommendations](https://github.com/DIRACGrid/DIRAC/pull/4521/files#diff-8fbff0b529227d93dafe6803fb288704R64) should IMHO become issues. E.g. already all the usages of `type` for type checks have been moved to `isinstance` in the _integration_ branch (but `basestring` is still used in many places).

andresailer:
I think it is good to have this as part of the documentation, similar to the coding conventions, having everything in one place makes it easier to find. Also let's us have a review and discussion process if someone wants to propose a different approach, which we already have in this PR.

chrisburr:
1. I'm inclined to agree with @andresailer about having this in the docs instead of a wiki though if we decide otherwise I'll keep it here for review then copy it over.

2. Yes, a lot of issues will come out of this PR once it's settled. Not only for code changes but also for things like DIRACOS v2 and the CI changes.

petricm:
I think some information about division should be also added.

docs/source/DeveloperGuide/Python3Migration/index.rst
```diff
@@ -0,0 +1,255 @@
...
+    with open("my_file.txt", "rt") as fp:
+       data = fp.read().split("\n")
+
+    # Bad: Fails on python 3 as "\n" is a string not bytes
+    with open("my_file.txt", "rb") as fp:
+       data = fp.read().split("\n")
+
+    # Good: Prefix the "\n" to make it a bytes object
+    with open("my_file.txt", "rb") as fp:
+       data = fp.read().split(b"\n")
```
petricm, line 152: What about a recommendation that one should rather use `io.open` instead of the built-in `open`?

chrisburr:
I think this is ready for merging now. I've just added a tiny bit more detail on `io.open` and strings vs bytes (881978e).

fstagni (approved):

chrisburr:
@atsareg I think this PR got missed when merging yesterday? As it's only docs so I see no reason to wait for the next set of patches.

atsareg:
You are right, I will add it

PR #4527, opened 2020-04-06
[v7r0] URGENT: Fix DIRACOS CI

chrisburr:
The DIRAC OS CI is currently broken as `$DIRACOS_TARBALL_PATH` is set to `/`. Unsurprisingly copying the entire system from the host to the container causes the container to get stuck until the job times out.

Instead, check if the folder actually contains a `diracos-*.tar.gz` file.

### DIFF
--- a/tests/CI/run_docker_setup.sh
+++ b/tests/CI/run_docker_setup.sh
@@ -188,12 +188,11 @@ function prepareEnvironment() {
   docker cp "$SERVERCONFIG" server:"$WORKSPACE/CONFIG"
   docker cp "$CLIENTCONFIG" client:"$WORKSPACE/CONFIG"
 
-  # if the DIRACOS_TARBALL_PATH is a local directory, copy it over
-  if [ -d "${DIRACOS_TARBALL_PATH}" ]; then
+  # Copy DIRACOS_TARBALL_PATH if it is a local directory containing a DIRACOS tarball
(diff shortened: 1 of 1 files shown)


petricm:
Please merge ASAP

tests/CI/run_docker_setup.sh
```diff
@@ -188,12 +188,11 @@ function prepareEnvironment() {
   docker cp "$SERVERCONFIG" server:"$WORKSPACE/CONFIG"
   docker cp "$CLIENTCONFIG" client:"$WORKSPACE/CONFIG"
 
-  # if the DIRACOS_TARBALL_PATH is a local directory, copy it over
-  if [ -d "${DIRACOS_TARBALL_PATH}" ]; then
+  # Copy DIRACOS_TARBALL_PATH if it is a local directory containing a DIRACOS tarball
+  if ls "${DIRACOS_TARBALL_PATH}"/diracos-*.tar.gz 1> /dev/null 2>&1; then
```
chaen, line 192: is the md5 file mandatory or not ? If yes, we may as well check it here too ?

chrisburr: It technically is so far as I know but I don't see the point in checking it here. It will fail soon after with an obvious error message when `dirac-install` is ran.

It would be more confusing if it ignores your custom DIRACOS build because you forgot the MD5 file (or if `dirac-install` is modified to support a different checksum and the check here wasn't updated to match).

chaen: Fine with me

PR #4528, opened 2020-04-06
[v7r0] Sleep and retry if the FC is unavailable when uploading output

chaen:
We have a design flaw which is that, when uploading the output of a job, if it fails, we just upload it somewhere else and create a failover request. However, in the case where the DFC is totally down, you can't even reach this state, because you cannot perform the `hasAccess` operation. 
In practice, this results in losing the result of your job. 
This PR introduces a sleep & retry in case this happens
@fstagni this is what we discussed

BEGINRELEASENOTES
*DMS
CHANGE: failoverTransfer sleep & retries in case of FC unavailability

ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Client/FailoverTransfer.py
+++ b/DataManagementSystem/Client/FailoverTransfer.py
@@ -19,9 +19,12 @@
 
 __RCSID__ = "$Id$"
 
+import time
+
 from DIRAC import S_OK, S_ERROR, gLogger
 
 from DIRAC.Core.Utilities.ReturnValues import returnSingleResult
+from DIRAC.Core.Utilities.DErrno import cmpError, EFCERR
 from DIRAC.DataManagementSystem.Client.DataManager import DataManager
 from DIRAC.DataManagementSystem.Utilities.DMSHelpers import DMSHelpers
 from DIRAC.RequestManagementSystem.Client.Request import Request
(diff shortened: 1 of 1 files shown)


DataManagementSystem/Client/FailoverTransfer.py
```diff
@@ -72,18 +75,33 @@ def transferAndRegisterFile(self,
...
+                                                                guid=fileGUID,
+                                                                checksum=fileChecksum)
+        self.log.verbose(result)
+
+        # If the FC is unavailable, we stay in the loop and retry
+        # otherwise we continue without retrying
+        if result['OK'] or not cmpError(result, EFCERR):
+          break
+        self.log.error("transferAndRegisterFile: FC unavailable, retry")
+        time.sleep(sleeptime)
```
andresailer, line 103: Is it necessary to disable the JobWrapper Watchdog CPU consumption monitoring here?

chaen: I must admit being totally unknowledgeable here, but I was under the impression that the download/upload were always done with the Watchdog disabled ?

andresailer: Indeed.
https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/blob/master/LHCbDIRAC/Workflow/Modules/UploadOutputData.py#L232-233

https://gitlab.cern.ch/CLICdp/iLCDirac/ILCDIRAC/-/blob/Rel-v29r2/Workflow/Modules/UploadOutputData.py#L180-183

But not in DIRAC itself, as far as I can tell.

chaen: ah ok. Well if it isn't disabled before, I don't see why i should be now. Also, this should be an extremely rare case !

andresailer: Yes, agreed. The disabling needs to happen at a higher level than this function here.

PR #4565, opened 2020-04-23
[rel-v7r1] Source bashrc in wrapper script

chrisburr:
BEGINRELEASENOTES

*Core
CHANGE: Source bashrc in wrapper script

ENDRELEASENOTES

As mentioned in #4480


atsareg:
Integration test segfaulted in the Test_DIRACCAProxyProvider. This was not the case before and is not likely to be provoked by this PR. Merging ?

chaen:
I would not. Sourcing the bashrc means different environment, and if there are conflicts, it could result in segfault. So I would not ignore the tests

atsareg:
An integration test failed again but now this is another integration test. The one failing before is OK now. Merging ?

atsareg:
After four attempts the test is finally green !

PR #4581, opened 2020-05-05
[v7r2] Use diraccfg module instead of DIRAC.Core.Utilities.CFG

chrisburr:
BEGINRELEASENOTES

*Core
CHANGE: Replace DIRAC.Core.Utilities.CFG with independent diraccfg module

ENDRELEASENOTES


requirements.txt
```diff
@@ -9,6 +9,7 @@ autopep8==1.3.3
 certifi
 coverage
 docutils
+diraccfg
```
andresailer, line 12: Also need to add this to docs/requirements.txt

PR #4582, opened 2020-05-05
[v7r2] use concurrent futures by default

zmathe:
BEGINRELEASENOTES

Please follow the template:
*Core
CHANGE: Use concurrent.future.ThreadPoolEcecutor by default.

ENDRELEASENOTES


chrisburr:
The integration test should be changed so it runs tests with this disabled (i.e. [this line](https://github.com/DIRACGrid/DIRAC/blob/69101f9b1bdd79399205a7df0920154428b79ec3/.github/workflows/integration.yml#L31)).

zmathe:
@chrisburr right. I changed the value.

PR #4585, opened 2020-05-05
[v7r0] Fail-fast for pipelines running against outdated commits

chrisburr:
As pull requests tend to be merged in batches it's easy to overload the CI. This PR adds a step to check if the current pipeline is running against an outdated commit. If this is the case, it fails early rather than wasting CI resources.

It might be useful to run this script between each stage of the integration tests but I'll only do that if it continues to be a problem.

### DIFF
--- a/.github/workflows/fail-fast.sh
+++ b/.github/workflows/fail-fast.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -euo pipefail
+IFS=$'\n\t'
+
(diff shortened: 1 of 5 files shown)


.github/workflows/fail-fast.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -euo pipefail
+IFS=$'\n\t'
+
+if [[ "$(jq --raw-output '.pull_request' "${GITHUB_EVENT_PATH}")" == "null" ]]; then
```
chaen, line 6: care to add just a few comments ? :D Reading this I understand the idea, but I have absolutely no idea what is actually being done, in particular in the tests :D

PR #4586, opened 2020-05-05
[v7r2] Add __future__ import for Python 3 like imports, division, and printing

chrisburr:
This change should be mostly harmless. If it causes any issues that in code that isn't covered by the automated tests they should be easy to debug and fix.

BEGINRELEASENOTES

*All
CHANGE: Use \_\_future\_\_ imports to get Python 3 like behaviour for imports, division and printing 

ENDRELEASENOTES


petricm:
Are you sure that `division` in all files is really harmless?

chrisburr:
> Are you sure that `division` in all files is really harmless?

No but the tests still pass and I've grepped through for then obvious failure patterns I can think of and haven't found anything.

I don't really know what else we can then do other than apply it and then check more with certifications. I'm fairly confident any errors will expose themselves in an obvious and easy-to-hotfix way so doing so I think it will be more efficient to just bite the bullet.

docs/diracdoctools/Config.py
```diff
@@ -1,4 +1,7 @@
 """Configuration for the documentation scripts."""
+from __future__ import absolute_import
+from __future__ import division
```
andresailer, line 3: This and other changes in diracdoctools are going to conflict with #4571

chrisburr: Yes, I avoided the Pilot 2 files (#4162) but any other conflicts will be easy to fix.

When #4571 is merged and propagated I'll rebase this PR.

andresailer: Ok, excellent!

fstagni:
image
Do you really think I'll open that tab? 

Also, consider that there is also https://github.com/DIRACGrid/DIRAC/pull/4145 (which should be rebased...)

fstagni:
Regarding divisions: we had a long discussion not long ago, when in PR https://github.com/DIRACGrid/DIRAC/pull/4189 I introduced a few. It was later decided to remove the changes affecting divisions. I would NOT introduce them now, not on this large scale.

chrisburr:
All this PR does is ensure every file starts with:
```python
```
And a handful of minor tests where the test is the problem not the code. That can be reviewed as an individual commit: 2fe8d18

The difference with #4189 is that that PR changed lots of divisions to be integer division when they shouldn't be, i.e. `a / b` was changed to `a // b` ([example](https://github.com/DIRACGrid/DIRAC/pull/4189#discussion_r309367190)). This PR can only really cause issues from floats getting into places they shouldn't like:

```
In [1]: (10 / 2) * "#"
Out[1]: '#####'

In [2]: from __future__ import division

In [3]: (10 / 2) * "#"
<ipython-input-3-528334909078> in <module>()
----> 1 (10 / 2) * "#"

TypeError: can't multiply sequence by non-int of type 'float'

In [4]: (10 // 2) * "#"
Out[4]: '#####'
```

DataManagementSystem/Client/CmdDirCompletion/DirectoryCompletion.py
```diff
@@ -84,7 +84,7 @@ def get_filename(self, path, dirname):
 
 
 if __name__ == "__main__":
-  from AbstractFileSystem import UnixLikeFileSystem
+  from AbstractFileSystem import UnixLikeFileSystem  # pylint: disable=import-error
```
andresailer, line 87: shouldn't this be
```suggestion
  from .AbstractFileSystem import UnixLikeFileSystem
```
I.e. explicitly importing from the neighbor module AbstractFileSystem?

chrisburr: No as it's called as a script with something like `python DataManagementSystem/Client/CmdDirCompletion/DirectoryCompletion.py`. Relative imports only work inside modules and `pylint` isn't smart enough to figure it out. The same applies to `config` and `make_dfc_plots_lib` in the tests.

```bash
> ls -R
my-folder-1

./my-folder-1:
my-folder-2

./my-folder-1/my-folder-2:
A.py	B.py	C.py
> cat my-folder-1/my-folder-2/A.py
print("Import A")
import B

> cat my-folder-1/my-folder-2/B.py
print("Import B")
import B

> python2.7 my-folder-1/my-folder-2/A.py
Import A
Import B
> python3 my-folder-1/my-folder-2/A.py
Import A
Import B
> cat my-folder-1/my-folder-2/C.py
print("Import C")
from . import B

> python2.7 my-folder-1/my-folder-2/C.py
Import C
  File "my-folder-1/my-folder-2/C.py", line 3, in <module>
    from . import B
ValueError: Attempted relative import in non-package
```

andresailer: Yes, I see.
Thanks for the detailed explanation!

PR #4588, opened 2020-05-06
[v7r0] Fix HTCondor cleanup

aldbr:
Every pilots outputs, errors and logs from HTCondor are stored in a specific directory: `workingDirectory`.

To avoid to get too many files stored in this directory, files are regularly deleted.
The number of days we want to keep these files in configurable in each CE section. 
However in reality, as the `workingDirectory` is the same for every CE, each of them is able to delete the pilots files of other CEs which can create issues to retrieve the pilot outputs:

```
WorkloadManagement/SiteDirector INFO: Retrieving output for pilot htcondorce://ce/1234.56
WorkloadManagement/SiteDirector ERROR: Failed to get pilot output ce: Failed to find condor job 1234.56
``` 

Since `v7r0`, the pilot files of a same CE are stored in `workinDirectory/<ceName>`. Thus, a given CE has the possibility to only delete its files and not the others.

The counterpart is that, when updating from `v6` to `v7`, pilots files stored in the old structure of directory would need to be manually deleted. 

BEGINRELEASENOTES
*Resources
FIX: HTCondorCE cleanup only the pilots files related to the CE 

ENDRELEASENOTES

### DIFF
--- a/Resources/Computing/HTCondorCEComputingElement.py
+++ b/Resources/Computing/HTCondorCEComputingElement.py
@@ -488,8 +488,9 @@ Queue %(nJobs)s
     if status:
       self.log.error("Failure during HTCondorCE __cleanup", stdout)
 
-    # remove all out/err/log files older than "DaysToKeepLogs" days
-    findPars = dict(workDir=self.workingDirectory, days=self.daysToKeepLogs)
+    # remove all out/err/log files older than "DaysToKeepLogs" days in the CE part of the working Directory
+    workDir = os.path.join(self.workingDirectory, self.ceName)
+    findPars = dict(workDir=workDir, days=self.daysToKeepLogs)
     # remove all out/err/log files older than "DaysToKeepLogs" days
     status, stdout = commands.getstatusoutput(
         r'find %(workDir)s -mtime +%(days)s -type f \( -name "*.out" -o -name "*.err" -o -name "*.log" \) -delete ' %


andresailer:
The code changes look fine, except for the issue you already write about when upgrading.

But, I don't understand how this is related to the error message you post, because that comes from this line
https://github.com/DIRACGrid/DIRAC/blob/74800ea2b7d9c4544008d42f352f1d6d9a707b6b/Resources/Computing/HTCondorCEComputingElement.py#L414
doesn't it?
And that would mean condor_q doesn't find the job ID, or the SUBMIT_Iwd property is not set or not set to the expected value. Which would be something the __cleanup function has no control over.

aldbr:
Yes, you're right.
Actually I observed two different issues, one seems bound to `cleanup`:
```
WorkloadManagement/SiteDirector/ce.a.ch ERROR: Failure during HTCondorCE __cleanup 
find: cannot delete ‘/workingDirectory/ce.b.fr/5/43/12345.5.err’: No such file or directory
find: cannot delete ‘/workingDirectory/ce.b.fr/5/43/23456.7.err’: No such file or directory
```
The change presented in this PR should resolve this one.
I thought the second issue, presented above, was also related to that but you're right, it is something else...

aldbr:
Well, in this case, I don't think this particular error has an important impact on the way the CE works and I think the fix could wait so that the pilot files from the old structure would not need to be manually deleted

PR #4592, opened 2020-05-10
[v7r0] Remove direct access to SandboxMetadataDB in the SandboxStoreClient

atsareg:
SandboxStoreClient has got a problem with some recent update of the DIRACOS when MySQLdb module was added: it tries to import it and, if successful, it attempts to connect to the SandboxMetadataDB. If the db access is impossible, the connection takes several minutes to return with error and then the client proceeds with the service access. As a result suddenly the job submission with the command line became terribly slow. Looking at the code of the SandboxStoreClient, methods that are using direct db access are actually not used anywhere. Therefore, the direct db access and the methods that were using it are dropped in this PR. 

BEGINRELEASENOTES

*WorkloadManagement
CHANGE: SandboxStoreClient - remove direct access to SandboxMetadataDB

ENDRELEASENOTES

### DIFF
--- a/Resources/Computing/ARCComputingElement.py
+++ b/Resources/Computing/ARCComputingElement.py
@@ -94,7 +94,7 @@ class ARCComputingElement(ComputingElement):
     # For the XRSL additional string from configuration - only done at initialisation time
     # If this string changes, the corresponding (ARC) site directors have to be restarted
     #
-    # Variable = XRSLExtraString (or for multi processor mode)
+    # Variable = XRSLExtraString (or XRSLMPExtraString for multi processor mode)
     # Default value = ''
     #   If you give a value, I think it should be of the form
     #          (aaa = "xxx")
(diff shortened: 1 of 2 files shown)


chaen:
This looks like a similar problem as https://github.com/DIRACGrid/DIRAC/pull/4214

phicharp:
PR #4214 was removing the direct DB access by default, thus why was there a problem still?

atsareg:
The problem is that Dirac.submitJob() uses WMSClient which in turn instantiates SandboxStoreClient without smdb=False. Then submitJob() takes ages to execute. 

There is one place where the SandboxStoreClient is supposed to be used with direct access to the database - JobSanity executor, this one I have overlooked. If we have to keep direct db access inside the SandboxStoreClient, I would put smdb=False by default and JobSanity would explicitely set it to None.

atsareg:
Replaced by #4594

phicharp:
Sorry @atsareg, my mistake... I thought the default was False but indeed it is None which keeps the check...
I overlooked myself the fact that the WMSClient was instantiating `SandboxStoreClient`, otherwise I would have set it to False there, sorry!

PR #4593, opened 2020-05-11
[v7r0] Add a new SiteDirector option to stop trying to get pilot outputs endlessly

aldbr:
+ When the option "GetPilotOutput" of a Site Director agent is enabled, this one tries to get the pilot outputs of the sites it manages, even if they are old.
Old pilot outputs may be completely deleted and may not be retrieved.
Thus, the Site Director endlessly tries to find pilot outputs that don't exist anymore and this operation may take a long time...

To prevent the Site Director continuing to search for an old pilot output that doesn't exist anymore, I just give it a pilot output with the following content: 
"Pilot output is no longer available"

Is this okay? 
I noticed that in `v7r0` but I think this is older than that, which version should I target?   

+ And a small improvement to retrieve some pilot outputs in HTCondorCEs using the fact that paths to the pilot outputs are now deterministic. 

BEGINRELEASENOTES
*WorkloadManagement
NEW: option to only retry to get pilot outputs a limited number of times
*Resources
ADD: retrieve pilot output paths in getJobOutput using the pilot IDs instead of calling condor
ENDRELEASENOTES

### DIFF
--- a/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -97,6 +97,8 @@ class SiteDirector(AgentModule):
     self.queueCECache = {}
     self.queueSlots = {}
     self.failedQueues = defaultdict(int)
+    # failedPilotOutput stores the number of times the Site Director failed to get a given pilot output
+    self.failedPilotOutput = defaultdict(int)
     self.firstPass = True
     self.maxJobsInFillMode = MAX_JOBS_IN_FILLMODE
     self.maxPilotsToSubmit = MAX_PILOTS_TO_SUBMIT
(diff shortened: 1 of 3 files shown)


atsareg:
Looks OK to me but fix the errors in the tests

andresailer:
The changes for HTCondorCE can only go to v7r0.
Also please rebase to get the latest changes to that file.

WorkloadManagementSystem/Agent/SiteDirector.py
```diff
@@ -97,6 +97,8 @@ def __init__(self, *args, **kwargs):
     self.queueCECache = {}
     self.queueSlots = {}
     self.failedQueues = defaultdict(int)
+    # failedPilotOutput stores the number of times the Site Director failed to get a given pilot output
+    self.failedPilotOutput = defaultdict(int)
```
andresailer, line 101: I know the dict gets cleared at every restart, but should there be `pop`s when the limit or success is reached? Unlike the queues, the number of pilots is potentially growing rapidly.

PR #4597, opened 2020-05-11
[integration] Use also time in addition to date to get list of PRs since last tag

petricm:
BEGINRELEASENOTES
*docs
FIX: Use also time in addition to date to get list of PRs since last tag

ENDRELEASENOTES
This is problem in diracos where you can have a pattern of merge tag merge in the same day and this causes problems. In addition the flag `--date` can now take any iso date as input.


andresailer:
Can you target rel-v7r0?

PR #4598, opened 2020-05-11
[v7r0] Use also time in addition to date to get list of PRs since last tag

petricm:
BEGINRELEASENOTES
*docs
FIX: Use also time in addition to date to get list of PRs since last tag

ENDRELEASENOTES
This is problem in diracos where you can have a pattern of merge tag merge in the same day and this causes problems. In addition the flag `--date` can now take any iso date as input.

### DIFF
--- a/docs/diracdoctools/scripts/dirac-docs-get-release-notes.py
+++ b/docs/diracdoctools/scripts/dirac-docs-get-release-notes.py
@@ -241,7 +241,9 @@ class GithubInterface(object):
                         help="branches to get release notes for")
 
     parser.add_argument("--date", action="store", default=self.startDate, dest="date",
-                        help="date after which PRs are checked, default (two weeks ago): %s" % self.startDate)
+                        help="date and optionally time after which PRs are checked (ISO 8601),\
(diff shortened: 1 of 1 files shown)


andresailer:
Thanks for changing the target branch, makes it easier to keep all the docs developments in sync.

Can you add an example of the date to the date CLI description, please?
Is the flag "sinceLatestTag" not suitable for your use case?

petricm:
The date works as before and also like
 - `2020-01-08`
 - `2020-01-08T15:08:00.000+01:00`
 -  `2018-05-20T15:23:45Z`
where `2020-01-08` is converted to `2020-01-08 00:00:00` in `UTC`

The previous comparison since last tag is based only on comparison of the date, thus you have a problem here
```py
if mergeDate[:10] < self.startDate:
```
if was merged on the tag date(`startDate`) it's added to the release notes, but it was most likely merged just before tagging, so you get the same PR info in two consecutive releases.

atsareg:
@andresailer : Approved ?

andresailer:
Yes, was just checking how the date help message looks when printed

PR #4599, opened 2020-05-12
[v7r0] Increase integration tests timeout to 45 minutes

chrisburr:
As the tests have been expanded we've started coming very close to the 30 minute timeout for each job. The majority of the time is spent in the server installation and optimising this would be worthwhile but for now I suggest increasing the timeout.


chrisburr:
The CI is failing because diracproject.web.cern.ch is down again

atsareg:
Amazing that the PR extending timeouts fails tests with ... timeouts :)

PR #4620, opened 2020-05-25
[v7r0] Fail HTCondorCE.getPilotOutput if workingDirectory not available or if condor_transfer_data fails

aldbr:
When multiple PilotManager services are installed on distinct machines and that the HTCondor working directory is not available (does not exist or permission denied), the command to get the pilot output fails silently.
This PR:
- adds a sentence in the documentation to inform admins that HTCondor working directory should be available on each and every machine hosting a PilotManager
- returns errors if the working directory is not available or if the command to get the pilot output fails
- also adds some suggestions made by @andresailer in a previous PR that were not merged:
https://github.com/DIRACGrid/DIRAC/pull/4593

BEGINRELEASENOTES
*WorkloadManagement
NEW: pop the pilot reference from the failedPilotOutput dictionary when maxRetryGetPilotOutput is reached
*Resources
FIX: in getPilotOutput, returns errors when the HTCondor working directory is not available and condor_transfer_data fails
ENDRELEASENOTES


andresailer:
Wrong target branch

Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -403,31 +403,39 @@ def getJobOutput(self, jobID, _localDir=None):
...
+        iwd = stdout_q
+
+      # Use the path extracted from the pilotID
       if iwd is None:
-        return S_ERROR("Failed to find condor job %s" % condorID)
+        iwd = os.path.join(self.workingDirectory, pathToResult)
 
       try:
-        os.makedirs(iwd)
+        mkDir(iwd)
```
andresailer, line 420: I think (or somewhere else in this block)
```suggestion
        mkDir(iwd)
        pathToResult = iwd
```
otherwise below the pathToResult from the pilot stamp is used, which might be different.

docs/source/AdministratorGuide/ServerInstallations/scalingAndLimitations.rst
```diff
@@ -138,7 +138,7 @@ Services
...
 +                    +---------------------------------------------------------------------------------------------------+-------------+-------------------------------------------------------------------------------------+
-|                    | :mod:`PilotManager <DIRAC.WorkloadManagementSystem.Service.PilotManagerHandler>`                  | YES         |                                                                                     |
+|                    | :mod:`PilotManager <DIRAC.WorkloadManagementSystem.Service.PilotManagerHandler>`                  | YES         | In case there are HTCondor CEs to deal with, the HTCondor "working directory" should exist and be accessible on each and every machine. The path is available in the configuration (/Resources/Computing/HTCondorCE/WorkingDirectory). The default one is hardcoded in the source code of the HTCondorCE itself.                                                                                      |
```
andresailer, line 141: This overlong line doesn't work. The table boundaries need to be all aligned and the same for all rows.
You have to break this into multiple lines

You could link to https://dirac.readthedocs.io/en/latest/AdministratorGuide/Configuration/ConfReference/Resources/Computing/index.html#htcondorce-parameters

Though maybe it would be better to replace YES with PARTIAL and have the explanation in the HTCondorCE documentation.

PR #4644, opened 2020-06-16
[v7r0] Cancel FTS3 Operation if the RMS request does not exist

chaen:
BEGINRELEASENOTES

*DMS
FIX: Cancel FTS3 Operation if the RMS request does not exist

*RMS
NEW: `getRequestStatus` returns `ENOENT` if the request does not exist
ENDRELEASENOTES


chaen:
Still this $#$##@$%#E PoolCE.

PR #4645, opened 2020-06-16
[v7r1] Few fixes for the integration tests on jenkins

chaen:
A whole bunch of small fixes. Including the controversial `readonly` variables. I really do not want to spend time debugging it, but as long as we have them, running the jenkins integartion tests are impossible as such. Up to everybody to measure pro/cons of having them readonly

BEGINRELEASENOTES

*Tests
FIX: dropDBs uses real array; 
FIX: remove readonly variables

*SMS
FIX: drop the tables in the correct order to avoid foreign key errors

ENDRELEASENOTES

### DIFF
--- a/StorageManagementSystem/DB/StorageManagementDB.sql
+++ b/StorageManagementSystem/DB/StorageManagementDB.sql
@@ -9,7 +9,11 @@
 
 use StorageManagementDB;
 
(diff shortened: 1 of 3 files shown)


chaen:
PoolCE....

PR #4651, opened 2020-06-18
[v7r0] add cfg option to commands to standardize behaviour

marianne013:
We propose this fix in response to user requirements to be able to upload/download/replicate .cfg files using the DIRAC command line tools. We are aware that there is a workaround, but it should just work properly to avoid the issue from resurfacing on a regular basis. 
We are aware that fully removing the non-option parsing of config files will require updates to other parts of the DIRAC code (pilot3). This would go into a later release (v7r4?).
BEGINRELEASENOTES
*Configuration
CHANGE: Add --cfg option for config files, deprecating old non-option version.
ENDRELEASENOTES

### DIFF
--- a/ConfigurationSystem/Client/LocalConfiguration.py
+++ b/ConfigurationSystem/Client/LocalConfiguration.py
@@ -98,6 +98,8 @@ class LocalConfiguration(object):
                         self.__setUseCertByCmd)
     self.registerCmdOpt("d", "debug", "Set debug mode (-ddd is extra debug)",
                         self.__setDebugMode)
+    self.registerCmdOpt("", "cfg=", "Load additional config file",
+                        None)
     devLoader = Devloader()
     if devLoader.enabled:
       self.registerCmdOpt("", "autoreload", "Automatically restart if there's any change in the module",
@@ -255,13 +257,29 @@ class LocalConfiguration(object):
       gLogger.fatal("Error when parsing command line arguments: %s" % str(x))
       self.showHelp(exitCode=2)
 
-    for o, _ in opts:
-      if o in ('-h', '--help'):
+    for opt, val in opts:
+      if opt in ('-h', '--help'):
         self.showHelp()
         sys.exit(2)
+      if opt == '--cfg':
+        self.cliAdditionalCFGFiles.append(val)
 
-    self.cliAdditionalCFGFiles = [arg for arg in args if arg[-4:] == ".cfg"]
-    self.commandArgList = [arg for arg in args if not arg[-4:] == ".cfg"]
+    # environment variable to ensure smooth transition
+    if os.getenv("DIRAC_NO_CFG", None):
+      self.commandArgList = args
+    else:
+      # to avoid issuing the warning for correctly passed cfg files
+      extraCfg = [arg for arg in args if arg.endswith(".cfg")]
+      self.cliAdditionalCFGFiles.extend(extraCfg)
+      self.commandArgList = [arg for arg in args if not arg.endswith(".cfg")]
+      if extraCfg:
+        # use error level to make sure users will always see the Warning
+        gLogger.error("""WARNING: Parsing of '.cfg' files as command line arguments is changing!
+          Set the environment variable 'export DIRAC_NO_CFG=1' to pass the file as a positional
+          argument (this will become the default).
+          To modify the local configuration use '--cfg <configfile>' instead.""")
+        if os.environ.get("DIRAC_DEPRECATED_FAIL", None):
+          raise NotImplementedError("ERROR: using deprecated config file passing option.")
     self.parsedOptionList = opts
     self.isParsed = True
 
--- a/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst
+++ b/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst
@@ -55,10 +55,18 @@ in the order of preference of the option resolution:
      dirac-wms-job-submit job.jdl -o /DIRAC/Setup=Dirac-Production
 
 *Command line argument specifying a CFG file*
-  If a filename with the *.cfg* extension is passed as an argument to any DIRAC command
-  it will be interpreted as a configuration file. For example::
+  A config file can be passed to any dirac command with the ``--cfg`` flag::
   
-     dirac-wms-job-submit job.jdl my.cfg
+     dirac-wms-job-submit job.jdl --cfg my.cfg
+
+  .. versionchanged:: v7r0
+
+     The passing of ``.cfg`` files was changed to require the ``--cfg`` flag.
+
+  .. deprecated:: v7r0
+
+     If a filename with the ``.cfg`` extension is passed as an argument to any DIRAC command
+     it will be interpreted as a configuration file, if the ``DIRAC_NO_CFG`` environment variable is not set.
 
 *Value of $DIRACSYSCONFIG environment variable*
   if the DIRACSYSCONFIG variable is set, it should point to a cfg file (written in *CFG* format)
--- a/docs/source/AdministratorGuide/ServerInstallations/environment_variable_configuration.rst
+++ b/docs/source/AdministratorGuide/ServerInstallations/environment_variable_configuration.rst
@@ -25,6 +25,9 @@ DIRAC_GFAL_GRIDFTP_SESSION_REUSE
 DIRAC_M2CRYPTO_SPLIT_HANDSHAKE
   If ``true`` or ``yes`` the SSL handshake is done in a new thread (default Yes)
 
+DIRAC_NO_CFG
+  If set to anything, cfg files on the command line must be passed to the command using the --cfg option.
+
 DIRAC_USE_M2CRYPTO
   If ``true`` or ``yes`` DIRAC uses m2crypto instead of pyGSI for handling certificates, proxies, etc.
 


chaen:
I agree that it would be better this way....
but there are quite a few places where we rely on that. For example, the `run` file which is generated for each components. So this should also be changed in the code. Also it should be updated by hand on existing installations. 
So it's probably a fix that needs to go through a certification procedure. 
You could also use the `deprecated` decorator. The integration tests will crash then :-)

fstagni:
I agree with Chris: we rely on the existing convention in probably too many places, so this is not something for v7r0. 
Anyway, in general, I also think that what you are proposing is indeed "more correct" than what's now.

ConfigurationSystem/Client/LocalConfiguration.py
```diff
@@ -255,13 +257,24 @@ def __parseCommandLine(self):
...
+    if os.getenv("DIRACNOCFG", None):
+      self.commandArgList = args
+    else:
+      # to avoid issuing the warning for correctly passed cfg files
+      extraCfg = [arg for arg in args if arg[-4:] == ".cfg"]
+      self.cliAdditionalCFGFiles.extend(extraCfg)
+      self.commandArgList = [arg for arg in args if not arg[-4:] == ".cfg"]
+      if extraCfg:
+        # use error level to make sure users will always see the Warning
+        gLogger.error("WARNING: Passing cfg files directly is deprecated. Use --cfg instead.")
```
andresailer, line 277: This warning is going to be printed also when people are calling dirac-dms-add-file with a cfg file, right?
With the addition of the env variable they could also just set DIRAC_NO_CFG in their own session or just for the command, right?

chrisburr: Also should `gLogger.always` be used instead of `gLogger.error`?

marianne013: @andresailer That was the intention.
@chrisburr We found warning wasn't enough, so we used error. Doesn't .always mean you can't suppress it at all ? That might be one step too far.

andresailer: As there are two questions in my post, I am not sure which was the intention, that the warning is printer when calling dirac-dms-add-file, or that people add the variable in their session, or both.

But I would then propose to update the message to mention the export of the environment variable.

marianne013: @andresailer: The message was very generic as it affects other commands besides dirac-dms-add-file. We were hoping that way we just catch everywhere its used.

ConfigurationSystem/Client/LocalConfiguration.py
```diff
@@ -255,13 +257,24 @@ def __parseCommandLine(self):
...
+      if opt in ('-h', '--help'):
         self.showHelp()
         sys.exit(2)
+      if opt == '--cfg':
+        self.cliAdditionalCFGFiles.append(val)
 
-    self.cliAdditionalCFGFiles = [arg for arg in args if arg[-4:] == ".cfg"]
-    self.commandArgList = [arg for arg in args if not arg[-4:] == ".cfg"]
+    # environment variable to ensure smooth transition
+    if os.getenv("DIRACNOCFG", None):
```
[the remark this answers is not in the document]
marianne013: @chrisburr The docs is easy. There doesn't seem to be the list of commented out environment variables in bashrc in v7r0.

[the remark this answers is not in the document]
marianne013: @chrisburr Just realized the bashrc would have to go into integration. I don't want to make this pull request against integration, so I would have to do this separately.

ConfigurationSystem/Client/LocalConfiguration.py
chrisburr: > We found warning wasn't enough, so we used error. Doesn't .always mean you can't suppress it at all ? That might be one step too far.

Error feels misleading to me as I think it should mean that something has wrong. If someone wants to suppress the message they should fix what they're doing and use `--cfg`.

(apologies for having merged two separate points into this thread)

andresailer: I just mean that for users it is still confusing if they try to add (remove, download...) a .cfg file and get this warning, which doesn't offer them the solution they probably need, as far as I understand.

marianne013:
@chrisburr et al: Was there a consensus wrt always vs error ? I don't feel that strongly about it.
@andresailer Wrt the error message: I can't think of a catchy one. The proposed one is still better than the current behaviour which (after Simon's patch) at least tells the user something is wrong ("wrong number of arguments"). If someone has a proposal, by all means just add it in. 
We were hoping the current state of affairs not to last too long, so only a limited number of users would ever run into this anyway.

ConfigurationSystem/Client/LocalConfiguration.py
andresailer:
```suggestion
                 gLogger.error("""WARNING: Parsing of '.cfg' files as command line arguments is changing!
Set the environment variable 'export DIRAC_NO_CFG=1' to pass the file as a positional argument. (This will become the default)
To modify the local configuration use '--cfg <configfile>' instead.""")
```
(better indentation of the code could be done with https://docs.python.org/2/library/textwrap.html#textwrap.dedent)

In addition, we could  consider setting the variable inside some of the scripts directly.

ConfigurationSystem/Client/LocalConfiguration.py
```diff
@@ -255,13 +257,28 @@ def __parseCommandLine(self):
...
+    if os.getenv("DIRAC_NO_CFG", None):
+      self.commandArgList = args
+    else:
+      # to avoid issuing the warning for correctly passed cfg files
+      extraCfg = [arg for arg in args if arg[-4:] == ".cfg"]
+      self.cliAdditionalCFGFiles.extend(extraCfg)
+      self.commandArgList = [arg for arg in args if not arg.endswith(".cfg")]
+      if extraCfg:
+        # use error level to make sure users will always see the Warning
+                 gLogger.error("""WARNING: Parsing of '.cfg' files as command line arguments is changing!
```
andresailer, line 277: Indentation is a bit off here, sorry.

marianne013: Yes, I should have checked.

ConfigurationSystem/Client/LocalConfiguration.py
```diff
@@ -255,13 +257,24 @@ def __parseCommandLine(self):
...
+        self.cliAdditionalCFGFiles.append(val)
 
-    self.cliAdditionalCFGFiles = [arg for arg in args if arg[-4:] == ".cfg"]
-    self.commandArgList = [arg for arg in args if not arg[-4:] == ".cfg"]
+    # environment variable to ensure smooth transition
+    if os.getenv("DIRACNOCFG", None):
+      self.commandArgList = args
+    else:
+      # to avoid issuing the warning for correctly passed cfg files
+      extraCfg = [arg for arg in args if arg[-4:] == ".cfg"]
```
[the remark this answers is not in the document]
fstagni: Please commit this one too.

ConfigurationSystem/Client/LocalConfiguration.py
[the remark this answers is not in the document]
fstagni: > @chrisburr Just realized the bashrc would have to go into integration. I don't want to make this pull request against integration, so I would have to do this separately.

Please do that before this PR is merged and the above "task" is forgotten.

andresailer:
There is another place in the documentation that should be updated.
I could push this to your branch, or you modify the file (docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst) yourself.
```diff
diff --git a/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst b/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst
index f063da98b..c43fa6adf 100644
--- a/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst
+++ b/docs/source/AdministratorGuide/Configuration/ConfigurationStructure/index.rst
@@ -55,10 +55,18 @@ in the order of preference of the option resolution:
      dirac-wms-job-submit job.jdl -o /DIRAC/Setup=Dirac-Production

 *Command line argument specifying a CFG file*
-  If a filename with the *.cfg* extension is passed as an argument to any DIRAC command
-  it will be interpreted as a configuration file. For example::
+  A config file can be passed to any dirac command with the ``--cfg`` flag::

-     dirac-wms-job-submit job.jdl my.cfg
+     dirac-wms-job-submit job.jdl --cfg my.cfg
+
+  .. versionchanged:: v7r0
+
+     The passing of ``.cfg`` files was changed to require the ``--cfg`` flag.
+
+  .. deprecated:: v7r0
+
+     If a filename with the ``.cfg`` extension is passed as an argument to any DIRAC command
+     it will be interpreted as a configuration file, if the ``DIRAC_NO_CFG`` environment variable is not set.

```

marianne013:
@andresailer By all means push the documentation. I have apparently again overlooked two whitespaces. Grrrr...

andresailer:
Ok, pushed the commit.

marianne013:
I did a squash merge to re-trigger the tests, which now all seem to have passed even though the code is unchanged. I think that's all outstanding issues dealt with.

PR #4661, opened 2020-06-29
[v7r2] Integration merged with v7r1

atsareg:


.github/workflows/basic.yml
```diff
@@ -12,11 +12,11 @@ jobs:
       fail-fast: False
       matrix:
         command:
-          - pytest
-          - DIRAC_USE_M2CRYPTO=Yes DIRAC_M2CRYPTO_SPLIT_HANDSHAKE=Yes pytest
+          - pytest --no-cov
+          - DIRAC_USE_M2CRYPTO=Yes DIRAC_M2CRYPTO_SPLIT_HANDSHAKE=Yes pytest --no-cov
           # Security tests are flakey due to reference counting bugs in pyGSI/src/crypto/asn1.c
-          - pytest Core/Security/test || (echo "Retrying..."; pytest Core/Security/test) || (echo "Retrying again..."; pytest Core/Security/test)
```
andresailer, line 18: The repeated running was added for failures due to the pyGSI library, so it should probably stay with the repetition, but adding --no-cov?

chrisburr: As this is targetting `integration` the repeated runs can be removed as pyGSI was removed in #4481

PR #4671, opened 2020-07-07
[integration] added example of DIRAC_NO_CFG into bashrc

marianne013:
BEGINRELEASENOTES
*Core
CHANGE: Add an example for DIRAC_NO_CFG to bashrc.
ENDRELEASENOTES

### DIFF
--- a/Core/scripts/dirac-install.py
+++ b/Core/scripts/dirac-install.py
@@ -2214,7 +2214,8 @@ def createBashrc():
       lines.append('# export DIRAC_USE_M2CRYPTO=true')
       lines.append('# export DIRAC_USE_NEWTHREADPOOL=yes')
       lines.append('# export DIRAC_VOMSES=$DIRAC/etc/grid-security/vomses')
-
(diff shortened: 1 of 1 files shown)


chaen:
Btw, it may be worth to add in the wiki a note about the `run` files on the servers that need updating. I am just not sure on which version it should be done.
I cooked up a one line command that will update the `run` files:
```
find -L /opt/dirac/startup/ -name run -exec sed -i.bak -E '/--cfg/! s| ((/[a-zA-Z_0-9]+)+\.cfg)| --cfg \1|g' {} \;
```

chaen:
Also the component installer should be modified to add the `--cfg` https://github.com/DIRACGrid/DIRAC/blob/integration/FrameworkSystem/Client/ComponentInstaller.py#L1915
The fact that the tests in github action have not failed make me believe that we are not setting this `DIRAC_DEPRECATED_FAIL` where we should be doing it. I'll have a look at that

chaen:
An attempt in https://github.com/DIRACGrid/DIRAC/pull/4690

PR #4688, opened 2020-07-17
[v7r0]M2Crypto: better handling of timeout

chaen:
An implementation of timeout with M2crypto, and a lot of docs about it in general. 
BEGINRELEASENOTES

*Subsystem
NEW/CHANGE/FIX: explanation

ENDRELEASENOTES


chaen:
Tested in LHCb, and all good.

PR #4693, opened 2020-07-23
[v7r1] retry stageout for any failure condition

marianne013:
BEGINRELEASENOTES
*DataManagementSystem
CHANGE: Retry file upload for any error condition, not just file catalogue failure.
ENDRELEASENOTES

### DIFF
--- a/DataManagementSystem/Client/FailoverTransfer.py
+++ b/DataManagementSystem/Client/FailoverTransfer.py
@@ -93,13 +93,14 @@ class FailoverTransfer(object):
                                                                 se,
                                                                 guid=fileGUID,
                                                                 checksum=fileChecksum)
-        self.log.verbose(result)
-
-        # If the FC is unavailable, we stay in the loop and retry
-        # otherwise we continue without retrying
-        if result['OK'] or not cmpError(result, EFCERR):
+        # retry on any failure
+        if result['OK']:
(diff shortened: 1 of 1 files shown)


chaen:
Although I have nothing against the principle of retrying, I would add an extra constraint. This `TransferAndRegisterFile` (https://github.com/DIRACGrid/DIRAC/blob/integration/DataManagementSystem/Client/FailoverTransfer.py#L66) method takes as parameter `destinationSEList`. As the non existing documentation does not state it, this list is a list of all possible destinations, and not all the destination we want. 
So in my opinion, it does not make any sense to wait between two storages. We should just continue. 
So I would retry only if `destinationSEList` has a single element.
What do you think ?

marianne013:
@chaen Yes, if there's more than one possible SE there's probably no point in Waiting. I have to admit, we don't really have that use case in GridPP (apart from t2k and they have their own internal replication system using DIRAC in some creative never foreseen manner ;-).

marianne013:
@chaen Would you mind coding your constraint (the SE list) as I have no experience with this ?
I a related note, does anyone know why the tests didn't run ?

chaen:
OK I'll do it

chaen:
How about https://github.com/DIRACGrid/DIRAC/pull/4699  ?

Issue #4682, opened 2020-07-16
Why MultiVOFileCatalog must be master in the tests ? 


chaen:
@martynia in the tests you've added for multiVO DFC, you expect this catalog to be the master. 
There already was a master catalog, and adding a second one is not good (as stated by the doc). 

I needed to add a trick in the HTTPs PR, where I set the MultiVO as Master, and remove the old one (see https://github.com/DIRACGrid/DIRAC/pull/4677/commits/13acc8e70040494d2aabafac2a794a331c93a865) . Note that this is not a good solution, because such expectations results in tests not being re-runnable.

Is there a reason why this catalog must be Master ? If yes, I suspect that it is because of a test design flaw, but maybe not.
If there is no reason, could you please fix the tests ?

martynia:
I've done a very similar trick in my Jenkins test setup. The original master File Catalog is configured for a single VO, so other tests can run. The MultiVO file catalog is configured using MultiVO metadata plug-ins so the tests can write and read from it.  MutliVO plugin cannot write into a single VO (original master) catalog. These two implementations are not compatible. Since there may be only one Master catalog, the original one has to be inactivated. This situaltion, where we have both single and multi VO catalogs is not a real life scenario, a real life server whould have just one master catalog, configured for a single or multi VO metadata handling. Does it make sense ?

chaen:
Hum, not sure it does. Where do you see that there is a requirement for the original single VO catalog to be master ?

martynia:
Could we have a chat over Skype or similar ?
 JM

chaen:
To summarize the discussion:
* Remove the definition of the `MultiVOFileCatalog` as master https://github.com/DIRACGrid/DIRAC/blob/integration/tests/Jenkins/dirac-cfg-update-server.py#L120
* Remove alltogether the script `dirac-cfg-update-filecatalog.py`
* In `Test_UserMetadata`, do not use `FileCatalog`, but `FileCatalogClient` as done in https://github.com/DIRACGrid/DIRAC/blob/integration/tests/Integration/DataManagementSystem/Test_Client_DFC.py#L47

fstagni:
In #4695

PR #4695, opened 2020-07-24
[v7r1] Fix MutiVO Catalog tests (adresses #4682)

martynia:
BEGINRELEASENOTES
Tests are modified so the MultiVO File catalog is not a master catalog anymore. Adresses #4682 .

ENDRELEASENOTES


chaen:
perfect to me. Just this strange test that did not run. I will retrigger.

PR #4698, opened 2020-07-28
[v7r0] Glue2: fix problem for ARC CEs without execution environments

andresailer:
resExeInfo is an internal loop variable that shouldn't be used outside of its loop
MANAGER has to be something compatible with the split/rsplit expectations
The result is the second entry in norgugrid-MANAGER-queueName, which is discarded in ARCComputingElement

Supersede #4696 , @fstagni can you try with this patch?

BEGINRELEASENOTES

*Core
FIX: Glue2: Fix crash in bdii2CSAgent, when ARC CEs do not have an ExecutionEnvironment

ENDRELEASENOTES

PS: should this go to v7r0?


andresailer:
Rebased to v7r0

PR #4724, opened 2020-08-20
[v7r1]Fix MessageBroker with new threadpool

chaen:
This MessageBroker is silly because it does not create any error. We just saw that no Optimizer was optimizing
 anything
BEGINRELEASENOTES
*Tests
FIX: fix rss-scripts syntax

*Core:
FIX: correctly pass the argument to the new ThreadPool in MessageBroker

ENDRELEASENOTES

### DIFF
--- a/Core/DISET/private/MessageBroker.py
+++ b/Core/DISET/private/MessageBroker.py
@@ -182,7 +182,7 @@ class MessageBroker(object):
                                                                       result['Message']))
       return self.removeTransport(trid)
     if useThreadPoolExecutor:
-      self.__threadPool.submit(self.__processIncomingData, (trid, result))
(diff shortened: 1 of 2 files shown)


andresailer:
Add a callback to log the errors?
https://cpython-test-docs.readthedocs.io/en/latest/library/concurrent.futures.html#concurrent.futures.Future.add_done_callback

chaen:
I must admit I did not intend to really dig into that... I may have a look, but looking at the method which is called in this particular case, it would be of no use, since what is returned is the not the result of the actual method (except in case of success), but the result of `self.removeTransport` which ultimately we do not really care. 
Anyway, I'll have a quick look, but I wont change the logic

fstagni (approved):

atsareg (approved):

PR #4726, opened 2020-08-26
[v7r2] Run most unit tests for Python 3

chrisburr:
This builds upon #4718 to add support for actually running the unit tests instead of just collecting them. As #4718 hasn't been merged yet, see here for the diff https://github.com/chrisburr/DIRAC/compare/py3-pytest-discover...chrisburr:run-py3-pytest. As usual for these changes, reviewing individual commits might be easier that the entire PR.

There is a lot left to port but I limited myself to the unit tests so we can be reasonably sure I haven't broken anything.  I suspect after this is merged it shouldn't take much more to get a functional client installation running with Python 3.

This PR includes #4391 as I didn't want waste time looking for a Python 3 build that uses OpenSSL 1.0.2, @chaen can #4391 be unmarked as WIP now M2Crypto has been updated?

There are three tests which are still disabled as they fail:

* `test_BaseType_Unicode` and `test_nestedStructure` fail due to DEncode's string and unicode types being poorly defined. I think this is hard to fix without major changes to DIRAC or backwads incompatible changes to DEncode itself. Hopefully we can ignore it until we've moved to JSON.
* `testLockedClass` deadlocks frequently due to the use of self.__processThread._Thread__stop(). Officially "threads cannot be destroyed, stopped, suspended, resumed, or interrupted" so the code should be rewritten to follow the supported alternative: "If you want your threads to stop gracefully, make them non-daemonic and use a suitable signalling mechanism such as an Event.".

I also add `flaky` to the CI as `test_cpuUsage` struggles to produce enough system CPU usage to be registered thanks performance improvements in Python 3. I think it's worthwhile so we can use it for the handful of other tests which are unreliable.

BEGINRELEASENOTES

*Python 3
NEW: Run most unit tests with Python 3

ENDRELEASENOTES


DataManagementSystem/Client/test/Test_Client_DataManagementSystem.py
```diff
@@ -41,7 +41,7 @@ class ConsistencyInspectorSuccess(UtilitiesTestCase):
 
   def test_getReplicasPresence(self):
     res = self.ci.getReplicasPresence(['/this/is/dir1/file1.txt', '/this/is/dir1/file2.foo.bar'])
-    self.assertEqual(res, (['/this/is/dir1/file1.txt', '/this/is/dir1/file2.foo.bar'], []))
+    self.assertEqual(tuple(map(set, res)), ({'/this/is/dir1/file1.txt', '/this/is/dir1/file2.foo.bar'}, set()))
```
chaen, line 44: why changing that ?

Core/Utilities/test/Test_DAG.py
```diff
@@ -13,6 +13,17 @@
 
 __RCSID__ = "$Id $"
 
+def listToSet(obj):
```
chaen, line 17: I am not sure this is correct. You may want to have duplicated items in a list. Why wasn't `sorted` good enough ?

chrisburr: Because the list contains a mixtures of types which have no sensible comparison operator between them. In Python 2 this was poorly defined and could result in bizarre failures. In [Python 3 this is now an error](https://portingguide.readthedocs.io/en/latest/comparisons.html#unorderable-types):

```python
In [2]: sorted(["a", 1, dict()])
<ipython-input-2-adfb206d711c> in <module>
----> 1 sorted(["a", 1, dict()])

TypeError: '<' not supported between instances of 'int' and 'str'
```

These lists internally come from `dict` and `set` objects which have no stable ordering in Python 2 (`dicts` are in Python 3.7+ but `set` still isn't and I don't like trusting either).

A lot of these tests that I've fixed were comparing to the "random" order than Python 2.7 happens to produce if executed in this order. A few places were so unstable that introducing `pdb.set_trace()` was enough to trigger them to reshuffle (probably due to the garbage collector having a chance to run).

Core/DISET/private/Transports/BaseTransport.py
```diff
@@ -25,6 +25,7 @@
 
 import time
 import select
+import six
```
chaen, line 28: changing this file makes me kind of nervous :-) Anyway, it has to be done. 
I just hope we won't get tricked with non UTF character since then `len` comparison would not make sense anymore..

Core/Utilities/DEncode.py
```diff
@@ -30,6 +30,18 @@
 from pprint import pprint
 
 
+def _ord(char):
```
chaen, line 33: Why not just applying `ord` everywhere since it works in python2 as well ?

Core/Utilities/List.py
```diff
@@ -127,7 +127,8 @@ def breakListIntoChunks(aList, chunkSize):
   """
   if chunkSize < 1:
     raise RuntimeError("chunkSize cannot be less than 1")
-  if isinstance(aList, (set, dict, tuple, six.moves.collections_abc.Iterable)):
+  if isinstance(aList, (set, dict, tuple, {}.keys().__class__,
```
chaen, line 130: why ? Also if there is a good reason, you may want to document it, because someone that knows of `collections_abc` and comes after you will surely modify it back :-D

WorkloadManagementSystem/Utilities/PilotWrapper.py
```diff
@@ -125,7 +125,7 @@ def pilotWrapperScript(pilotFilesCompressedEncodedDict=None,
 """ % {'name': name,
        'value': value}
 
-  # add X509_USER_PROXY to establish pilot env in Cluster WNs
+  # add X509_USER_PROXY to etablish pilot env in Cluster WNs
```
chaen: establish is correct no ?

tests/Integration/WorkloadManagementSystem/Test_GenerateAndExecutePilotWrapper.py
```diff
@@ -74,12 +74,9 @@
...
 
 diracInstall = os.path.join(os.getcwd(), 'dirac-install.py')
-with open(diracInstall, "r") as fd:
+with open(diracInstall, "rb") as fd:
   diracInstall = fd.read()
-if sys.version_info >= (3,):
-  diracInstallEncoded = base64.b64encode(bz2.compress(diracInstall.encode(), 9))
-else:
-  diracInstallEncoded = base64.b64encode(bz2.compress(diracInstall, 9))
+diracInstallEncoded = base64.b64encode(bz2.compress(diracInstall, 9)).decode()
```
chaen, line 79: so, this variable called `somethingEncoded` is in fact a decoded bytes into string of a base 64 encoded bytes... and who said py3 was easier and more logical ? :D

Core/Utilities/test/Test_ProcessPool.py
```diff
@@ -24,18 +24,30 @@
...
 from DIRAC import gLogger
 ## SUT
 from DIRAC.Core.Utilities.ProcessPool import ProcessPool
 
+
+@pytest.fixture(autouse=True)
+def capture_wrap():
+  """ Avoid https://github.com/pytest-dev/pytest/issues/5502 """
+  sys.stderr.close = lambda *args: None
+  sys.stdout.close = lambda *args: None
```
chaen, line 47: ouch :D 
Though I learned something about `autouse` :-)

environment.yml
```diff
@@ -50,6 +50,7 @@ dependencies:
   - pytest-cov >=2.2.0
   - pytest-mock
   - shellcheck
+  - flaky
```
chaen, line 53: also in the `requirements.txt` ?

Core/Utilities/List.py
chrisburr: I added the `six.moves.collections_abc.Iterable` in an earlier commit but in integration it's currently:

https://github.com/DIRACGrid/DIRAC/blob/34c753dc0cd827d7df262cd1989bc7adfbb05f1c/Core/Utilities/List.py#L130-L131

It needs to be changed as `dict_keys`, `dict_values` and `dict_items` are all distinct types in Python 3.

My change to check against `Iterable` seemed nicer but it was a stupid thing to do. (`str` is iterable but `list("LFN:/myfile.txt")` is not what I wanted ;))

Core/Utilities/DEncode.py
chrisburr: It doesn't:

Python 2:
```
In [1]: ord("a") == b"a"[0]
Out[1]: False
```
Python 3:
```
In [1]: ord("a") == b"a"[0]
Out[1]: True
```

At first I tried to add both `ord(x)`  and `x` to the `g_d(En|De)codeFunctions` dictionary to work around it becomes a mess. I prefer this solution as it minimises the chance of breaking Python 2.

tests/Integration/WorkloadManagementSystem/Test_GenerateAndExecutePilotWrapper.py
chrisburr: Base64 is [defined to be encoded in ASCII](https://en.wikipedia.org/wiki/Base64). The only way to achieve that for all possible system locales is to return a `bytes` object so technically the `.decode()` shouldn't be there but then I'd have to start fixing `pilotWrapperScript`.

Core/Utilities/test/Test_ProcessPool.py
chrisburr: It's actually still a mess as DIRAC's logging prevents the stderr messages from being captured by each test. What actually happens is `pytest` attributes all of the output to the first test that is executed. (That was a fun 20 minutes to make sense of...)

WorkloadManagementSystem/Utilities/PilotWrapper.py
chrisburr: Yes it is!

Core/DISET/private/Transports/BaseTransport.py
chrisburr: > changing this file makes me kind of nervous :-)

Wait until you see my next PR...

>  I just hope we won't get tricked with non UTF character since then len comparison would not make sense anymore..

This won't be a problem as this treats everything as a sequence of bytes instead of a string so the length will always be correct.

environment.yml
chrisburr: Yes it should!

DataManagementSystem/Client/test/Test_Client_DataManagementSystem.py
chrisburr: `getReplicasPresence` uses dictionaries internally and doesn't maintain the order consistently. The fact that is worked in Python 2 is pure luck.

ConfigurationSystem/Client/Utilities.py
```diff
@@ -274,8 +274,7 @@ def addToChangeSet(entry, changeSet):
         if submissionMode == "Unknown" and newSubmissionMode:
           addToChangeSet((ceSection, 'SubmissionMode', submissionMode, newSubmissionMode), changeSet)
 
-        queues = ceInfo['Queues'].keys()
-        for queue in queues:
+        for queue in ceInfo['Queues'].keys():
```
andresailer, line 277:
```suggestion
        for queue in ceInfo['Queues']:
```
actually 
```
    for queue, queueInfo in ceInfo['Queues'].items():
```
and dropping the next line, right?

chrisburr: Yeah that can be done. In general I've avoided stylistic improvements as this PR is already larger than is reasonable.

Core/Utilities/ClassAd/ClassAdLight.py
```diff
@@ -254,7 +254,7 @@ def asJDL(self):
     """
 
     result = ''
-    for name, value in self.contents.items():
+    for name, value in sorted(self.contents.items()):
```
andresailer, line 257: Why `sorted`?

chrisburr: As the tests for the JDL and XML compare the raw bytes in the files. As there is no ordering of dictionaries in Python 2.7 this is a quirk of the interpreter state at the particular time the file is written.

In ef16451eaaf10eac817941ddbd30a64b6ae21d7a I make the XML and JDL deterministic and update the references accordingly.

Core/Utilities/test/Test_DAG.py
```diff
@@ -14,6 +14,18 @@
 __RCSID__ = "$Id $"
 
 
+def listToSet(obj):
+  """ Convert a list of objects into a set which can be used for comparision """
```
andresailer, line 18:
```suggestion
  """ Convert a list of objects into a set which can be used for comparison """
```

Core/Utilities/test/Test_DAG.py
```diff
@@ -14,6 +14,18 @@
...
+  result = set()
+  for x in result:
```
andresailer, line 20:
```suggestion
  for x in obj:
```
as `result = set()`

chrisburr: Ouch...well spotted!

DataManagementSystem/Client/test/Test_Client_DataManagementSystem.py
chaen: Maybe just add a comment, cause it makes it look very obscure :-)
You say it's pure luck in Python 2, but Python 3 is still random when it comes to dictionary order, isn't it ?

andresailer: Python 3.7 guarantees insertion order: https://stackoverflow.com/a/39980744

chaen: Ah forge the question, I just found in https://docs.python.org/3/tutorial/datastructures.html

> Performing list(d) on a dictionary returns a list of all the keys used in the dictionary, in insertion order (if you want it sorted, just use sorted(d) instead). To check whether a single key is in the dictionary, use the in keyword.

Core/Utilities/test/Test_DAG.py
chaen: Hum, another "nice" feature of python 3 I guess....

Core/DISET/private/Transports/BaseTransport.py
chaen: > Wait until you see my next PR...
Now you scare me :-) 

>This won't be a problem as this treats everything as a sequence of bytes instead of a string so the length will always be correct.
Yeah that's what I understood, but since I still have a hard time getting my head around that.. :-P

Core/Utilities/List.py
chaen: > It needs to be changed as dict_keys, dict_values and dict_items are all distinct types in Python 3.
The more I learn about Python 3, the less of an improvement it looks to me :thinking: 

Anyway, I understand now. Ahhhhhh if only we had unit tests :-p

Maybe a bit less hacky and more understandable way ? 

```
if isinstance(aList, six.moves.collections_abc.Iterable) and not isinstance(aList, six.string_types):
```

chrisburr: Well the code should be modernised everywhere and this function should be deleted (`[::chunkSize]`) but that should be a separate PR. I prefer it like this as it definitely won't change how DIRAC currently works in Python 2.

Core/Utilities/test/Test_ProcessPool.py
chaen: The more I think of it, the more I feel like scrapping gLogger altogether. Although I suspect base logging would do the same no ?

Core/Utilities/DEncode.py
chaen: Ah indeed.
Btw, for consistency, should all the 
`g_dDecodeFunctions[_ord("i")] = decodeInt`
be replaced with 
`g_dDecodeFunctions[_ord(b"i")] = decodeInt`
for consistency ?

Core/Utilities/test/Test_ProcessPool.py
chrisburr: I think the base logging would do it slightly differently as I've used it in other places and not had this issue (but I didn't look into it properly).

Core/Utilities/DEncode.py
chrisburr: I don't think so as `ord("i")` is saying "what is the ASCII value of this character" whereas `b"i"` is saying this is a byte array with a single value (`115`). You could argue for `b"i"[0]` for consistency (as this is functionally what is done when decoding the incoming data) but `ord` is clearer IMHO.

WorkloadManagementSystem/Utilities/test/Test_PilotCStoJSONSynchronizer.py
```diff
@@ -173,9 +174,19 @@ def test_syncchecksum(self):
...
     assert synchroniser._checksumDict[self.testCfgFileName] == expectedHash
-    assert open('checksums.sha512', 'rb').read().split('\n')[0] == '%s  %s' % (expectedPJHash, 'pilot.json'), \
-        'pilot.json content: ' + repr(open('pilot.json', 'rb').read())
-    assert open('checksums.sha512', 'rb').read().split('\n')[1] == '%s  %s' % (expectedHash, self.testCfgFileName)
+    with open('checksums.sha512', 'rt') as fp:
+      checksums = fp.read()
+    with open('pilot.json', 'rt') as fp:
+      pilot_json_content = fp.read()
+
+    if not six.PY2:
```
andresailer, line 182: Why this just for python3? Is json in python2 deterministic?
Did this fail without re-calculation?

chrisburr: Neither is deterministic and it's an inherent "feature" of JSON specification. It's probably stable for a given Python version but it shouldn't be relied upon given JSON explicitly allows reading and writing to shuffle the order of maps. 

The reason I only applied this for Python 3 is that the test already works for Python 2 so I don't want to break anything that is already working. Next time the hashes change we should probably drop the hard coded values.

andresailer: But then it could be someone breaks the pilot.json writing and we just happily calculate the checksum for the broken file.

chrisburr: That's why I left it as-is for Python 2. If the name is to be believed (`test_syncchecksum`) this test is only validating the checksumming but 🤷.

andresailer: two tests for the price of one!

andresailer (commented):
I've clicked through all the files.

Core/Utilities/List.py
chaen: yep

PR #4732, opened 2020-08-31
[v7r2] Run client integration tests for Python 3

chrisburr:
The main thing this PR does is crudely add support for passing `--pythonVersion=3` to `./dirac-install.py`. When this happens it gets the latest conda based installer from https://github.com/chrisburr/DIRACOS2.

As this builds on #4726 which hasn't been merged, see here for the diff https://github.com/chrisburr/DIRAC/compare/python3-client...chrisburr:python3-client-integration.

BEGINRELEASENOTES

*Core
NEW: Support installing Python 3 clients using by passing `--pythonVersion=3` to `./dirac-install.py`

ENDRELEASENOTES


DataManagementSystem/Client/FileCatalogClientCLI.py
```diff
@@ -6,7 +6,12 @@
 from __future__ import division
 __RCSID__ = "$Id$"
 
-import commands
+# TODO: This should be moderised to use subprocess(32)
```
fstagni, line 9:
```suggestion
# TODO: This should be modernised to use subprocess(32)
```
(in few places)

DataManagementSystem/Client/FileCatalogClientCLI.py
```diff
@@ -6,7 +6,12 @@
...
+try:
+  import commands
+except ImportError:
+  # Python 3's subprocess module contains a compatability layer
```
fstagni, line 13:
```suggestion
  # Python 3's subprocess module contains a compatibility layer
```
(in few places)

ResourceStatusSystem/Agent/test/Test_Agent_SiteInspectorAgent.py
```diff
@@ -40,7 +40,10 @@ def test__execute(mocker, sitesToBeCheckedValue):
   """
 
   mocker.patch("DIRAC.ResourceStatusSystem.Agent.SiteInspectorAgent.AgentModule.__init__")
-  mocker.patch("DIRAC.ResourceStatusSystem.Agent.SiteInspectorAgent.AgentModule.am_getOption", side_effect=mockAM)
+  mocker.patch(
+      "DIRAC.ResourceStatusSystem.Agent.SiteInspectorAgent.AgentModule._AgentModule__moduleProperties",
+      side_effect=lambda x, y=None: y, create=True
+  )
```
fstagni, line 46: Why these changes?

Resources/Catalog/FileCatalog.py
```diff
@@ -202,7 +202,7 @@ def w_execute(self, *parms, **kws):
       fileInfo, lfnMapDict = result['Value']
       # No need to check the LFNs again in the clients
       kws['LFNChecking'] = False
-      allLfns = fileInfo.keys()
+      allLfns = list(fileInfo.keys())
```
fstagni, line 205:
```suggestion
      allLfns = list(fileInfo)
```
should be enough

WorkloadManagementSystem/Client/DIRACbenchmark.py
```diff
@@ -55,7 +55,7 @@ def singleDiracBenchmark(iterations=1, measuredCopies=None):
       start = os.times()
 
     # Now the iterations
-    for _j in xrange(n):
+    for _j in range(n):
```
fstagni, line 58: This is the benchmarking module, here performance is key. I would use `six.moves.range`

WorkloadManagementSystem/Service/WMSAdministratorHandler.py
```diff
@@ -184,7 +184,7 @@ def export_getSiteMaskSummary(cls):
     return S_OK(siteDict)
 
   ##############################################################################
-  types_getJobPilotOutput = [(basestring, int, long)]
+  types_getJobPilotOutput = [six.string_types + (int, long)]
```
fstagni, line 187: For `long` shouldn't we we `six.integer_types`?

tests/CI/run_docker_setup.sh
```diff
@@ -150,20 +150,23 @@ prepareEnvironment() {
...
+  if [[ "${SERVER_USE_PYTHON3:-}" == "Yes" ]]; then
+    echo "INSTALLOPTIONS+=(\"--pythonVersion=3\")" >> "${SERVERCONFIG}"
   fi
 
-  if [[ -n "${CLIENT_USE_M2CRYPTO+x}" ]]; then
-    echo "export DIRAC_USE_M2CRYPTO=${CLIENT_USE_M2CRYPTO}" >> "${CLIENTCONFIG}"
-    echo "export DIRAC_M2CRYPTO_SPLIT_HANDSHAKE=Yes" >> "${CLIENTCONFIG}"
+  echo "${CLIENT_USE_PYTHON3}"
+  if [[ "${CLIENT_USE_PYTHON3:-}" == "Yes" ]]; then
+    echo "INSTALLOPTIONS+=(\"--pythonVersion=3\")" >> "${CLIENTCONFIG}"
```
fstagni, line 161: We just take the last version?

Core/Utilities/MySQL.py
```diff
@@ -161,11 +161,15 @@
...
-  MySQLdb.server_init(['--defaults-file=/opt/dirac/etc/my.cnf', '--datadir=/opt/mysql/db'], ['mysqld'])
-except MySQLdb.ProgrammingError:
-  pass
+# The mysql_server_init function called here is deprecated and not used by mysqlclient
+# https://dev.mysql.com/doc/c-api/8.0/en/mysql-server-init.html
+# TODO: Check if any changes are needed
+if six.PY2:
+  # This is for proper initialization of embedded server, it should only be called once
+  try:
+    MySQLdb.server_init(['--defaults-file=/opt/dirac/etc/my.cnf', '--datadir=/opt/mysql/db'], ['mysqld'])
```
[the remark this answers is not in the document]
chrisburr: See the second to last commit (`5a66509`) in the PR description of: https://github.com/DIRACGrid/DIRAC/pull/4718

tests/CI/run_docker_setup.sh
chrisburr: For now DIRACOS2 is using Python 3.8 but I don't think this option should be used to pick the Python version. Instead the full Python version will be implicit from the DIRACOS version.

ResourceStatusSystem/Agent/test/Test_Agent_SiteInspectorAgent.py
chrisburr: In Python 3 this mock wasn't being applied correctly but I've forgotten why...

Core/Utilities/MySQL.py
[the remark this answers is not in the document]
chrisburr: The proposal is just to remove these lines: https://github.com/DIRACGrid/DIRAC/commit/5a665093296ae3831c7182dc1ebda5347c833606#r41777842

chrisburr:
I think this is now ready to be merged. The last few commits do the following:
* to build upon the HTTPS PR
* always treat data from X509 certificates as bytes
* remove the `force_bytes` parameter that I had added in #4731 in favour of using [Unicode surrogate escapes](https://www.python.org/dev/peps/pep-0383/). This allows the interfaces to be kept the same while still passing arbitrary bytes in strings. This workaround is only needed while DEncode still exists.

I haven't tidied up the dirac-install script as I think this should be done in a separate PR to make it easier to review (#4775).

If people want to see the reviews of the earlier commits, they can be found in #4718, #4726 and #4731.

Core/scripts/dirac-install.py
```diff
@@ -2390,6 +2387,20 @@ def installDiracOS(releaseConfig):
   return True
 
 
+def installDiracOSPython3(releaseConfig):
+  """
+  Install DIRAC OS for Python 3.
+
+  :param ReleaseConfig releaseConfig: The ReleaseConfig object for configuring the installation
+  """
+  url = "https://github.com/chrisburr/DIRACOS2/releases/latest/download/DIRACOS-Linux-x86_64.sh"
```
fstagni, line 2396: Just reminder for changing this link.

fstagni:
IMHO this can be merged, at least to avoid conflicts. 

I would merge this BEFORE any other PRs targeting the integration branch. @atsareg

fstagni (approved):

Core/scripts/dirac-install.py
chrisburr: This should be done as part of #4775. I haven't set up `DIRACGrid/DIRACOS2` yet as I'd like to have tags for the tornado packages rather than building from a branch.

Core/Utilities/Grid.py
```diff
@@ -147,7 +147,7 @@ def ldapsearchBDII(filt=None, attr=None, host=None, base=None, selectionString="
 def ldapSite(site, attr=None, host=None):
   """ Site information from bdii.
 
-:param  site: Site as it defined in GOCDB or part of it with globing, for example: \UKI-*
+:param  site: Site as it defined in GOCDB or part of it with globing, for example: \\UKI-*
```
andresailer, line 150:
```suggestion
:param  site: Site as it defined in GOCDB or part of it with globing, for example: UKI-*
```
IIRC, I commented somewhere that this didn't make any difference? (Neither single nor double slash produce any output in the html documentation)

Core/Utilities/ObjectLoader.py
```diff
@@ -100,7 +100,7 @@ def __recurseImport(self, modName, parentModule=None, hideExceptions=False, full
       if impData[0]:
         impData[0].close()
     except ImportError as excp:
-      if str(excp).find("No module named %s" % modName[0]) == 0:
+      if "No module named" in str(excp) and modName[0] in str(excp):
```
andresailer, line 103:
```suggestion
      if ("No module named %s" % modName[0]) in str(excp):
```
Otherwise it isn't completely the same test, if the exception message is  `"No module named Something"+modName[0] `
(Not sure if parenthesis needed or operator precedence is already enough

Core/Utilities/Grid.py
chrisburr: A single slash is a problem as it means something:

Screenshot 2020-10-21 at 17 33 01

I assumed the slash was put there for a reason without realising it needed to be escaped?

andresailer (commented):
Can't answer the other one until I submit?

Core/Utilities/Grid.py
andresailer: From this diff, I seriously doubt that:
https://github.com/DIRACGrid/DIRAC/commit/98a7d8588aa403ae473ba4cfd518352d0ca57c09#diff-f360c4ed4c1d7f46ac67ea9aeb520dad2d4b3af6ba46a93357d883c1ec8c39baR134

It also isn't needed for the input to the ldap search. I could understand if the `*` needed to be escaped, but I don't see any warning in my local compilation runs when just removing all the slashes here.

DataManagementSystem/Client/FileCatalogClientCLI.py
```diff
@@ -2157,7 +2162,7 @@ def dataset_show(self, argss):
         print(dName)
     else:
       fields = ['Key', 'Value']
-      datasets = datasetDict.keys()
+      datasets = list(datasetDict.keys())
```
andresailer, line 2165:
```suggestion
      datasets = list(datasetDict)
```

Interfaces/API/Dirac.py
```diff
@@ -65,6 +66,13 @@
 
 COMPONENT_NAME = 'DiracAPI'
 
+try:
```
andresailer, line 69: Why not `if six.PY2` here?

andresailer (dismissed):
Apart from the minor comments posted

Core/Utilities/ObjectLoader.py
chrisburr: This is intentional as the message is different.

Python 2:

```python
In [1]: import abcs
<ipython-input-1-9ec4e562403a> in <module>()
----> 1 import abcs

ImportError: No module named abcs
```

Python 3:
```python
In [1]: import abcs
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-9ec4e562403a> in <module>
----> 1 import abcs

ModuleNotFoundError: No module named 'abcs'
```

Interfaces/API/Dirac.py
chrisburr: Its hard to explain but feels more natural to use `try...except` here. I want to check if `file` is defined rather than use my knowledge about the Python versions involved.

Core/Utilities/ObjectLoader.py
andresailer: Wow, that is annoying.
```
# import re  # already done in the module
if re.findall("No module named '?%s'?" % modName[0], str(excp)):
```


```In [1]: import re

In [2]: s1 = "No module named abcs"

In [3]: s2=  "No module named 'abcs'"

In [4]: re.findall("No module named '?abcs'?", s1)
Out[4]: ['No module named abcs']

In [5]: re.findall("No module named '?abcs'?", s2)
Out[5]: ["No module named 'abcs'"]

In [6]: re.findall("No module named '?abcs'?", 'No module named abcsfoo')
Out[6]: ['No module named abcs'] # even the old test was broken in this regard, similar python3

In [7]: re.findall("No module named '?abcs'?", 'No module named fooabcs')
Out[7]: []
```

Interfaces/API/Dirac.py
andresailer: There are other cases where try..except are used, but then six wasn't imported, so I felt I had to ask.
It isn't a problem, as long as no one does something like `file = 'myfile.txt'` somewhere above (unlikely, but 🤷).

atsareg (approved):

PR #4733, opened 2020-09-01
[v7r1] Use singularity userns flag if possible and allow CVMFS fallback

### LINKED ISSUE #4723: Add userns + singularity-on-CVMFS support to SingularityCE
An increasing common configuration at sites is to use user namespace support rather than setuid for singularity; requiring the singularity parameters to include "--userns". The SingularityCE code should detect this case and add the option automatically.

If userns is enabled on a node and singularity is not found by the usual checks, the CVMFS version of singularity should also be considered.

sfayer:
This patch closes #4723. It adds the --userns option to singularity if user namepsaces are enabled on the node. It also adds a standard CVMFS location to the search path for singularity, for sites that don't have it installed locally any more.

BEGINRELEASENOTES
*Resources
CHANGE: SingularityComputingElement: Enable userns option in singularity if possible.
CHANGE: SingularityComputingElement:  Use CVMFS as singularity fallback location if userns is enabled.
ENDRELEASENOTES


Resources/Computing/SingularityComputingElement.py
```diff
@@ -54,6 +54,9 @@
 echo "Finishing inner container wrapper scripts at `date`."
 
 """
+# Path to a directory on CVMFS to use as a fallback if no
+# other version found: Only used if node has user namespaces
+FALLBACK_SINGULARITY = "/cvmfs/oasis.opensciencegrid.org/mis/singularity/current/bin"
```
chrisburr, line 59: Do we need this now a user namespaces build of `singularity` is in DIRACOS?

Resources/Computing/SingularityComputingElement.py
```diff
@@ -75,6 +78,20 @@ def __init__(self, ceUniqueID):
...
+    try:
+      with open("/proc/sys/user/max_user_namespaces", "r") as proc_fd:
+        maxns = int(proc_fd.readline().strip())
+        # Any "reasonable number" of namespaces is sufficient
+        return (maxns > 100)
+    except Exception:
+      # Any failure, missing file, doesn't contain a number, etc. and we
+      # assume they are disabled.
+      return False
+    return False
```
chrisburr, line 94: This is unreachable
```suggestion

```

Resources/Computing/SingularityComputingElement.py
sfayer: Ah, I hadn't realised it was in DIRACOS, although I'm of the opinion that it doesn't hurt to have another fallback.

fstagni (dismissed):

Resources/Computing/SingularityComputingElement.py
andresailer: Would it make sense to make this configurable?

chrisburr: It's already configurable on a per CE basis:
```python
  def __hasSingularity(self):
    """ Search the current PATH for an exectuable named singularity.
        Returns True if it is found, False otherwise.
    """
    if self.ceParameters.get('ContainerBin'):
      binPath = self.ceParameters['ContainerBin']
      if os.path.isfile(binPath) and os.access(binPath, os.X_OK):
        self.__singularityBin = binPath
        self.log.debug('Use singularity from "%s"' % self.__singularityBin)
```

andresailer: Please document: https://dirac.readthedocs.io/en/latest/AdministratorGuide/Configuration/ConfReference/Resources/Computing/index.html#singularity-ce-parameters :grin:

sfayer:
Grah, segfault in the tests... Closing/Opening to attempt a re-run.

sfayer:
Grah, segfault in the tests... Closing/Opening to attempt a re-run.

fstagni (dismissed):

andresailer (dismissed):

atsareg (dismissed):

PR #4736, opened 2020-09-08
[integration] dirac-install: use system tar for extracting tarballs

### LINKED ISSUE #4246: dirac-install fails to update with special permission files
Since DIRAC release v6r21 (we use v6r21p10 for now), dirac client update with dirac-install.py failed with "Bus error", in our environment.
I then found some special permission files from Externals-client-v6r6p8-Linux_x86_64_glibc-2.12-python27.tar.gz
e.g.
% ls -l Linux_x86_64_glibc-2.12/lib/libpython2.7.so.1.0
-r-xr-xr-x 1 hideki b2_belle2 1942984 Mar 22  2018 Linux_x86_64_glibc-2.12/lib/libpython2.7.so.1.0*

Since v6r21 dirac-install.py exploits tarfile module instead of tar command.
I confirmed tarfile.extract() fails if existing file has such permission while "tar xzf", previous way, does not show error.
First of all I'd know if this issue happens on other communities.
If yes, I think there are several options, e.g. remove existing file or change permission before extraction, replace directory after all files successfully extracted...

What do you think?

andresailer:
BEGINRELEASENOTES
*Core
FIX: dirac-install: use system tar command for extracting tarballs, fixes #4246 

ENDRELEASENOTES

### DIFF
--- a/Core/scripts/dirac-install.py
+++ b/Core/scripts/dirac-install.py
@@ -1532,24 +1532,11 @@ def downloadAndExtractTarball(tarsURL, pkgName, pkgVer, checkHash=True, cache=Fa
   #  tf.extract( member )
   # os.chdir(cwd)
   if not isSource:
-    try:
-      with closing(tarfile.open(tarPath, mode="r:*")) as tar:
-        for tarinfo in tar:  # pylint: disable=not-an-iterable
-          try:
-            tar.extract(tarinfo, cliParams.targetPath)  # pylint: disable=no-member
-          except IOError:
-            os.remove(tarinfo.name)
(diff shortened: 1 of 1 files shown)


chrisburr:
SLC6 is failing as we don't have tar in the docker image:

```bash
$ docker run --rm -it diracgrid/slc6-dirac:latest bash -c 'tar --help'
bash: tar: command not found
```

I think it's a reasonable requirement so I've made a PR for it: https://github.com/DIRACGrid/management/pull/20

fstagni (approved):

fstagni (approved):

atsareg (approved):

PR #4753, opened 2020-10-05
[v7r0] NotificationHandler: spam protection: send immediately, but not again too quickly

andresailer:
These changes avoid having to wait long for an email, but still avoids a deluge if the same email is sent separately. However, emails are no longer bundled together, if that was the intention of the previous implementation.

BEGINRELEASENOTES

*Framework
CHANGE: NotificationService: if ``avoidSpam=True`` the email is now sent immediately and cached for up to one hour, instead of being cached and then sent after up to an hour.

ENDRELEASENOTES


fstagni (commented):
I think it's fine.

fstagni (approved):

atsareg (approved):

PR #4809, opened 2020-11-05
[v7r1] Add CERNLDAPSyncPlugin

chrisburr:
Requires (though safe to merge beforehand):
* https://github.com/DIRACGrid/DIRACOS/pull/151 
* https://github.com/DIRACGrid/DIRAC/pull/4784

BEGINRELEASENOTES

*Configuration
NEW: Add CERNLDAPSyncPlugin for VOMS2CSAgent

ENDRELEASENOTES

### DIFF
--- a/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
+++ b/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
@@ -1,0 +1,84 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import ldap3
+
+
+class CERNLDAPSyncPlugin(object):
+  """Synchronization plugin for mapping new users to CERN accounts.
+
+  This plugin results in new users having two additional fields in the CS,
+  CERNAccountType and PrimaryCERNAccount. If the new nickname does have a
+  corresponding CERN account it will be rejected.
+  """
+  def __init__(self):
+    """Initialise the plugin and underlying LDAP connection."""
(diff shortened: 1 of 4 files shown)


atsareg:
This plugin adds some user info but how this extra info is supposed to be used by the VOMS2CSSyncronizer ?

chrisburr:
The information is put in the CS for other tools to read directly. For my use case this is entirely outside of DIRAC.

Even without using the extra fields this plugin is still useful for validating that nicknames are correct in VOMS before they start getting into the databases. I was planning to put this plugin in LHCbDIRAC but I was told this might be of interest to @andresailer?

chaen:
It would be lovely to have a documentation of all the `VOMS2CS` business, where we could add this feature

ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
```diff
@@ -0,0 +1,84 @@
...
+from __future__ import print_function
+
+import ldap3
+
+
+class CERNLDAPSyncPlugin(object):
+  """Synchronization plugin for mapping new users to CERN accounts.
+
+  This plugin results in new users having two additional fields in the CS,
+  CERNAccountType and PrimaryCERNAccount. If the new nickname does have a
```
andresailer, line 12:
```suggestion
  CERNAccountType and PrimaryCERNAccount. If the new nickname does not have a
```
right?

andresailer:
Yes, this could be useful for us.

atsareg (approved):

PR #4843, opened 2020-11-25
[v7r2] PY3: Fixes for X509Chain

chrisburr:
BEGINRELEASENOTES

*Python 3
FIX: Fix downloading X509 certificates

ENDRELEASENOTES

### DIFF
--- a/Core/Security/m2crypto/X509Chain.py
+++ b/Core/Security/m2crypto/X509Chain.py
@@ -1055,7 +1055,7 @@ class X509Chain(object):
       return S_OK(self.__hash)
     sha1 = hashlib.sha1()
     for cert in self._certList:
-      sha1.update(str(cert.getSubjectNameObject()))
+      sha1.update(str(cert.getSubjectNameObject()["Value"]))
     sha1.update(str(self.getRemainingSecs()['Value'] / 3600))
     sha1.update(self.getDIRACGroup()['Value'])
     if self.isVOMS():
--- a/Core/Security/m2crypto/X509Chain.py
+++ b/Core/Security/m2crypto/X509Chain.py
@@ -234,7 +234,7 @@ class X509Chain(object):
     """
     try:
       self._certList = self.__certListFromPemString(data)
-    except BaseException as e:
+    except Exception as e:
       return S_ERROR(DErrno.ECERTREAD, "%s" % repr(e).replace(',)', ')'))
 
     if not self._certList:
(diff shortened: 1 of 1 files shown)


Core/Security/m2crypto/X509Chain.py
```diff
@@ -1050,20 +1053,20 @@ def hash(self):
 
         :returns: S_OK(string hash)
     """
-
     if self.__hash:
       return S_OK(self.__hash)
     sha1 = hashlib.sha1()
     for cert in self._certList:
-      sha1.update(str(cert.getSubjectNameObject()))
```
chrisburr, line 1058: @chaen Currently this line is broken as `cert.getSubjectNameObject()` returns a `S_OK` object, i.e.

```python
(Pdb) str(cert.getSubjectNameObject()["Value"])
'/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=cburr/CN=761704/CN=Chris Burr/CN=5399599166/CN=2787290085/CN=8974817748'
(Pdb) str(cert.getSubjectNameObject())
"{'OK': True, 'Value': <M2Crypto.X509.X509_Name object at 0x7f24f0461ca0>}"
```

Should it be fixed in `rel-v7r0`?

chaen: yeah I saw that. It can be but it is not strictly necessary. As per comment, this is used exclusively to generate a hash to store in a dict. So as long as it is deterministic and unique, it is fine. (yes, I am sure that there would have been much better way of doing :-D )

chrisburr: I doubt the pointer (`0x7f24f0461ca0`) is deterministic?

Core/Security/m2crypto/X509Chain.py
```diff
@@ -252,6 +252,9 @@ def __certListFromPemString(certString):
     # To get list of X509 certificates (not X509 Certificate Chain) from string it has to be parsed like that
     # (constructors are not able to deal with big string)
     certList = []
+    # If the certificate is downloaded from the server it will be a str in Python 3
+    if six.PY3 and isinstance(certString, six.string_types):
```
chaen, line 256: I understand and agree that this needs to be done, but should it really be hidden there, or should we rather do it upon reception by the server ?

chrisburr: I'm not sure where it can be better put as it would have to be fixed wherever RPC calls are made to `Framework/ProxyManager`, for which there is at least 4 places in `ProxyManagerClient`. Plus if you get it wrong the error gets lost in DIRAC's error handling making it a pain to debug.

As it stands this is still technically "correct" as the string is a still a valid representation of the certificate chain. In other Python 3 libraries it's quite common for interfaces like this to be permissive and accept both `bytes` and `str` where it can be done safely and without ambiguities.

chrisburr: > upon reception by the server

For context, this crashes on the client when running the LHCb "submit jobs as different users" test where a admin user is explicitly downloading data from the ProxyDB.

Core/Security/m2crypto/X509Chain.py
chaen: Good point, I overlooked the pointer :-) Then yes, please backport it :-)

Core/Security/m2crypto/X509Chain.py
chaen: ok then good enough for me :-)

chaen (dismissed):

chrisburr:
@chaen Please re-approve, I've rebased so there isn't a merge conflict with #4845

chaen (approved):

fstagni (approved):

PR #4850, opened 2020-11-26
[v7r2] Hackathon fixes: logging and tests

chaen:
BEGINRELEASENOTES
*Framework
FIX: MQ logging sends also the message,asctime and levelname field

ENDRELEASENOTES

### DIFF
--- a/FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
+++ b/FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
@@ -37,6 +37,9 @@ class MessageQueueHandler(logging.Handler):
     result = createProducer(queue)
     if result['OK']:
       self.producer = result['Value']
+    else:
+      # print because logging not available
+      print("ERROR initializing MessageQueueHandler: %s" % result)
     self.hostname = socket.gethostname()
 
   def emit(self, record):
--- a/Resources/LogBackends/MessageQueueBackend.py
+++ b/Resources/LogBackends/MessageQueueBackend.py
@@ -14,6 +14,9 @@ from DIRAC.FrameworkSystem.private.standardLogging.LogLevels import LogLevels
 from DIRAC.Resources.LogBackends.AbstractBackend import AbstractBackend
 
 DEFAULT_MQ_LEVEL = 'verbose'
+# These are the standard logging fields that we want to see
+# in the json. All the non default are printed anyway
+DEFAULT_FMT = '%(levelname)s %(message)s %(asctime)s'
 
 
 class MessageQueueBackend(AbstractBackend):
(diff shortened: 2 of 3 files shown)


Resources/LogBackends/MessageQueueBackend.py
```diff
@@ -32,6 +35,14 @@ def __init__(self, backendParams=None):
     """
     Initialization of the MessageQueueBackend
     """
+    # The `Format` parameter is passed as `fmt` to libJsonFormatter
+    # which uses it to know which "standard" fields to keep in the
+    # json output. So we need these
+    if not backendParams:
+      backendParams = {'Format': DEFAULT_FMT}
+    elif 'Format' not in backendParams:
+      backendParams['Format'] = DEFAULT_FMT
```
chrisburr, line 44: Rather that duplicating, how about:
```suggestion
    if backendParams is None:
      backendParams = {}
    if 'Format' not in backendParams:
      backendParams['Format'] = DEFAULT_FMT
```
or more concisely:
```suggestion
    if backendParams is None:
      backendParams = {}
    backendParams.setdefault("Format", DEFAULT_FMT)
```

FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, queue):
     result = createProducer(queue)
     if result['OK']:
       self.producer = result['Value']
+    else:
+      # print because logging not available
+      print("ERROR initializing MessageQueueHandler: %s" % result)
```
chrisburr, line 42:
```suggestion
      print("ERROR initializing MessageQueueHandler:", result)
```

chaen: What's wrong with the `%s` ?

Resources/LogBackends/MessageQueueBackend.py
chaen: Good idea

FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
chrisburr: I think it's better style to use multiple arguments to `print` as `%` formatting is longer, harder to lint and more likely to result in an exception due to accidentally passing the wrong number of parameters.

That said obviously in this case it makes no difference as I'm 99% sure it's correct as-is so I'm happy either way.

chaen: yeah, the multitude of way to format string does not really help... between the C-style, the multiple number of arguments, and the `.format`....

aldbr (approved):

FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
chrisburr: You missed the one true way! (`f"Hello {name}"`) 😆

chrisburr (approved):

FrameworkSystem/private/standardLogging/Handler/MessageQueueHandler.py
chaen: True ! :D

atsareg (approved):

PR #4876, opened 2020-12-18
Fix bad move in #4867

chrisburr:


fstagni (approved):

atsareg:
I will make v7r20-pre26 as soon as the tests are OK

PR #4894, opened 2021-01-14
[v7r0] Modify __init__.py to avoid conflicts with rel-v7r2

chrisburr:
This isn't strictly needed but I think it will make the conflicts with #4865 easier to handle.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -88,12 +88,15 @@ __RCSID__ = "$Id$"
 import _strptime
 
 
-# Define Version
-
-majorVersion = 7
-minorVersion = 0
-patchLevel = 44
-preVersion = 0
+# Define Version, use an unsual structure to minimise conflicts with rel-v7r2
+pythonVersion = pyPlatform.python_version_tuple()
+if pythonVersion[0] == "3":
+  pass
+else:
+  majorVersion = 7
+  minorVersion = 0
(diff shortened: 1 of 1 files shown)


fstagni (approved):

__init__.py
```diff
@@ -88,12 +88,15 @@
...
+pythonVersion = pyPlatform.python_version_tuple()
+if pythonVersion[0] == "3":
+  pass
+else:
+  majorVersion = 7
+  minorVersion = 0
+  patchLevel = 44
+  preVersion = 0
 
 version = "v%sr%s" % (majorVersion, minorVersion)
```
atsareg, line 101: If the (majorVersion, minorVersion) are not defined above (pythonVersion[0] == "3") this line will fail. This show up in the test now. I guess this is ot a problem yet since we do not yet use python3 but somehow this should be corrected

chrisburr: It's intended to be combined with https://github.com/DIRACGrid/DIRAC/pull/4865

atsareg: I see. Do you think that #4865 is ready to be merged ? Since I have not yet went through it I did not merge it as formally it is not yet approved. I have made a simple fix in the integration DIRAC/__init__.py to let the tests pass for the moment

PR #4906, opened 2021-01-20
[v7r2] Deploy releases to PyPI

chrisburr:
This PR:
* improves the setuptools metadata slightly
* automatically uploads releases to PyPI

To facilitate the upload to PyPI, a CI job on all tags which converts the DIRAC-style version (`v7r2-pre1`) to a PEP-440 style version (`v7.2.0a1`). This is pushed to `DIRACGrid/DIRAC` before building the source and binary distributions for PyPI. Once we drop support for Python two we can stop the dance with having two tags per version and just create the PEP-440 style version directly.

BEGINRELEASENOTES

*Python 3
NEW: Deploy releases to PyPI

ENDRELEASENOTES


fstagni:
Why moving unit tests in a separate base directory?

chrisburr:
I got distracted thinking about how to make the unit tests compatible with a `pip install`ed installation (i.e. deal with #4859) but it's a task for a separate PR. (I've just moved that work to a separate branch.)

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: "ubuntu-latest"
+    if: github.event_name != 'push' || github.repository == 'DIRACGrid/DIRAC'
+    defaults:
+      run:
+        shell: bash -l {0}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          git fetch --prune --unshallow
+      - uses: conda-incubator/setup-miniconda@master
```
fstagni, line 17: Would you add a comment on what is this `conda-incubator`?

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          environment-file: environment-py3.yml
+          miniforge-variant: Mambaforge
```
fstagni, line 20: A comment on what is `Mambaforge`? (and why this variant is needed here)

.github/workflows/make_release.py
```diff
@@ -0,0 +1,73 @@
...
+            "target_commitish": commit_hash,
+            "body": release_notes,
+            "draft": True,
+            "prerelease": Version(version).is_prerelease,
+        },
+        headers=headers,
+    )
+    r.raise_for_status()
+    release_data = r.json()
+    print(f"Created draft release at: {release_data['html_url']}")
```
fstagni, line 35: This format is python 3 only, IIUC (f-strings).
And that is fine, of course, but in general, for our own sanity, can you use the old style that we are all used to?

fstagni: This applies in a few places in this file.

README.rst
```diff
@@ -118,12 +118,6 @@ Unit tests are provided within the source code. Integration, regression and syst
...
-~~~~~~~
+~~~~~~~~~~~~~~~~
 
 This work is co-funded by the EOSC-hub project (Horizon 2020) under Grant number 777536
-
-.. raw:: html 
-  <img src="https://wiki.eosc-hub.eu/download/attachments/1867786/eu%20logo.jpeg?version=1&modificationDate=1459256840098&api=v2" height="24"/>
-
-.. raw:: html 
-  <img src="https://wiki.eosc-hub.eu/download/attachments/18973612/eosc-hub-web.png?version=1&modificationDate=1516099993132&api=v2" height="24"/>
```
fstagni, line 129: Why removing the aboves?

chrisburr: It seems this was added to workaround the fact that GitHub refuses to support scaling images in RST (https://github.com/github/markup/issues/295). Though if I look in the README on GitHub it doesn't even work and the images aren't displayed.

PyPI refuses to accept packages containing invalid RST or those containing disabled features (mostly raw HTML due to security concerns).

Other options:
* Use markdown instead of rst for the README
* Have a separate README for PyPI
* Resize the images manually and host them somewhere else

.github/workflows/make_release.py
chrisburr: This function is copied from DIRACOS2 and I don't want to start changing it's contents as it will just make maintenance a hassle if anything needs to be changed.

We also already have f-strings in [management](https://github.com/DIRACGrid/management/blob/master/dirac-distribution/WebAppCompiler.py) and the moment we can drop Python 2 we should starting using them in DIRAC. I know they're odd at first sight but you quickly become accustomed to them and they feel like the obvious and most readable way of formatting string. There are also technical benefits (e.g. much faster performance/better static analysis).

README.rst
fstagni: Use markdown...?

chrisburr: Sure, I'll convert it after lunch

.github/workflows/deployment.yml
chrisburr: Done. This is all about making the jobs faster (mamba saves maybe a minute or two vs conda) and the configuration easier to maintain than running the commands directly.

At some point I'll unify this with the other CI workflows so they all look similar.

README.rst
chrisburr: Actually GitHub has removed support for the syntax I was going to use so it's actually impossible to resize an image in a way which will work on both GitHub and PyPI.

Given that, I think our best solution would be to commit it somewhere under docs (it's only a few kB).

fstagni: ..whatever works is fine...

chrisburr: Done, assuming I renamed the repo/branch in the URL correctly it should now look like this once merged:

Screenshot 2021-01-22 at 17 09 57

andresailer (approved):
Looks good to me

fstagni (approved):

chrisburr:
@fstagni Did you add the credentials to the the GitHub configuration?

fstagni:
Yes. Do I need to do the steps also for TEST_PYPI_API_TOKEN  (point 4 of https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github ?)

chrisburr:
Thanks!

> Yes. Do I need to do the steps also for TEST_PYPI_API_TOKEN (point 4 of https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github ?)

Nope, I didn't bother with setting up test-pypi. =

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Check tag is for v7r2 or later
+        id: check-tag
+        run: |
+          if [[ "${{ github.event.ref }}" =~ ^refs/tags/v7r2(p[0-9]+)?(-pre[0-9]+)?$ ]]; then
+              echo ::set-output name=create-release::true
+          fi
+      - name: Make PEP-440 style release on GitHub
+        if: steps.check-tag.outputs.create-release == 'true'
+        run: |
+          OLD_STYLE=${GITHUB_REF##*/}
```
chaen, line 37: This is the standard bash parameter expansion or is it some github action syntax ?

chrisburr: Standard bash parameter expansion using the recommended way to do this in the GitHub docs

PR #4929, opened 2021-01-29
[v7r2] Python 3 compatibility fixes found when porting LHCbDIRAC

chrisburr:


src/DIRAC/Resources/Catalog/PoolXMLCatalog.py
```diff
@@ -162,15 +162,13 @@ def __init__(self, xmlfile=''):
       if isinstance(xmlfile, list):
         for xmlf in xmlfile:
           try:
-            _sfile = file(xmlf, 'r')
```
chrisburr, line 165: I'm not convinced this line is needed but I'm not sure. Perhaps it's supposed to be deleting + recreating the file if it already exists? i.e.

```python
with open(xmlf):
   pass
```

Same applies slightly lower down.

chrisburr, line 165: Actually, is this supposed to be checking if the file exists?

chrisburr: I've tried to simplify it in 943cc7219

fstagni (approved):

PR #4942, opened 2021-02-04
[v7r2] Revert scripts to before Python 3 incompatible merge

chrisburr:


chrisburr:
@atsareg Please include this in today's tag, without it the Python 3 installs are completely broken. I believe @TaykYoku will reapply the changes from https://github.com/DIRACGrid/DIRAC/pull/4865 in a separate PR. Most were lost in the merge of `rel-v7r1` to `integration` so reverting these as well doesn't make much difference.

PR #4951, opened 2021-02-08
Revert "[v7r1] Absolute import to avoid conflicts with system modules"

chaen:
Reverts DIRACGrid/DIRAC#4919

These changes have nasty side effects wherever dynamic loading is done. It broke the LHCbDIRAC CI.  Example of such places is https://github.com/DIRACGrid/DIRAC/blob/b47d9319debd0c0464660aa1ce3a15fe47a51043/Core/Utilities/Graphs/Graph.py#L280
Instead of trying to hunt down every little corner like that one (which is anyway impossible), I'd rather just revert and wait for v7r2 to be out, with its proper implementation of absolute import. 
We have lived like this for many years, so it shouldn't be a problem.

Please merge asap, we are blocked in our releases because broken CI

### DIFF
--- a/Core/Utilities/CFG.py
+++ b/Core/Utilities/CFG.py
@@ -1,7 +1,6 @@
 """ This is the main module that interprets DIRAC cfg format
 """
 
-from __future__ import absolute_import
 from __future__ import print_function
 
 __RCSID__ = "$Id$"
--- a/Core/Utilities/DEncode.py
+++ b/Core/Utilities/DEncode.py
@@ -11,8 +11,6 @@ Encoding and decoding for dirac, Ids:
  t -> tuple
  d -> dictionary
 """
-
-from __future__ import absolute_import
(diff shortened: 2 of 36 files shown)


atsareg:
CI procedure is to reveal this kind of problems and fix them. This way you fix "your" problems by creating them for "others" defintely.

chaen:
I totally agree with you. And pre-releases are for that kind of changes, not patch releases. And again, what you are trying to achieve is already done properly in the next release, so why hurrying it in now ?

atsareg:
If the v7r2 release is for tomorrow, I would be happy to revert it. BTW, the absolute import in v7r2 will also create problems like in your example that should be fixed ? Or they are already fixed ?

chaen:
well, you've seen the outcome of the last hachathon, so it does not seem so far off :-) 
@chrisburr believes it was done correctly in v7r2

PR #4955, opened 2021-02-09
[v7r1] Better error message when the SingularityCE image is not found

chrisburr:


Resources/Computing/SingularityComputingElement.py
```diff
@@ -444,7 +444,7 @@ def submitJob(self, executableFile, proxy=None, **kwargs):
       cmd.extend([rootImage, innerCmd])
     else:
       # if we are here is because there's no image, or it is not accessible (e.g. not on CVMFS)
-      self.log.error('Singularity image to exec not found')
+      self.log.error('Singularity image to exec not found: %s' % rootImage)
```
atsareg, line 447:
```suggestion
      self.log.error('Singularity image to exec not found: ', rootImage)
```

atsareg (approved):

PR #4971, opened 2021-02-16
[Please merge now] [v7r2]Core:re-align dirac-install(s).py

chaen:
Because of these multiplications of dirac-install, they have diverged, and changes were lost. 
This bring them back, and align the two versions of this repo.

BEGINRELEASENOTES
*Core
FIX: re-align dirac-install(s).py

ENDRELEASENOTES


atsareg:
Fine with me. Do you need it in the distribution tar ball also ?

chrisburr (approved):

chaen:
Hum, I don't know :-D The reason I need them now is for some tests that fetch it directly from the integration branch, so simple merge works for me. Now I believe there is yet another dirac-install somewhere in the management repo. I don't want to touch taht one, as it will only create confusion. But let's discuss it on #4860

PR #4977, opened 2021-02-18
[v7r2] fix dirac-admin-sysadmin-cli

TaykYoku:
PR fix log message and resolve problem with unicode:
```
  File "/opt/dirac/cli/DIRAC/Core/DISET/private/BaseClient.py", line 471, in _connect
    retVal = transport.initAsClient()
  File "/opt/dirac/cli/DIRAC/Core/DISET/private/Transports/M2SSLTransport.py", line 145, in initAsClient
    self.oSocket.set_tlsext_host_name(host)
  File "/opt/dirac/cli/diracos/usr/lib64/python2.7/site-packages/M2Crypto/SSL/Connection.py", line 708, in set_tlsext_host_name
    m2.ssl_set_tlsext_host_name(self.ssl, name)
TypeError: in method 'ssl_set_tlsext_host_name', argument 2 of type 'char const *'
```

BEGINRELEASENOTES

*Core
FIX: encode serviceName

*Framework
FIX: fix log message about responded hosts in sysadmin cli

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/DISET/private/BaseClient.py
+++ b/src/DIRAC/Core/DISET/private/BaseClient.py
@@ -70,8 +70,8 @@ class BaseClient(object):
     if not isinstance(serviceName, six.string_types):
       raise TypeError("Service name expected to be a string. Received %s type %s" %
                       (str(serviceName), type(serviceName)))
-    self._destinationSrv = serviceName
-    self._serviceName = serviceName
+    self._destinationSrv = str(serviceName)
+    self._serviceName = str(serviceName)
     self.kwargs = kwargs
     self.__useCertificates = None
     # The CS useServerCertificate option can be overridden by explicit argument
--- a/src/DIRAC/Core/DISET/private/BaseClient.py
+++ b/src/DIRAC/Core/DISET/private/BaseClient.py
@@ -70,6 +70,7 @@ class BaseClient(object):
     if not isinstance(serviceName, six.string_types):
       raise TypeError("Service name expected to be a string. Received %s type %s" %
                       (str(serviceName), type(serviceName)))
+    # Explicitly convert to a str to avoid Python 2 M2Crypto issues with unicode objects
     self._destinationSrv = str(serviceName)
     self._serviceName = str(serviceName)
     self.kwargs = kwargs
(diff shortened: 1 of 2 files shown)


src/DIRAC/FrameworkSystem/Client/SystemAdministratorClientCLI.py
```diff
@@ -1295,7 +1296,7 @@ def do_showall(self, args):
                 records.append(record)
       printTable(fields, records, sortOption)
       if silentHosts:
-        print("\n %d out of %d hosts did not respond" % (len(silentHosts), len(respondingHosts)))
+        print("\n %d out of %d hosts did not respond" % (len(silentHosts), allHosts))
```
chrisburr, line 1299:
```suggestion
        print("\n %d out of %d hosts did not respond" % (len(silentHosts), totalHosts))
```

src/DIRAC/FrameworkSystem/Client/SystemAdministratorClientCLI.py
```diff
@@ -1253,6 +1253,7 @@ def do_showall(self, args):
     client = SystemAdministratorIntegrator()
     silentHosts = client.getSilentHosts()
     respondingHosts = client.getRespondingHosts()
+    allHosts = len(silentHosts) + len(respondingHosts)
```
chrisburr, line 1256: I would expect a variable named `allHosts` to contain a list of all hosts, how about:
```suggestion
    totalHosts = len(silentHosts) + len(respondingHosts)
```

src/DIRAC/Core/DISET/private/BaseClient.py
```diff
@@ -70,8 +70,8 @@ def __init__(self, serviceName, **kwargs):
     if not isinstance(serviceName, six.string_types):
       raise TypeError("Service name expected to be a string. Received %s type %s" %
                       (str(serviceName), type(serviceName)))
-    self._destinationSrv = serviceName
-    self._serviceName = serviceName
+    self._destinationSrv = str(serviceName)
```
chaen, line 73: Please add a comment that this is for python3 compatibility. 
One day, we may want to cleanup...

chrisburr: As the issue is Python 2 specific and has nothing to do with Python 3:
```suggestion
    # Explicitly convert to a str to avoid Python 2 M2Crypto issues with unicode objects
    self._destinationSrv = str(serviceName)
```

chrisburr (approved):

fstagni (approved):

PR #4990, opened 2021-02-24
[v7r2]  Run Python 3 pylint and fix associated issues 

chrisburr:


src/DIRAC/Core/Utilities/File.py
```diff
@@ -209,39 +209,6 @@ def getGlobbedFiles(files):
   return globbedFiles
 
 
-def getCommonPath(files):
```
chrisburr, line 212: I'm 90% sure this is equivalent to `os.path.commonprefix`?

chaen: How DARE you use standard libraries ?!?! :fearful:

chrisburr: How DARE pylint find bugs in the DIRAC implementation that I can't be bothered to fix?! 😄

chaen: I think pylint should be fixed

src/DIRAC/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
```diff
@@ -76,6 +76,7 @@ def _getUserInfo(self, commonName):
     )
     if not status:
       raise ValueError("Bad status from LDAP search: %s" % result)
+    response = list(response)
```
chaen, line 79: why do we need the cast here ?

src/DIRAC/Core/Utilities/test/Test_List.py
```diff
@@ -57,7 +57,7 @@ def testRandomize(self):
     randList = List.randomize(aList)
     self.assertEqual(randList, [])
     # non empty
-    aList = [1, 2, 3]
+    aList = ["1", "2", "3"]
```
chaen, line 60: why changing the test  ?

src/DIRAC/WorkloadManagementSystem/Utilities/ParametricJob.py
```diff
@@ -51,7 +51,7 @@ def getParameterVectorLength(jobClassAd):
       else:
         nThisParValues = jobClassAd.getAttributeInt(attribute)
       if nParValues is not None and nParValues != nThisParValues:
-        return S_ERROR(EWMSJDL, "Different length of parameter vectors: for %s, %d != %d" % (attribute,
+        return S_ERROR(EWMSJDL, "Different length of parameter vectors: for %s, %s != %d" % (attribute,
```
chaen, line 54: why ? It is supposed to be an `int` too no ?

.github/workflows/basic-python3.yml
```diff
@@ -7,6 +7,9 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name != 'push' || github.repository == 'DIRACGrid/DIRAC'
     timeout-minutes: 30
+    defaults:
+      run:
+        shell: bash -l {0}
```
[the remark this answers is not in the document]
chrisburr: So the `conda` activation scripts are ran before each step

src/DIRAC/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
chrisburr: Pylint thought `response` was a generator and can't be indexed a few lines lower

src/DIRAC/Core/Utilities/test/Test_List.py
chrisburr: It confuses pylint and makes it think that `List.randomize` always returns a list of `int`. It doesn't affect the actual test so I think it's harmless and better than disabling some rules.

src/DIRAC/WorkloadManagementSystem/Utilities/ParametricJob.py
chrisburr: It starts off as `nParValues = None` and pylint isn't smart enough to understand it definitely isn't `None` anymore because of the `if`. I'm pretty sure this change is harmless for the string and still allows it to validate the other parameters in the string to still be validated.

src/DIRAC/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
chaen: so why not just disabling the rule of the offending line ? In any case, a comment would surely help, otherwise it looks like it's part of t he logic

.pylintrc
```diff
@@ -32,7 +32,7 @@ load-plugins=
 #R0903 = Too few public methods (%s/%s)
 #C0326 = bad-whitespace
 #I0011 = locally-disabling
-disable=R0903,C0326,I0011,redefined-variable-type,c-extension-no-member
+disable=R0903,C0326,I0011,redefined-variable-type,c-extension-no-member,W,C,R
```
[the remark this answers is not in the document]
chrisburr: We only ever run `pylint -E` in CI so this does the same. At some point we should reassess these settings and clean-up a lot of useless `# pylint: disable=xxxx` comments but that's a task for another day...

.pylintrc
```diff
@@ -137,15 +137,16 @@ ignore-docstrings=yes
 ignore-mixin-members=yes
 
 # List of classes names for which member attributes should not be checked
-# (useful for classes with attributes dynamically set).
-ignored-classes=SQLObject
+ignored-classes=SQLObject,Session
```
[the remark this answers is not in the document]
chrisburr: 👍 

```suggestion
# Skip some sqlalchemy objects which cause trouble 
ignored-classes=SQLObject,Session
```

src/DIRAC/ConfigurationSystem/Client/SyncPlugins/CERNLDAPSyncPlugin.py
chrisburr: Somehow I persuaded myself `pylint` was correct but it wasn't. Fixed.

.github/workflows/basic-python3.yml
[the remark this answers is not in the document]
atsareg: Adding an appropriate comment here ?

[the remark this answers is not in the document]
chrisburr: Done

src/DIRAC/Core/Utilities/test/Test_File.py
```diff
@@ -107,10 +108,10 @@ def testGetMD5ForFiles():
 def testGetCommonPath():
```
fstagni, line 108: This test, I guess, passes. Does it really need to be here? (we are not testing the standard library...)

src/DIRAC/Core/Utilities/Graphs/PieGraph.py
```diff
@@ -119,9 +119,8 @@ def pie(self, explode=None,
                            horizontalalignment=halign,
                            verticalalignment=valign)
 
-          t.set_family(self.prefs['font_family'])
```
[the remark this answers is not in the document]
chrisburr: `t.set_family` is an alias for `t.set_fontsize` and was removed in matplotlib 3 (the first Python 3 only release): https://github.com/matplotlib/matplotlib/blob/v2.2.x/lib/matplotlib/text.py#L1043-L1045

.pylintrc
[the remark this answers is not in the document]
andresailer: We run `pylint -E` because those are things that need to be fixed because they are broken.  (apart from pylint acting strangely)
We cannot run without `-E` because there are too many things being reported otherwise. (unless we can run pylint on a diff)
The .pylintrc isn't only for the CI though. Developers should also take care of fixing warnings (maybe at least be aware) when they develop new things, so we shouldn't disable all of them in the standard pylint config.

src/DIRAC/Core/Utilities/test/Test_File.py
chrisburr: I mostly left it to make sure the standard library did what DIRAC expected rather than something subtly different. Now it's passed once I agree it can be removed.

src/DIRAC/WorkloadManagementSystem/Utilities/ParametricJob.py
chaen: but then I would advocate to disable locally pylint rather than changing the actual code to please a static analyzer

chrisburr: If it's disabled locally it would miss bugs like:
```suggestion
        return S_ERROR(EWMSJDL, "Different length of parameter vectors: for %d, %s != %d" % (attribute,
```
There are limits to how many changes you should make (https://github.com/DIRACGrid/DIRAC/pull/4990#discussion_r582139236 is an example of too far) but this seems harmless enough.

.pylintrc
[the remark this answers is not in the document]
chrisburr: Done

fstagni (approved):

atsareg (approved):

PR #4992, opened 2021-02-24
[v7r2] Setup.cfg: split server scripts from the rest

chaen:
Closes https://github.com/DIRACGrid/DIRAC/issues/4915

I don't think it needs release notes

@chrisburr I could not find any instructions/tests that would require update to specify `[server]` in the `pip install` commands. Am I right or did I miss it ?


chaen:
@chrisburr I don't think the `Deployment` CI is meant to be there, is it ?

chrisburr:
> @chrisburr I don't think the `Deployment` CI is meant to be there, is it ?
> It did show a typo though :D

It is, to ensure that we're able to successfully build wheels rather than finding out during deployment and having a broken tag. Only the tagging and upload steps are skipped.

chaen:
> > @chrisburr I don't think the `Deployment` CI is meant to be there, is it ?
> > It did show a typo though :D
> 
> It is, to ensure that we're able to successfully build wheels rather than finding out during deployment and having a broken tag. Only the tagging and upload steps are skipped.

Ah ! Sorry, I was tricked by the name :-)

chrisburr (dismissed):

atsareg:
I would add also:
  dirac-install-db
  dirac-install-component
  dirac-install-tornado-service
  dirac-populate-component-db

chaen:
> I would add also:
> dirac-install-db
> dirac-install-component
> dirac-install-tornado-service
> dirac-populate-component-db

from what I could see, these can be launched from anywhere, not just the local machine. Or am I mistaken ?

atsareg:
I think this is not only the matter of where it can be executed from. These are the commands to operate on the server installation. So, looks logical to me that they should be execute either on the server host (in most if not all the cases) or by somebody who knows what he is doing, so having a "server" client. 
And in fact, most of them (except  dirac-populate-component-db) must be executed on the local machine.

chaen:
Indeed, I read the code to quickly.
Now I disagree with your interpretation. To me `server` installation is what needs to be on the server. If indeed we want a basic installation and one with powerful admin command, that's a different matter, and that is what @chrisburr proposed in #4915 I was under the impression that is was a bit useless, especially that we do not want to maintain several installations.
If more people want the admin installation, we can do the split (LHCb can always install that on CVMFS). However, if that is the case, people should react on the issue in a timely manner

chaen:
And btw, executing these commands from the host is always highly impractical to me, because you always end up using the server `dirac.cfg`, which enforces using the host certificate. So you need to pass extra options to use your proxy, It's a detail, but an annoying one to me. So I'd rather do it from another machine when I can

chaen:
I merged @fstagni so I guess it's good as is  ?

fstagni (approved):

chrisburr (approved):

PR #5026, opened 2021-03-09
[v7r2] Remove DIRAC.Core.Utilities.ExitCallback

chrisburr:
This fixes the issue where if you launch `ipython` and run:

```python
import DIRAC
import time
time.sleep(60)
```

and press `Control+C` before it finishes the entire process is killed instead of just being interrupted. I don't see any good reason for using signal handlers so they should just go.

BEGINRELEASENOTES

*Core
CHANGE: Remove DIRAC.Core.Utilities.ExitCallback

ENDRELEASENOTES


chaen:
I would assume that these were done because of the very fragile multi threaded/process environment we run into (`pyGSI` and the like). So rather die that half dead with mysterious behavior. But I am happy to see them go anyway.

fstagni (approved):
Fine for me.

PR #5029, opened 2021-03-10
[v7r2] Re-add thousands separators to PrettyScalarFormatter

chrisburr:
In #4990 I removed `PrettyScalarFormatter.pprint_val` as it is no longer available in recent matplotlib versions. This breaks the LHCbDIRAC tests due to the change of image. While it's only cosmetic I think it's useful so I've re-added it in a way which should be compatible with all matplotlib releases.

cc @petricm 

BEGINRELEASENOTES

*Core
FIX: Display thousands separators in PrettyScalarFormatter again

ENDRELEASENOTES


chrisburr:
This PR still needs some work to make the tests happy, it's a purely cosmetic bug in the webapp so it can be ignored for the v7r2 release

chrisburr:
This is ready for review now. This corner could use some work still but I think this is enough for now.

fstagni (approved):

chrisburr:
@atsareg Is there a reason this wasn't merged into the latest release?

PR #5032, opened 2021-03-12
Validate v7r2 pre39

atsareg:


chrisburr:
The warning in the docs is:
```
/home/docs/checkouts/readthedocs.org/user_builds/dirac/checkouts/5032/docs/source/AdministratorGuide/Resources/storage.rst:313: WARNING: undefined label: dirac-dms-protocol-matrix
```

src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
```diff
@@ -48,6 +48,15 @@
 from DIRAC.Core.Base import Script
 from DIRAC.Core.Utilities.DIRACScript import DIRACScript
 
+Script.registerSwitch('', 'FromSE=', 'SE1[,SE2,...]')
```
chrisburr, line 51: This will need to be inside `main`

atsareg: Yes. Script.setUsageMessage(... should also go inside main ?

atsareg: Or it can be simply scrapped ?

chrisburr: Everything with `Script.xxxx` must go inside `main`. Actually I think the Script stuff is broken again as `from DIRAC.Core.Base import Script` is setting it globally...

atsareg: I see other scripts where import Script is outside main()

atsareg: The doc build is successful. Should the script be still fixed to bring everything inside main ?

chrisburr: The docs are still wrong when you call `--help` (see below). The issue is from https://github.com/DIRACGrid/DIRAC/pull/4958:

https://github.com/DIRACGrid/DIRAC/blob/00f9c4ff0323cdc5d80311887722fb1f56aa37e4/src/DIRAC/Core/Base/Script.py#L22-L23

Examples:

```
$ dirac-info --help

Core implementation of import.

This module is NOT meant to be directly imported! It has been designed such
that it can be bootstrapped into Python as the implementation of import. As
such it requires the injection of specific modules and attributes in order to
work. One should use importlib as the public-facing version of this module.


Usage:

  dirac-info [options] ...

General options:
  -o  --option <value>         : Option=value to add
  -s  --section <value>        : Set base section for relative parsed options
  -c  --cert <value>           : Use server certificate to connect to Core Services
  -d  --debug                  : Set debug mode (-ddd is extra debug)
  -   --cfg=                   : Load additional config file
  -   --autoreload             : Automatically restart if there's any change in the module
  -   --license                : Show DIRAC's LICENSE
  -h  --help                   : Shows this help

Options:
  -v  --version                : print version of current DIRAC installation
  -p  --platform               : print platform of current DIRAC installation
```

```
$ dirac-proxy-info --help

Core implementation of import.

This module is NOT meant to be directly imported! It has been designed such
that it can be bootstrapped into Python as the implementation of import. As
such it requires the injection of specific modules and attributes in order to
work. One should use importlib as the public-facing version of this module.


Usage:

  dirac-proxy-info [options] ...

General options:
  -o  --option <value>         : Option=value to add
  -s  --section <value>        : Set base section for relative parsed options
  -c  --cert <value>           : Use server certificate to connect to Core Services
  -d  --debug                  : Set debug mode (-ddd is extra debug)
  -   --cfg=                   : Load additional config file
  -   --autoreload             : Automatically restart if there's any change in the module
  -   --license                : Show DIRAC's LICENSE
  -h  --help                   : Shows this help

Options:
  -f  --file <value>           : File to use as user key
  -i  --version                : Print version
  -n  --novoms                 : Disable VOMS
  -v  --checkvalid             : Return error if the proxy is invalid
  -x  --nocs                   : Disable CS
  -e  --steps                  : Show steps info
  -j  --noclockcheck           : Disable checking if time is ok
  -m  --uploadedinfo           : Show uploaded proxies info
```

chrisburr: If you look in the readthedocs all scripts have the same description: https://dirac--5032.org.readthedocs.build/en/5032/UserGuide/CommandReference/WorkloadManagement/index.html

src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
```diff
@@ -48,13 +48,18 @@
 from DIRAC.Core.Base import Script
 from DIRAC.Core.Utilities.DIRACScript import DIRACScript
 
+Script.setUsageMessage('\n'.join([__doc__,
+                                  'Usage:',
+                                  ' %s [option|cfgfile]  % Script.scriptName']))
```
TaykYoku: With #4958 there is no need for that

src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
TaykYoku: I'll take a look

src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
atsareg: OK, I have scrapped it

src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
atsareg: It looks like the \_\_doc\_\_ strings are not added to the command descriptions if there is not setUsageMessage() call

TaykYoku: @chrisburr I'm trying to reproduce this wrong help output.. how I can get it?

PR #5049, opened 2021-03-18
[v7r2] newer versions of setuptools_scm do not support py2

petricm:
BEGINRELEASENOTES
*tests
CHANGE: fix version, newer versions of setuptools_scm do not support py2

ENDRELEASENOTES

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -60,3 +60,5 @@ caniusepython3
 subprocess32
 flaky
 ldap3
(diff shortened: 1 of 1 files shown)


petricm:
This is kind of a blocker for CIs all around as many places the DIRAC requirements are used, would be good to be merged soon.

fstagni (approved):

PR #5062, opened 2021-03-26
[v7r1] Move cpu work left computation at the beginning of the jobagent execution

aldbr:
The CPU work left computation was performed at the end of cycle *N* and the value was used in cycle *N+1*. 
If no slot is available or no job is matched for a while, the `timeleft` value is not updated and wrong.

BEGINRELEASENOTES
*WorkloadManagement
FIX: Move the CPU work left computation at the beginning of the JobAgent execution

ENDRELEASENOTES


phicharp:
Thanks Alexandre! I had been asking for this for so many years!

fstagni (approved):

PR #5069, opened 2021-03-29
[v7r0] Interface: correctly strip arguments

chaen:
One of our users did 
```
dirac-wms-job-get-output XXX, YYY, ZZZ...
```

The net result was 

```
ERROR : invalid literal for int() with base 10: ''
```

and an empty current directory !! The reason is that `parseArguments` was returning an empty string in the list, which then led to the current directory recursively being removed

BEGINRELEASENOTES
*Interface
FIX: parseArguments strip arguments correctly

ENDRELEASENOTES


Interfaces/API/Dirac.py
```diff
@@ -63,7 +63,7 @@
 def parseArguments(args):
   argList = []
   for arg in args:
-    argList += arg.split(',')
+    argList += [a.strip() for a in arg.split(',') if a]
```
chrisburr, line 66: Just in case someone manages to accidentally pass a variable containing only whitespace?
```suggestion
    argList += [a.strip() for a in arg.split(',') if a.strip()]
```

atsareg:
Did you converge on this, guys ?

Interfaces/API/Dirac.py
chaen: even better

atsareg (approved):

PR #5077, opened 2021-03-31
[v7r1] Fix for the case where no subprocess32 module found

atsareg:
For SSH hosts as CEs with old versions python interpreters subprocess32 module does not exist,  in this case subprocess module is as good

BEGINRELEASENOTES

*Resources
FIX: Host.py - Fix for the case where no subprocess32 module found

ENDRELEASENOTES

### DIFF
--- a/Resources/Computing/BatchSystems/Host.py
+++ b/Resources/Computing/BatchSystems/Host.py
@@ -16,7 +16,7 @@ import shutil
 import signal
 try:
   import subprocess32 as subprocess
-except:
+except ImportError:
   import subprocess
 import stat
 import json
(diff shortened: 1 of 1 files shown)


chrisburr:
Can you explain a little more here? Shouldn't this only ever be running with DIRACOS's Python interpreter?

chrisburr:
Ah I see [how this works now](https://github.com/DIRACGrid/DIRAC/blob/aaeb94f498a23778948b3739ba20666cfa147f08/src/DIRAC/Resources/Computing/SSHComputingElement.py#L421-L441). In that case this should just use `import subprocess` and never try to use `subprocess32`. I think this should also target `rel-v7r0`?

atsareg:
Well, the problem was encountered on an installation with v7r1, I do not want to favor v7r0  installations :)

fstagni (approved):

PR #5087, opened 2021-04-06
[v7r0] Fix undefined variables in RabbitMQAdmin

chrisburr:
BEGINRELEASENOTES

*Core
FIX: Undefined variable error when RabbitMQAdmin fails to connect

ENDRELEASENOTES

### DIFF
--- a/Core/Utilities/RabbitMQAdmin.py
+++ b/Core/Utilities/RabbitMQAdmin.py
@@ -25,12 +25,10 @@ def executeRabbitmqctl(arg, *argv):
   """
   command =['sudo','/usr/sbin/rabbitmqctl','-q', arg] + list(argv)
   timeOut = 30
-  result = Subprocess.systemCall(timeout = timeOut, cmdSeq = command)
-  if result['OK']:
-    errorcode, cmd_out, cmd_err = result['Value']
-  else:
-    return S_ERROR(errno.EPERM, "%r failed, status code: %s stdout: %r stderr: %r" %
-                                (command, errorcode, cmd_out, cmd_err) )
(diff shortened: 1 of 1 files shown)


chrisburr:
The tests are failing as `joinedload_all` has been removed from SQLAlchemy 1.4. See also #5071.

chaen:
This was actually fixed by you in 8d9747c6fd405aa4edde5702f16a5373d5dc459c
The problem here is that we are running the CI of an *old* version (v7r0) against master of DIRACOS. Is that really what we should be doing ?

chrisburr:
> The problem here is that we are running the CI of an old version (v7r0) against master of DIRACOS. Is that really what we should be doing ?

This ties in to https://github.com/DIRACGrid/DIRACOS2/issues/30. IMHO we should be either:
* keeping all active branches of DIRAC functional with the latest DIRACOS
* splitting DIRACOS in to incompatible versions and document the mapping of minor DIRAC releases to minor DIRACOS releases (this would likely need a patch number adding to DIRACOS)

At some point we'll have breaking changes and SQLAlchemy 1.3 -> 1.4 might be a big enough migration that we want to calling breaking. Equally it's a hassle so maybe we'd rather just backport the fixes?

chaen:
For this specific case, we could backport the fix, but I agree that a longer term solution should be considered. Let's continue the discussion on the issue you linked. In the meantime, do you want to backport the fixes or would you rather have me doing it ?

chrisburr:
Done in https://github.com/DIRACGrid/DIRAC/pull/5089

fstagni (approved):

PR #5091, opened 2021-04-06
[v7r2] Ensure BLOB columns are always passed bytes data

chrisburr:
BEGINRELEASENOTES

*RequestManagement
FIX: Ensure BLOB columns are always passed bytes data

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/RequestManagementSystem/Client/Operation.py
+++ b/src/DIRAC/RequestManagementSystem/Client/Operation.py
@@ -81,14 +81,15 @@ class Operation(object):
 
     self.TargetSE = None
     self.SourceSE = None
-    self.Arguments = None
+    self._Arguments = None
     self.Error = None
     self.Type = None
     self._Catalog = None
 
-    fromDict = fromDict if isinstance(fromDict, dict)\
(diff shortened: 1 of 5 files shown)


chaen:
Although this is stored as bytes, I believe it has always been used as test. So with this explicit change, we will end up with binary data percolating to places where we expect text. Either we should carefully scrutinize this (in extensions as well, so a note in the wiki), or we should make this blob a varchar or a text. Not sure yet which one is best

chrisburr:
As far as I know it's always used with `DEncode` which outputs bytes, e.g.

https://github.com/DIRACGrid/DIRAC/blob/ebf667824eb9bd3bf289018defa01d838a8ca4b7/src/DIRAC/TransformationSystem/Client/FileReport.py#L77-L78

chaen:
I guess the only problem could come from here https://github.com/DIRACGrid/DIRAC/blob/f20575f871aae83df969d494e4cbb97651c3e70f/src/DIRAC/DataManagementSystem/Agent/RequestOperations/StagingCallback.py#L37  

But this is not used yet, so I'd rather fix that later on

chaen (approved):

PR #5093, opened 2021-04-08
[v7r3] Remove pointless RNG seeding

chrisburr:


chaen:
Should it go earlier version ? Otherwise fine by me

chrisburr:
> Should it go earlier version ? Otherwise fine by me

I think it's fine in v7r3 as it should make no difference.

atsareg (approved):

PR #5097, opened 2021-04-14
Validate v7r2

atsareg:


chrisburr (approved):
Looks good, the failure is from a https://diracos.web.cern.ch/ timeout so nothing to worry about.

PR #5107, opened 2021-04-19
[v7r1] Freeze SQLALchemy to 1.3 until problems with 1.4 are solved

chaen:
Just freeze SQLAlchemy to 1.3 until we/they solve all the issues

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -33,7 +33,8 @@ dependencies:
   - recommonmark
   - requests >=2.9.1
   - six >=1.10
-  - sqlalchemy >=1.0.9
+  # Freeze until all problems with 1.4 are solved
+  - sqlalchemy ==1.3.*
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -33,7 +33,8 @@ dependencies:
   - recommonmark
   - requests >=2.9.1
   - six >=1.10
-  - sqlalchemy >=1.0.9
+  # Freeze until all problems with 1.4 are solved
+  - sqlalchemy >=1.3.*
```
chrisburr, line 37: I guess you mean?
```suggestion
  - sqlalchemy 1.3.*
```

chaen: that should work better, thanks

chrisburr (approved):

fstagni (approved):

PR #5143, opened 2021-05-11
[v7r1] Clarify tarfile error message in pilot logs

chrisburr:
This message is a constant source of confusion for people in the pilot logs so I propose we put at least something to help people realise this isn't actually the problem. I've gone for the concise option but an alternative would be to actually check if there are locations remaining in the list.

BEGINRELEASENOTES

*WorkloadManagement
CHANGE: Clarify that "tarfile failed with message" can be normal

ENDRELEASENOTES


VladimirRomanovsky:
The error messages from tar command and urllib few lines later  also are the source of confusion. In most cases they are coming together.

fstagni (approved):

PR #5153, opened 2021-05-20
[v7r1] Framework/Monitoring: increase perf at the cost of data safety

chaen:
This PR improves the performance of the monitoring service by an order of magnitude at the cost of running in a less safe mode. However, given that it's monitoring data, it's acceptable.
It also reduces the number of 'database is locked' exceptions from several thousands per hour to a handful.
This image shows the response time of just a `ping` call to the service before and after the change. The difference for actual method call is even bigger 
 
image

BEGINRELEASENOTES

*Framework
CHANGE: change SQLITE connection option for speed improvement 

ENDRELEASENOTES

### DIFF
--- a/FrameworkSystem/private/monitoring/MonitoringCatalog.py
+++ b/FrameworkSystem/private/monitoring/MonitoringCatalog.py
@@ -5,6 +5,7 @@ import six
 import sqlite3
 import os
 import hashlib
+import random
 import time
 
 import DIRAC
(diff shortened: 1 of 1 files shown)


chrisburr (dismissed):

fstagni (dismissed):

FrameworkSystem/private/monitoring/MonitoringCatalog.py
```diff
@@ -50,7 +58,7 @@ def __dbExecute(self, query, values=False):
         executed = True
       except Exception as e:
         self.log.exception("Exception executing statement", "query: %s, values: %s" % (query, values))
-        time.sleep(0.01)
+        time.sleep(random.random())
```
chrisburr, line 61: I think I can guess but maybe add a comment explaining why a random pause is being used?

atsareg (approved):

PR #5163, opened 2021-05-28
[v7r1] fix query syntax

arrabito:
BEGINRELEASENOTES

*DataManagement
FIX: Fix the syntax of the query to get dataset files

ENDRELEASENOTES


arrabito:
I see that some tests failed, but it does not seem to me related to the small change I've done in the code.
What should be done in this case?
Thank you.

atsareg (approved):

PR #5170, opened 2021-06-01
[v7r2] Pin astroid to avoid linter false positives

chrisburr:
I've opened issues upstream so hopefully these can be fixed rather than littering the code with disabled checks.

This is another case where `pre-commit` would be nice as it automatically pins versions of packages like `astroid`/`pylint` so we can bump the versions when it suits us rather than being randomly hit by regressions (or previously unknown bugs). Example: https://github.com/scikit-hep/uproot4/pull/373

### DIFF
--- a/environment-py3.yml
+++ b/environment-py3.yml
@@ -7,7 +7,7 @@ channels:
 
 dependencies:
   # Temporary workarounds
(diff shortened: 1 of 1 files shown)


fstagni (approved):

chrisburr:
@atsareg @fstagni Can we get this merged and propagated to get the CI green again?

PR #5175, opened 2021-06-02
[v7r2] Improve git describe command used by setuptools-scm

chrisburr:
There is an invalid tag in the history which cases the issue reported by @TaykYoku on slack when pip installing directly from a fork which pre-dates the new style versioning. This change makes the `git describe` command more robust (causing the version to default to something like `0.1.dev33206+g68d885b.d20210602` instead of erroring).

BEGINRELEASENOTES

*Python 3
FIX: Improve "git describe" command used by setuptools-scm to ignore v6e7p27

ENDRELEASENOTES


fstagni (approved):

TaykYoku:
Thanks @chrisburr, it fix my issue

PR #5180, opened 2021-06-07
[v7r2] Add callStack argument to S_ERROR to optimise CS

chrisburr:
When profiling some agents I've noticed that most of the time is spent generating the `CallStack` field in `S_ERROR`. This is due the CS being heavily used and even default keys have to format a full traceback. This PR adds a new `callStack` keyword argument to `S_ERROR` to avoid this which makes `gConfig.getOption` ~13x faster. Benchmarked with:

```python
import DIRAC
parseCommandLine()
DIRAC.gConfig.getOption("/Systems/Transformation/Production/Agents/WorkflowTaskAgent/PollingTim", 120)
%timeit DIRAC.gConfig.getOption("/Systems/Transformation/Production/Agents/WorkflowTaskAgent/PollingTim", 120)
```

Before:
```
137 µs ± 343 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

After:
```
11.3 µs ± 35.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

BEGINRELEASENOTES

*Core
NEW: Add callStack keyword argument to S_ERROR to allow object to be created much faster

*Configuration
CHANGE: No longer include a full CallStack in the S_ERROR object returned by ConfigurationClient.getOption

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/ConfigurationSystem/private/ConfigurationClient.py
+++ b/src/DIRAC/ConfigurationSystem/private/ConfigurationClient.py
@@ -145,7 +145,7 @@ class ConfigurationClient(object):
     if optionValue is None:
       return S_ERROR(
           "Path %s does not exist or it's not an option" % optionPath,
-          withStack=False,
+          callStack=["ConfigurationClient.getOption"],
       )
 
     # Value has been returned from the configuration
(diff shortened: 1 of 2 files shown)


fstagni (approved):

chaen:
I totally see the point, but how about instead of putting an empty list in the `callstack`, you just pass as argument `withStack=[ConfigurationClient.getOption]` ? This way it is very fast still, and yet we keep the information of the source at least ?

chrisburr:
I like that idea a lot (except it should be `["ConfigurationClient.getOption"]` to keep it serialisable).

I've renamed `withStack=True` to be `callStack=None` that can be an override containing a list of strings.

chaen:
that's what I meant yes ! Thanks a lot :-)

PR #5202, opened 2021-06-16
v7r2p11 notes and tags

atsareg:


atsareg:
The tests are failing due to conda environment and Rucio. I can only hope that this is irrelevant for the release

chrisburr:
Python 3 will fail to deploy.

PR #5222, opened 2021-06-23
[v7r3] py3 fixes

TaykYoku:
BEGINRELEASENOTES
PR to collect python3 DIRAC server fixes.

*Configuration
FIX: decode bytes to string before encoding it to JSON in TornadoService

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/ConfigurationSystem/Service/TornadoConfigurationHandler.py
+++ b/src/DIRAC/ConfigurationSystem/Service/TornadoConfigurationHandler.py
@@ -48,7 +48,7 @@ class TornadoConfigurationHandler(TornadoService):
       Returns the configuration
     """
     sData = self.ServiceInterface.getCompressedConfigurationData()
-    return S_OK(b64encode(sData).decode('utf-8'))
+    return S_OK(b64encode(sData).decode())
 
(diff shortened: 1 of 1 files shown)


src/DIRAC/ConfigurationSystem/Service/TornadoConfigurationHandler.py
```diff
@@ -48,7 +48,7 @@ def export_getCompressedData(self):
       Returns the configuration
     """
     sData = self.ServiceInterface.getCompressedConfigurationData()
-    return S_OK(b64encode(sData))
+    return S_OK(b64encode(sData).decode('utf-8'))
```
chaen, line 51: @chrisburr can confirm, but I was under the impression that we should just use `decode()`

chrisburr: Yup, it's the default in Python 3: https://docs.python.org/3/library/stdtypes.html#bytes.decode

(let's not think about Python 2 too much but `b64encode` will only ever return ASCII data so it doesn't matter)

chrisburr (approved):

fstagni (approved):

PR #5246, opened 2021-07-01
[v7r3] Python 3 hackathon fixes

TaykYoku:
BEGINRELEASENOTES
Python 3 hackathon fixes.

*Core
CHANGE: modify MySQL._to_string

*WorkloadManager
FIX: decode values using _to_string method in JobDB
FIX: mock _escapeString in Test_Agent_PilotStatusAgent to pass test
FIX: escape values in PilotAgentsDB
CHANGE: remove pilots before start Test_PilotsClient test

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/MySQL.py
+++ b/src/DIRAC/Core/Utilities/MySQL.py
@@ -942,15 +942,6 @@ class MySQL(object):
     """
     return str(param[0])
 
-  def _to_string(self, value):
-    """
-      Convert value to string
-    """
-    try:
-      return value.decode()
-    except (UnicodeDecodeError, AttributeError):
-      return str(value)
-
   def _getConnection(self, retries=MAXCONNECTRETRY):
     """ Return  a new connection to the DB,
 
--- a/src/DIRAC/Core/Utilities/MySQL.py
+++ b/src/DIRAC/Core/Utilities/MySQL.py
@@ -942,10 +942,14 @@ class MySQL(object):
     """
     return str(param[0])
 
-  def _to_string(self, param):
+  def _to_string(self, value):
     """
+      Convert value to string
     """
-    return param[0].tostring()
+    try:
+      return value.decode()
+    except (UnicodeDecodeError, AttributeError):
+      return str(value)
 
   def _getConnection(self, retries=MAXCONNECTRETRY):
     """ Return  a new connection to the DB,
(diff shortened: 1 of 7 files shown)


src/DIRAC/WorkloadManagementSystem/Service/WMSAdministratorHandler.py
```diff
@@ -217,7 +217,7 @@ def export_getJobPilotOutput(self, jobID):
             c = cycle
 
     if pilotReference:
-      return getGridJobOutput(pilotReference)
+      return getGridJobOutput(pilotReference.decode())
```
chrisburr, line 220: This should be fixed a level deeper in one or more of: `getAtticJobParameters`/`getJobParameter`/`getJobParameters`

chrisburr: @TaykYoku has debugged this further and got to the bottom of it on slack: https://lhcbdirac.slack.com/archives/C3AGWCA8J/p1625147097062700

In summary, the JobParameters table is using `BLOB` for the value when it should be using a text type.

TaykYoku:
restart tests

TaykYoku:
re-test

TaykYoku:
re-test

TaykYoku:
rebased and ready for review.

NOTE: PR fixes do not affect the database structure, IMHO it can be considered separately from issue.

fstagni (dismissed):

src/DIRAC/Core/Utilities/MySQL.py
```diff
@@ -942,10 +942,14 @@ def _to_value(self, param):
...
-  def _to_string(self, param):
+  def _to_string(self, value):
     """
+      Convert value to string
     """
-    return param[0].tostring()
+    try:
+      return value.decode()
+    except (UnicodeDecodeError, AttributeError):
+      return str(value)
```
chrisburr, line 952: If there is a `UnicodeDecodeError` then `return str(value)` is definitely the wrong thing to do as we'll get junk that starts to spread around. I'm also no sure when `AttributeError` would trigger (`value` is already a string maybe)?
```suggestion
    return value.decode()
```

chrisburr: Looking further down this PR: I think `_to_string(xxx)` can be deleted in favour of using `xxx.decode()`

TaykYoku: looking at the scope of this method, if only to convert a variable that is a string, then agree

atsareg (approved):

PR #5257, opened 2021-07-08
[v7r1] fix syntax query

arrabito:
*Accounting
FIX: GROUP BY statement syntax to be compatible with more recent SQL reference (e.g. as used in MariaDB 10.5). This fix is backward compatible.

### DIFF
--- a/AccountingSystem/DB/AccountingDB.py
+++ b/AccountingSystem/DB/AccountingDB.py
@@ -1145,7 +1145,13 @@ class AccountingDB(DB):
     if sqlLinkList:
       cmd += " AND %s" % " AND ".join(sqlLinkList)
     if groupFields:
-      cmd += " GROUP BY %s" % (groupFields[0] % tuple(groupFields[1]))
+      testGroupFields = " %s" % selectFields[0] % tuple(realFieldList)
+      testGroupFieldsList = testGroupFields.split(",")
+      realGroupFields = ()
+      for testGroupFields in testGroupFieldsList:
+        if "sum" not in testGroupFields.lower():
+          realGroupFields += (testGroupFields.strip(),)
+      cmd += " GROUP BY " + ','.join(realGroupFields)
     if orderFields:
       cmd += " ORDER BY %s" % (orderFields[0] % tuple(orderFields[1]))
     self.log.verbose(cmd)
--- a/AccountingSystem/DB/AccountingDB.py
+++ b/AccountingSystem/DB/AccountingDB.py
@@ -1145,13 +1145,16 @@ class AccountingDB(DB):
     if sqlLinkList:
       cmd += " AND %s" % " AND ".join(sqlLinkList)
     if groupFields:
-      testGroupFields = " %s" % selectFields[0] % tuple(realFieldList)
-      testGroupFieldsList = testGroupFields.split(",")
-      realGroupFields = ()
-      for testGroupFields in testGroupFieldsList:
-        if "sum" not in testGroupFields.lower():
-          realGroupFields += (testGroupFields.strip(),)
-      cmd += " GROUP BY " + ','.join(realGroupFields)
+      if len(groupFields[1]) == 1:
+        testGroupFields = " %s" % selectFields[0] % tuple(realFieldList)
+        testGroupFieldsList = testGroupFields.split(",")
+        realGroupFields = ()
+        for testGroupFields in testGroupFieldsList:
+          if "sum" not in testGroupFields.lower():
+            realGroupFields += (testGroupFields.strip(),)
+        cmd += " GROUP BY " + ','.join(realGroupFields)
+      else:
+        cmd += " GROUP BY %s" % (groupFields[0] % tuple(groupFields[1]))
     if orderFields:
       cmd += " ORDER BY %s" % (orderFields[0] % tuple(orderFields[1]))
     self.log.verbose(cmd)


arrabito:
This fix is needed because GROUP BY statements must contain all the non aggregated SELECT fields. 
This fix the kind of errors reported below when making Accounting plots:

```
SELECT `ac_key_CTA_Job_Site`.`value`,
SUM(`ac_bucket_CTA_Job`.`CPUTime`)/86400 FROM `ac_bucket_CTA_Job`,
`ac_key_CTA_Job_Site` WHERE `ac_bucket_CTA_Job`.`startTime` >=
1325116800 AND `ac_bucket_CTA_Job`.`startTime` <= 1624921200 AND
`ac_bucket_CTA_Job`.`Site` = `ac_key_CTA_Job_Site`.`id` GROUP BY
`ac_bucket_CTA_Job`.`Site`;
ERROR 1055 (42000): 'AccountingDB_cta.ac_key_CTA_Job_Site.value' isn't
in GROUP BY
```
This fix assumes that the only aggregation operation is SUM, but it should be extended if there are other aggregation operations. I've also checked that the fix is backward compatible with MySQL 5.7. It seems that this syntax was already the recommended one, but mysql was not doing strict checking.

Finally, there are might be more elegant ways to apply this fix.

AccountingSystem/DB/AccountingDB.py
```diff
@@ -1145,7 +1145,13 @@ def __queryType(
     if sqlLinkList:
       cmd += " AND %s" % " AND ".join(sqlLinkList)
     if groupFields:
-      cmd += " GROUP BY %s" % (groupFields[0] % tuple(groupFields[1]))
+      testGroupFields = " %s" % selectFields[0] % tuple(realFieldList)
```
fstagni, line 1148: This looks wrong.

atsareg:
I also think that the fix is not correct. It works if the grouping is by one field but not if by more than one field, e.g. channel in data operations which is defined by source and destination

arrabito:
I've checked that even for multiple group fields queries were working but it's true that for some queries, the group by statement was containing I guess unwanted operations, like 'CONCAT' in the example below:

`SELECT CONCAT( `ac_key_CTA_DataOperation_Source`.`value`, ' -> ', `ac_key_CTA_DataOperation_Destination`.`value` ), `ac_bucket_CTA_DataOperation`.`startTime`, `ac_bucket_CTA_DataOperation`.`bucketLength`, SUM(`ac_bucket_CTA_DataOperation`.`TransferSize`) FROM `ac_bucket_CTA_DataOperation`, `ac_key_CTA_DataOperation_Source`, `ac_key_CTA_DataOperation_Destination` WHERE `ac_bucket_CTA_DataOperation`.`startTime` >= 1627974000 AND `ac_bucket_CTA_DataOperation`.`startTime` <= 1628578800 AND `ac_bucket_CTA_DataOperation`.`Source` = `ac_key_CTA_DataOperation_Source`.`id` AND `ac_bucket_CTA_DataOperation`.`Destination` = `ac_key_CTA_DataOperation_Destination`.`id` GROUP BY CONCAT( `ac_key_CTA_DataOperation_Source`.`value`,' -> ',`ac_key_CTA_DataOperation_Destination`.`value` ),`ac_bucket_CTA_DataOperation`.`startTime`,`ac_bucket_CTA_DataOperation`.`bucketLength``

So, I've just added a condition to apply the change only when there is a single group field.

My tests indicate the fix works fine for different kind of queries (single or multiple group fields), but please cross-check.

fstagni:
We have not many tests for AccountingDB, so changes there make me slightly nervous. Also because I didn't reproduce this error. I added a test in https://github.com/DIRACGrid/DIRAC/pull/5321 but I tested it in v7r3. I can add your fix there too, but later.

fstagni:
Maybe I misunderstand, but I don't manage to reproduce this case in https://github.com/DIRACGrid/DIRAC/pull/5321/files and using MariaDB 10.6

fstagni:
Coming back to this. I have a few questions:
- do this issue pops up only when making plots for the Job category?
- have you tried with a later DIRAC version (v7r2)?
- have you tried by chance with MariaDB 10.6? (this is what is used for testing purposes in the DIRAC CI)

Finally: @marianne013 @sfayer you guys run MariaDB I think. Which version?

arrabito:
Hi,
Here my answers:
* It does not affect only Job Category. I haven't tested all possible combinations, but all "Pie" plots of the Job Category grouped by Site fail (while I tested one of them grouped by Country and it doesn't fail). 
Then, I get errors also for the Data Operation Category for the PieChartOfTransferredData_ByDestination but not for the other kinds of plots when grouped by destination. 
Finally I get errors for the WMS_History category for the AverageNbOfJobs grouped By Site (see screenshots in attachment).
* I get these errors on my server running v7r2p11
* I haven't tried MariaDB 10.6, only 10.5.9
WMS_History_AverageNbOfJobs_BySite
Pilot_TotalNumberOfPilots
Job_PiePlotOfWallTimeUsage_BySite
Job_PiePlotOfExecutedJobs_BySite
Job_PiePlotOfCPUUsed_BySite
DataOperation_PieChartOfTransferredData_ByDestination

fstagni:
I still don't manage to reproduce it, I fail to understand.
Anyway, I applied this to the DIRAC certification server and no issue arises. It's probably OK

fstagni:
Anyway, the docs failing is probably just because you need to rebase on the HEAD of v7r1.

PR #5265, opened 2021-07-12
[v7r2] Optimisation for Transformation system queries

chrisburr:
The three commits are worth viewing separately as they're three separate improvements:
* 48bdcf3: Add indices which make several queries considerably faster
* bba7880: When escaping a long list of values the database is pinged for every element by `MySQL._getConnection` making queries like `TransformationID is in (1,2,3,4,5,6,7,8,9,10)` unnecessarily slow from all the round trips to the database
* af2560a: Add `columns` argument to `TransformationClient.getTransformations` which is primarily useful for avoiding the download of hundred of megabytes of XML in the `Body` column when getting lots of transformations.

BEGINRELEASENOTES

*Core
FIX: Avoid pinging the database every time MySQL.__escapeString is called

*Transformation
NEW: Add columns keyword argument to TransformationClient.getTransformations
NEW: Add index on (TransformationID,Status) in TransformationFiles table
NEW: Add index on (TransformationID,ExternalStatus) in TransformationTasks table
NEW: Add index on (TransformationFamily) in Transformations table

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/MySQL.py
+++ b/src/DIRAC/Core/Utilities/MySQL.py
@@ -456,17 +456,17 @@ class MySQL(object):
     except Exception:
       return False
 
-  def __escapeString(self, myString):
+  def __escapeString(self, myString, connection=None):
     """
     To be used for escaping any MySQL string before passing it to the DB
     this should prevent passing non-MySQL accepted characters to the DB
     It also includes quotation marks " around the given string
     """
-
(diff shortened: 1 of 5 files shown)


fstagni (approved):

atsareg:
This will need a note in the releases WiKi about the DB schema changes

atsareg (approved):

PR #5297, opened 2021-07-29
 [integration] fix to allow options from Operations to be used by file catalog clients.

martynia:
BEGINRELEASENOTES

The `FileCatalog.py` obtains options both from Resources and Operations but they are not passed in to `FileCatalogfFctory` and consequently are not available to file catalog clients. This PR updates the options from Resources with the options from Operations and passes them to a the factory, so they are all available to a client.

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Catalog/FileCatalog.py
+++ b/src/DIRAC/Resources/Catalog/FileCatalog.py
@@ -427,7 +427,7 @@ class FileCatalog(object):
         return res
       catalogConfig = res['Value']
       if catalogConfig['Status'] == 'Active':
-        res = self._generateCatalogObject(catalogName)
+        res = self._generateCatalogObject(catalogName, **catalogConfig)
         if not res['OK']:
           return res
         oCatalog = res['Value']
@@ -473,10 +473,10 @@ class FileCatalog(object):
     catalogConfig['Master'] = (catalogConfig.setdefault('Master', False) == 'True')
     return S_OK(catalogConfig)
 
-  def _generateCatalogObject(self, catalogName):
+  def _generateCatalogObject(self, catalogName, **options):
     """ Create a file catalog object from its name and CS description
     """
     useProxy = gConfig.getValue('/LocalSite/Catalogs/%s/UseProxy' % catalogName, False)
     if not useProxy:
       useProxy = self.opHelper.getValue('/Services/Catalogs/%s/UseProxy' % catalogName, False)
-    return FileCatalogFactory().createCatalog(catalogName, useProxy)
+    return FileCatalogFactory().createCatalog(catalogName, useProxy, **options)
--- a/src/DIRAC/Resources/Catalog/FileCatalogFactory.py
+++ b/src/DIRAC/Resources/Catalog/FileCatalogFactory.py
@@ -21,7 +21,7 @@ class FileCatalogFactory(object):
     self.log = gLogger.getSubLogger('FileCatalogFactory')
     self.catalogPath = ''
 
-  def createCatalog(self, catalogName, useProxy=False):
+  def createCatalog(self, catalogName, useProxy=False, **options):
     """ Create a file catalog object from its name and CS description
     """
     catalogPath = getCatalogPath(catalogName)
@@ -31,6 +31,8 @@ class FileCatalogFactory(object):
     result = gConfig.getOptionsDict(catalogPath)
     if result['OK']:
       optionsDict = result['Value']
+
+    optionsDict.update(options)
 
     if useProxy:
       result = self.__getCatalogClass(catalogType)


src/DIRAC/Resources/Catalog/FileCatalog.py
```diff
@@ -427,7 +427,7 @@ def _getCatalogs(self):
         return res
       catalogConfig = res['Value']
       if catalogConfig['Status'] == 'Active':
-        res = self._generateCatalogObject(catalogName)
+        res = self._generateCatalogObject(catalogName, **catalogConfig)
```
andresailer, line 430: Why do you need the `**`? Why not pass catalogConfig as an optionsDict dictionary parameter?

martynia: By doing this, I don't have to test for the options being None in the line below:https://github.com/DIRACGrid/DIRAC/blob/6778bbef2a67f5d946bdb6cd1ab0261ec5034eb1/src/DIRAC/Resources/Catalog/FileCatalogFactory.py#L35

Besides I enforce it to be a dict.

andresailer: I think it would be better to be explicit that these functions now takes in addition a parameter that is a dictionary.

chaen:
I am a bit confused. If I understand correctly, you want to specify options in the code on top of what is in the CS ? If that is the case, that would be terrible practice

martynia:
> I am a bit confused. If I understand correctly, you want to specify options in the code on top of what is in the CS ? If that is the case, that would be terrible practice

No. I would like to use the options specified in Operations and to update those defined in Resources. Currently all VO specific options defined in Operations are ignored.

src/DIRAC/Resources/Catalog/FileCatalog.py
martynia: Optionally, only if there are any options in Operations. Adding an '{}` as a default parameter is out of question, so I find ** an elegant way.

chaen:
> > I am a bit confused. If I understand correctly, you want to specify options in the code on top of what is in the CS ? If that is the case, that would be terrible practice
> 
> No. I would like to use the options specified in Operations and to update those defined in Resources. Currently all VO specific options defined in Operations are ignored. See: `FileCatalogFactory`.

Wouldn't it be more logical that the `FileCatalogFactory` fetches the `Operations` content rater ?

martynia:
> > > I am a bit confused. If I understand correctly, you want to specify options in the code on top of what is in the CS ? If that is the case, that would be terrible practice
> > 
> > 
> > No. I would like to use the options specified in Operations and to update those defined in Resources. Currently all VO specific options defined in Operations are ignored. See: `FileCatalogFactory`.
> 
> Wouldn't it be more logical that the `FileCatalogFactory` fetches the `Operations` content rater ?

Yes, it could be. However when you look into `FileCatalog.py` those options are needed there (actually all options, both ones from Resources and ones from Operations (`_getCatalogConfigDetails` does this) So passing them all to the factory (and this is what is done in this PR) does not not require the Factory to fetch them at all.

chaen:
It's not quite correct. The `FileCatalog` needs a few of the information that are found in `Operations` to know how to use the catalog (e.g. `Status` or `AccessType`). 

But when it comes to creating the object, the `FileCatalogFactory` fetches the information from the CS itself : https://github.com/DIRACGrid/DIRAC/blob/b8e2abbe58dbf6a7b3fb730b43a41b010be11e3c/src/DIRAC/Resources/Catalog/FileCatalogFactory.py#L31

Although it may seems like a stupid detail, it actually is not. It is a design encouraged by the factory pattern. Btw, it also makes tests easier (the `FileCatalogFactory` would really deserve one...)

martynia:
> It's not quite correct. The `FileCatalog` needs a few of the information that are found in `Operations` to know how to use the catalog (e.g. `Status` or `AccessType`).
> 
> But when it comes to creating the object, the `FileCatalogFactory` fetches the information from the CS itself :
> 
> https://github.com/DIRACGrid/DIRAC/blob/b8e2abbe58dbf6a7b3fb730b43a41b010be11e3c/src/DIRAC/Resources/Catalog/FileCatalogFactory.py#L31
> 
> Although it may seems like a stupid detail, it actually is not. It is a design encouraged by the factory pattern. Btw, it also makes tests easier (the `FileCatalogFactory` would really deserve one...)

So are you suggesting to duplicate the code and fetch information from Operations from within the factory ?

chaen:
I do. 

Although I still fail at seeing the use case. Is it for MultiVO ?

martynia:
> I do.
> 
> Although I still fail at seeing the use case. Is it for MultiVO ?

Yes, for multiVO. Imagine you have a RucioFileCatalog defined in Resources and you specify more that one server, one for each VO in Operations.

chaen:
The original design would be that you define one catalog per VO in your `Resources` section and then have different `Operations/vo/Services/Catalogs/` section.

Of course, all this might change depending on the discussion here https://github.com/DIRACGrid/DIRAC/issues/5287

src/DIRAC/Resources/Catalog/FileCatalog.py
andresailer: It is elegant until someone adds another keyword argument, someone makes a typo somewhere and hilarity ensues because `**options`` swallows everything.

So then having `options=None` as a keyword argument, and adding two lines of checking for None seem like a small price to pay to me for having pylint save people from typos

martynia: OK, stylistic choice, I can make a change you are suggesting. Another issue is that @chaen questions the usefulness of this PR ...

martynia:
OK, I moved all the options to Resources, where the catalog is defined (more like server-by-server, not vo by vo) and referenced the catalog in Operations for gridpp for now. It should not be a problem to reference it for other VOs if they wish to use the same server.  It works now w/o the PR.

martynia:
Closed, as the same effect can be achieved by defining the options in Resources.

PR #5149, opened 2021-05-19
[v7r3] login with tokens (first part of the #5045)

TaykYoku:
This is the first part of the main release #5045 and attempt to implement the code with the least changes to the main project.
Here the opportunity to log in using identity providers and access to DIRAC services with the help of tokens is realized.

UPDATE:
- Added TokenManager service

TODO:
- [x]  update to use last authlib, pyjwt packages.
- [ ] add required packages to `DIRACOS2`
- [ ] approve [puppet commit](https://gitlab.cern.ch/ai/it-puppet-module-dirac/-/merge_requests/62)

# How to test it?

## Usual tests
 If you are testing without the use of tokens, for example to make sure that the usual behavior has not changed without the use of token authorization, then you do not need to do anything special. Everything must be backward compatible. Just do the tests as usual.

## Try tokens
- register new account in https://wlcg.cloud.cnaf.infn.it if you don't have it.
- joing to `wlcg` groups

### **DIRAC CLI test** 
- Install DIRAC Client with last DIRACOS2, e.g.:
```
curl -LO https://github.com/DIRACGrid/DIRACOS2/releases/latest/download/DIRACOS-Linux-x86_64.sh
bash DIRACOS-Linux-x86_64.sh
source ~/diracos/diracosrc
pip install git+https://github.com/TaykYoku/DIRAC.git@integration_oauth_p01
```
- configure client, this will create `~/diracos/etc/dirac.cfg` with required options _/DIRAC/Security/Authorization/issuer=https://lbcertifdiracoauth.cern.ch/auth_, _/DIRAC/Setup=DIRAC-CertifOauth_, _DIRAC/Configuration/Servers=https://lbcertifdiracoauth.cern.ch:9135/Configuration/Server_ (if you already have configured **dirac.cfg** you can add it manually):
```
dirac-configure --login=https://lbcertifdiracoauth.cern.ch/auth
```
Note If you are not yet registered in DIRAC, the first time you try to log in through DIRAC admins will receive a message asking to add you, after which you can try again.
- check installation and authorization status, when you configure script:
```
dirac-info
dirac-login --info
```
Note when you use login you get access token from Identity provide with scopes that described for DIRAC group. If you whant to get proxy as usual use `--proxy` key.
- try to access to https service:
```
dirac-dms-filecatalog-cli
```
- Most services work through DISET so you can try to get proxy instead of token (proxy must be uploaded into `ProxyManager` as usual):
```
dirac-login -g dirac_user --proxy
```
- check proxy:
```
[...]

PR #5333, opened 2021-08-17
[v7r2] fix doc setInputMetaQuery, better grep for :param

andresailer:
@TaykYoku FYI for the param grep


fstagni (approved):

TaykYoku:
@andresailer thanks! this fix will be useful for #5149

PR #5341, opened 2021-08-19
[v7r3] Fix running tornado-start-all without a proxy

chrisburr:
@fstagni you added this [here](https://github.com/DIRACGrid/DIRAC/pull/5214/files#diff-8990c56b0602bc15b3b5b3525ca77bd304daa6d4a725f060fe00da6d90c2e3daR39-R40) but I don't understand why it would be needed? It's a problem as it causes the configuration system to fail to initialise as it's before the option to use server certificates is set.


fstagni (approved):
Indeed, it's not needed.

PR #5356, opened 2021-08-24
[v7r2] SSHBatchComputingElement fixes

atsareg:
BEGINRELEASENOTES

*Resources
FIX: executeBatch - fallback to urllib if six module is not available
FIX: Host.py - return Message in case of non-zero status in getCEStatus
FIX: SSHBatchComputingElement - added forgotten instantiation of the BatchSystem plugin

*WMS
CHANGE: SiteDirector - continue queues initialization if one of them fails 

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/BatchSystems/Host.py
+++ b/src/DIRAC/Resources/Computing/BatchSystems/Host.py
@@ -194,9 +194,9 @@ exit 0
       infoFile.close()
       jobInfo = json.loads(jobInfo)
       pid = jobInfo['PID']
-      sp = subprocess.Popen( ["ps", "-f", "-p", str( pid ), "--no-headers"],
-                             stdout = subprocess.PIPE,
-                             stderr = subprocess.PIPE )
+      sp = subprocess.Popen(["ps", "-f", "-p", str(pid), "--no-headers"],
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE)
       output, error = sp.communicate()
       if len(output.split('\n')) == 2 and 'wrapper' in output:
         running += 1
@@ -217,9 +217,9 @@ exit 0
 
     if pid == 0:
       return "Unknown"
-    sp = subprocess.Popen(["ps", "-f", "-p", str( pid ), "--no-headers"],
-                           stdout = subprocess.PIPE,
-                           stderr = subprocess.PIPE)
+    sp = subprocess.Popen(["ps", "-f", "-p", str(pid), "--no-headers"],
+                          stdout=subprocess.PIPE,
+                          stderr=subprocess.PIPE)
     output, error = sp.communicate()
     status = sp.returncode
     if status == 0 and len(output.split('\n')) == 2 and user in output:
--- a/src/DIRAC/Resources/Computing/BatchSystems/Host.py
+++ b/src/DIRAC/Resources/Computing/BatchSystems/Host.py
@@ -223,7 +223,7 @@ exit 0
 
     if pid == 0:
       return "Unknown"
-    cmd = 'ps  -f -p %s | grep %s | wc -l' % (pid, user)
+    cmd = 'bash -c "ps -f -p %s | grep %s | wc -l"' % (pid, user)
     sp = subprocess.Popen(shlex.split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     output, error = sp.communicate()
     status = sp.returncode
--- a/src/DIRAC/Resources/Computing/BatchSystems/Host.py
+++ b/src/DIRAC/Resources/Computing/BatchSystems/Host.py
@@ -195,7 +195,7 @@ exit 0
       infoFile.close()
       jobInfo = json.loads(jobInfo)
       pid = jobInfo['PID']
-      cmd = 'ps -f -p %s --no-headers | wc -l' % pid
+      cmd = 'bash -c "ps -f -p %s --no-headers | wc -l"' % pid
       sp = subprocess.Popen(shlex.split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
       output, error = sp.communicate()
       status = sp.returncode
--- a/src/DIRAC/Resources/Computing/BatchSystems/Host.py
+++ b/src/DIRAC/Resources/Computing/BatchSystems/Host.py
@@ -15,7 +15,6 @@ from __future__ import division
 import os
 import glob
 import shutil
-import shlex
 import signal
 try:
   import subprocess32 as subprocess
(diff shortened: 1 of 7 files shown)


src/DIRAC/Resources/Computing/BatchSystems/Host.py
```diff
@@ -195,7 +195,7 @@ def getCEStatus(self, **kwargs):
       infoFile.close()
       jobInfo = json.loads(jobInfo)
       pid = jobInfo['PID']
-      cmd = 'ps -f -p %s --no-headers | wc -l' % pid
+      cmd = 'bash -c "ps -f -p %s --no-headers | wc -l"' % pid
```
chrisburr, line 198: If you want to count the lines use `len()` in Python instead of launching two extra processes and introducing a possible arbitrary code execution vulnerability.

Or just use `psutil` that will be faster and more robust.

atsareg: This code is executed by native python on a node without any special DIRAC environment. So, psutil apriori is not available.

atsareg: I tried to consider your suggestion and this looks ugly. "ps" gives non-zero status code if the process is not found but also it can be for some other reasons, so analyzing the returncode is not relevant then. Filtering the output through "wc -l" does not seem to be heavy, this was working for ages.

chrisburr: Sorry I had read to quickly when checking.

Looking more closely this is still broken as adding the `bash -c` just makes the problem even worse as `shlex.split(cmd)`. I think you're looking for the `shell=True` argument to `Popen` but we shouldn't use that.

1. Am I right in thinking that checking the status code is a waste of time as it checks the return code of `wc -l` which will only fail if `wc` isn't installed on the host?

2. If so, how about this instead:

```python
sp = subprocess.Popen(["ps", "-f", "-p", str(pid), "--no-headers"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
output, error = sp.communicate()
if output:
  running += 1
  usedCores += jobInfo['NCores']
else:
  stamp = jobInfo['JOBID']
  jobLife = datetime.utcnow() - datetime.strptime(jobInfo['SubmissionTime'], "%Y-%m-%d %H:%M:%S")
  if jobLife > CLEAN_DELAY:
    self.__cleanJob(stamp, infoDir, workDir, outputDir, errorDir)
```

atsareg: The fix I did I tested in a production environment where I have suffered from this problem. Your suggestion can work also, I can try. But "ps" can give some errors on some nodes which in your case will be interpreted as detecting a running process (non empty output). So, I would prefer still to filter it through wc -l and check that the output is '1'.
The returncode checking does not make much sense indeed but it is always good to do. This is how I discovered that Popen without shell does not work.

src/DIRAC/Resources/Computing/SSHBatchComputingElement.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, ceUniqueID):
 
   def _reset(self):
 
+    batchSystemName = self.ceParameters.get('BatchSystem', 'Host')
+    if 'BatchSystem' not in self.ceParameters:
+      self.ceParameters['BatchSystem'] = batchSystemName
+    result = self.loadBatchSystem(batchSystemName)
+    if not result['OK']:
+      self.log.error('Failed to load the batch system plugin %s', batchSystemName)
```
andresailer, line 47:
```suggestion
      self.log.error('Failed to load the batch system plugin', batchSystemName)
```

src/DIRAC/Resources/Computing/BatchSystems/executeBatch.py
```diff
@@ -18,8 +18,12 @@
 
   import sys
   import json
-  from six.moves.urllib.parse import quote as urlquote
-  from six.moves.urllib.parse import unquote as urlunquote
+  try:
+    from six.moves.urllib.parse import quote as urlquote
+    from six.moves.urllib.parse import unquote as urlunquote
+  except ImportError:
+    from urllib import unquote as urlunquote
```
andresailer, line 25: What about sites where python3 is the available python?

src/DIRAC/Resources/Computing/BatchSystems/Host.py
atsareg: shlex is splitting the line as:

    $  shlex.split('bash -c "ps -f -p 111 --no-headers | wc -l"')
    ['bash', '-c', 'ps -f -p 111 --no-headers | wc -l']

And apparently it works with Popen.

atsareg: OK, did it without shell.

atsareg:
The fixes are tested in the production EGI installation as hot fix. Please, rereview

src/DIRAC/Resources/Computing/BatchSystems/Host.py
```diff
@@ -195,22 +194,18 @@ def getCEStatus(self, **kwargs):
...
-      if status == 0:
-        if output.strip() == '1':
-          running += 1
-          usedCores += jobInfo['NCores']
-        else:
-          stamp = jobInfo['JOBID']
-          jobLife = datetime.utcnow() - datetime.strptime(jobInfo['SubmissionTime'], "%Y-%m-%d %H:%M:%S")
-          if jobLife > CLEAN_DELAY:
-            self.__cleanJob(stamp, infoDir, workDir, outputDir, errorDir)
+      if len(output.split('\n')) == 2 and 'wrapper' in output:
```
chrisburr, line 201: I'm not sure but is `wrapper` guaranteed to be in the output? Would it be better to just check:
```suggestion
      if output.strip():
```

src/DIRAC/Resources/Computing/BatchSystems/Host.py
```diff
@@ -222,11 +217,12 @@ def __checkPid(self, pid, user):
...
       return "Unknown"
-    cmd = 'ps  -f -p %s | grep %s | wc -l' % (pid, user)
-    sp = subprocess.Popen(shlex.split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    sp = subprocess.Popen(["ps", "-f", "-p", str(pid), "--no-headers"],
+                          stdout=subprocess.PIPE,
+                          stderr=subprocess.PIPE)
     output, error = sp.communicate()
     status = sp.returncode
-    if status == 0 and output.strip() == "1":
+    if status == 0 and len(output.split('\n')) == 2 and user in output:
```
chrisburr, line 225: I don't think the length check really adds anything?
```suggestion
    if status == 0 and user in output:
```

src/DIRAC/Resources/Computing/BatchSystems/Host.py
atsareg: wrapper is guaranteed to be there by construction. I prefer len check as it is also guaranteed in success case and redundant check can help against some ps failures that I can not imagine right now

src/DIRAC/Resources/Computing/BatchSystems/Host.py
atsareg: Same comment as above

chrisburr (approved):

PR #5368, opened 2021-08-27
[v7r1] fix show jobs

TaykYoku:
BEGINRELEASENOTES

Fix JobMonitoring MySQL error:

```
WorkloadManagement/JobMonitoring (88.84 ms) ERROR: MySQL Error ( 1131 : Execution failed.: ( 1054: Unknown column 'PilotJobReference' in 'where clause' ))
```

*WorkloadManagement
FIX: Unknown column 'PilotJobReference'

ENDRELEASENOTES

### DIFF
--- a/WorkloadManagementSystem/Service/JobMonitoringHandler.py
+++ b/WorkloadManagementSystem/Service/JobMonitoringHandler.py
@@ -44,6 +44,8 @@ class JobMonitoringHandler(RequestHandler):
         '/Services/JobMonitoring/useESForJobParametersFlag', False)
     if useESForJobParametersFlag:
       cls.gElasticJobParametersDB = ElasticJobParametersDB()
+    
+    cls.pilotManager = PilotManagerClient()
     return S_OK()
 
   def initialize(self):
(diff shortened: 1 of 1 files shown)


WorkloadManagementSystem/Service/JobMonitoringHandler.py
```diff
@@ -63,6 +66,44 @@ def initialize(self):
...
+    # If that occurs, use the JobID instead of the PilotJobReference
+    pilotJobRefs = selectDict.get('PilotJobReference')
+    if pilotJobRefs:
+      del selectDict['PilotJobReference']
+      if not selectDict.get('JobID'):
+        for pilotJobRef in [pilotJobRefs] if isinstance(pilotJobRefs, six.string_types) else pilotJobRefs:
+          res = cls.pilotManager.getPilotInfo(pilotJobRef)
+          if res['OK'] and 'Jobs' in res['Value'][pilotJobRef]:
+            selectDict['JobID'] = selectDict.get('JobID', [])
+            selectDict['JobID'].extend(res['Value'][pilotJobRef]['Jobs'])
```
aldbr, line 103: I guess you don't need to execute this line for each `pilotJobRef` (line 102): 

```
selectDict['JobID'] = selectDict.get('JobID', [])
```
Because:
- If you are here, then it means there is no `JobID` in `selectDict`: (see line 98)
- Once you have it, it is sufficient to extend it, there is no need to get it multiple time.


```suggestion
    if pilotJobRefs:
      del selectDict['PilotJobReference']
      if not selectDict.get('JobID'):
        selectDict['JobID'] = []
        for pilotJobRef in [pilotJobRefs] if isinstance(pilotJobRefs, six.string_types) else pilotJobRefs:
          res = cls.pilotManager.getPilotInfo(pilotJobRef)
          if res['OK'] and 'Jobs' in res['Value'][pilotJobRef]:
            selectDict['JobID'].extend(res['Value'][pilotJobRef]['Jobs'])
```

aldbr (commented):
The `parseSelectors` method is a good idea.

There is probably an issue but I doubt it is bound to `showJobs`, `PilotJobReference` was already removed from `selectDict` and I don't manage to reproduce the error with `v7r2` for instance (when I click on `showJobs` in the webapp). How can I reproduce it?

Thanks

TaykYoku:
this error occurred on the last hackathon:
PilotManager -> show jobs -> this opens the JobMonitor tab with PilotJobReference filter -> Submit

WorkloadManagementSystem/Service/JobMonitoringHandler.py
TaykYoku: If nothing happens below the 99 line, a `JobID in (  )` query is generated, which also raise a MySQL error.

aldbr: Unless I miss something, `pilotJobRefs` cannot be `None` or empty at this point (see line 96). Am I wrong?

TaykYoku: It may be, why not?

TaykYoku: and below the 96th line `pilotJobRefs` can no longer be empty. it's true.

atsareg (approved):

PR #5371, opened 2021-08-30
[v7r2] Multi-node allocations: Embed executableFile in the srun Wrapper

aldbr:
So far, I didn't manage to make the `Srun` parallel library work with the `SSHCE` because it is hard to test and the mechanism is complex: `Srun` generates a new executable but the original one has to be submitted along with the new.
- Both executables have to be placed in a specific remote directory and the new one has to know the location of the original one. 
- When we use `LocalCE` with `Srun`, the original executable must not be deleted before the end of the execution, which is cumbersome.

This PR aims to solve this complexity: the original executable is embedded into the new one.
Thus, there is only one file to submit instead of 2, which:
- implies less potential issues with the location of the files
- allows to remove the original executable just after the submission when `Srun` is used in combination of `LocalCE`.

BEGINRELEASENOTES
*Resources
CHANGE: ParallelLibrary: Embed the executable in the srun wrapper
ENDRELEASENOTES


atsareg:
I was trying to understand the logic of the parallelLibrary that you have introduced. The only implementation that you provide is SLURM specific. I wonder, why this can not be fully implemented inside the SLURM.py BatchSystems class ? I can imagine that it can greatly simplify the things, especially the SSHCE

fstagni:
The reason why only SLURM is used is because as of now SLURM is the de-facto standard for HPCs. The reasoning behind the current complexity is to possibly use other batch systems, but as now it's not needed.

Code-wise this looks OK to me.

fstagni (approved):

atsareg:
I think that we have agreed with Alexandre that this should be implemented in the Slurm.py. Especially is that it is a de facto standard and little chance that elsewhere it can be applicable. This implementation here is overly "object-oriented" to me.

atsareg:
How approval of this PR can be redrawn ?

andresailer:
what do you mean with "redrawn"? withdrawn? disapproved? Changes requested?

chrisburr:
I suspect you want to leave a review and use this button:
Screenshot 2021-10-06 at 11 16 17

fstagni:
I dismissed my PR approval.

aldbr:
> I think that we have agreed with Alexandre that this should be implemented in the Slurm.py. Especially is that it is a de facto standard and little chance that elsewhere it can be applicable. This implementation here is overly "object-oriented" to me.

Well I am not completely convinced actually. `srun` is only working for SLURM, this is right but:
- There is another example of parallel library that I have barely tested, which seems available on supercomputers we use: `mpirun`. According to [this presentation](https://www.hlrn.de/doc/download/attachments/9667587/HLRN%20Anwenderschulung%202020%20-%20srunvsmpirun.pdf?version=1&modificationDate=1604515334919&api=v2), it is "a wrapper script that mimics srun". It is able to work with various batch systems such as Torque, SGE, Slurm (see [this link](https://software.intel.com/content/www/us/en/develop/documentation/mpi-developer-reference-linux/top/command-reference/mpirun.html))
- I have not tested that yet, but in the context of the supercomputers with no external connectivity, if I want to run multiple jobs in parallel (to avoid wasting many-core nodes with only 1 job) by passing them in a CE interacting with Slurm, the current approach allows me to do it.

What do you mean exactly by "overly object-oriented"?

atsareg:
By "overly object-oriented" I mean that a relatively small and self-contained feature is generalized on the level which is not on the right scale, too large, affecting objects for which it will be an unnecessary heaviness.

atsareg:
Redrawn, withdrawn, ... So, be careful while interpreting my english statements :) Thanks for the hints !

atsareg (approved):

PR #5373, opened 2021-08-31
[v7r2] Fix un-returned use of S_ERROR and S_OK

chrisburr:
#5367 made me wonder if there were any other cases of `S_ERROR`/`S_OK` being created without actually being returned and it turns out the answer is yes!

### DIFF
--- a/src/DIRAC/Core/Security/VOMS.py
+++ b/src/DIRAC/Core/Security/VOMS.py
@@ -123,9 +123,8 @@ class VOMS(BaseSecurity):
     :return:  status, output, error, pyerror.
     """
     validOptions = ['actimeleft', 'timeleft', 'identity', 'fqan', 'all']
-    if option:
-      if option not in validOptions:
-        S_ERROR(DErrno.EVOMS, "valid option %s" % option)
+    if option and option not in validOptions:
+      return S_ERROR(DErrno.EVOMS, "valid option %s" % option)
 
     retVal = multiProxyArgument(proxy)
     if not retVal['OK']:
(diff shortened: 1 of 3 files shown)


src/DIRAC/Core/Security/VOMS.py
```diff
@@ -123,9 +123,8 @@ def getVOMSProxyInfo(self, proxy, option=False):
     :return:  status, output, error, pyerror.
     """
     validOptions = ['actimeleft', 'timeleft', 'identity', 'fqan', 'all']
-    if option:
-      if option not in validOptions:
-        S_ERROR(DErrno.EVOMS, "valid option %s" % option)
+    if option and option not in validOptions:
+      return S_ERROR(DErrno.EVOMS, "valid option %s" % option)
```
andresailer, line 127:
```suggestion
      return S_ERROR(DErrno.EVOMS, "invalid option %s" % option)
```

src/DIRAC/DataManagementSystem/Client/ConsistencyInspector.py
```diff
@@ -498,8 +498,7 @@ def set_prod(self, value):
       value = int(value)
       res = self.transClient.getTransformation(value, extraParams=False)
       if not res['OK']:
-        S_ERROR(errno.ENOENT, "Couldn't find transformation %d: %s" %
-                (value, res['Message']))
+        raise Exception("Couldn't find transformation %d: %s" % (value, res['Message']))
```
atsareg, line 501: Shouldn't the S_ERROR be just returned ?

chrisburr: It can't be as it's used in a `property`'s setter so there is no way to access the return value.

PR #5391, opened 2021-09-07
[8.0] Remove LcgFileCatalogClient

chrisburr:
The `lfcthr` Python module isn't included in DIRACOS2 but I don't think anyone is still using it so I guess this code can be simply removed?

BEGINRELEASENOTES

*DataManagement
CHANGE: The LcgFileCatalogClient has been removed

ENDRELEASENOTES


chaen:
Green light on my side, but let's just make sure with Belle II (@hmiyake @iueda )

iueda:
I think it is ok for Belle II, but just double checking... for confirmation.
(Even if we find something, we should solve it by the time we move to 8.0...)

iueda:
Talking with the Belle II colleagues, we all think it is ok for us.

fstagni (approved):

chaen (approved):

atsareg:
What form should be the version in the __init__.py now ?

chrisburr:
> What form should be the version in the init.py now ?

It's not needed at all as it comes from the tag, see #5417

PR #5403, opened 2021-09-15
[v7r2] Add semantic tests for commits

chaen:
Because I am more than fed up to browse a history with 376 consecutive "typo" or "fix" commit messages for a single file, I propose this extra action check which enforces a specific format (inspired by https://www.conventionalcommits.org/en/v1.0.0/#summary)

Details in the PR directly 

It's nothing really new, we already do the same effort to write it in a sort of similar way for the release news. 
That's an extra step towards a better life together.

BEGINRELEASENOTES
*Test
NEW: force commit format

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/semantic.yml
+++ b/.github/workflows/semantic.yml
@@ -1,0 +1,33 @@
+# Check that the commit messages are between 20 and 150 chars,
+# and that they follow the rule
+# <type> (<scope>): <subject>
+# type: docs, feat, fix, refactor, style or test
+# scope (optional): any extra info, (like DMS or whatever)
+
+name: 'Commit Message Check'
+on: pull_request
+
+jobs:
+  check-commit-message:
+    name: Check Commit Message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Commit Format
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^(docs|feat|fix|refactor|style|test)( ?\(.*\))?: .+$'
+          excludeDescription: 'true' # optional: this excludes the description body of a pull request
+          excludeTitle: 'true' # optional: this excludes the title of a pull request
+          checkAllCommitMessages: 'true' # optional: this checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }} # github access token is only required if checkAllCommitMessages is true
+          flags: 'gm'
+          error: 'Your commit has to follow the format "<type>(<scope>): <subject>"".'
+      - name: Check Commit Length
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.{20,150}$'
+          error: 'Commit messages should be between 20 and 150 chars'
+          excludeDescription: 'true' # optional: this excludes the description body of a pull request
+          excludeTitle: 'true' # optional: this excludes the title of a pull request
+          checkAllCommitMessages: 'true' # optional: this checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }} # github access token is only required if checkAllCommitMessages is true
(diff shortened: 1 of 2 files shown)


atsareg:
I agree in general but this will make current PRs all invalid - hardly acceptable.

.github/workflows/semantic.yml
```diff
@@ -0,0 +1,33 @@
...
+          excludeDescription: 'true' # optional: this excludes the description body of a pull request
+          excludeTitle: 'true' # optional: this excludes the title of a pull request
+          checkAllCommitMessages: 'true' # optional: this checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }} # github access token is only required if checkAllCommitMessages is true
+          flags: 'gm'
+          error: 'Your commit has to follow the format "<type>(<scope>): <subject>"".'
+      - name: Check Commit Length
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.{20,150}$'
```
andresailer, line 28: This only limits the length of the first line of the commit, or also the total length of the commit message?

I  occasional longer commit message, and many times multiple lines

.github/workflows/semantic.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Check Commit Format
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^(docs|feat|fix|refactor|style|test)( ?\(.*\))?: .+$'
+          excludeDescription: 'true' # optional: this excludes the description body of a pull request
+          excludeTitle: 'true' # optional: this excludes the title of a pull request
+          checkAllCommitMessages: 'true' # optional: this checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }} # github access token is only required if checkAllCommitMessages is true
+          flags: 'gm'
+          error: 'Your commit has to follow the format "<type>(<scope>): <subject>"".'
```
TaykYoku, line 24: Since the type value is strictly limited, it may be worth noting the list of possible values here? For example:
`"<type (docs, feat, fix, refactor, style, test)>(<scope>): <subject>"`

Or check the type separately and report that it does not match the possible values.

chaen:
> I agree in general but this will make current PRs all invalid - hardly acceptable.

* the existing PR will not be affected, as the CI already ran
* hardly acceptable: very acceptable IMO when balanced with the advantages brought by this PR. And `git` offers many options to edit commit messages.

.github/workflows/semantic.yml
chaen: I just tested it (https://github.com/chaen/DIRAC/pull/20/checks?check_run_id=3610068989), and it limits the first line, as of now. To be honest, long commits are very rare. And it's more the `fix`, `typo`, `test` commits that I want to avoid....

And looking at your commit, the longest I could find was 155 char 

```
commit fc08962f39d9abfcd61c0bfb90cc8e61c6076127
Author: Andre Sailer <andre.philippe.sailer@cern.ch>

    FileCatalogDirectoryTreeBase: prevent RuntimeError (maximum recursion depth exceeded) in FileCatalog service when illegal path is given to some functions.
```

.github/workflows/semantic.yml
chaen: I can add it in the doc. Making a different job for each value is a waste of resources

.github/workflows/semantic.yml
andresailer: If it only limits the first line, this is OK.

I was thinking of commits like this 99da524399ff62f6611df97c9775ce1973cda2b8

fstagni (approved):
We did not have time to discuss this in yesterday's BiLD meeting, but I think it's fine.

atsareg:
"Merge pull request" commit fails this tests, this should be taken into account

PR #5408, opened 2021-09-16
[v7r2] making use of freesize in WLCGAccountingJson if exists

iueda:
BEGINRELEASENOTES

*Resources
FIX: making use of freesize in WLCGAccountingJson if exists, instead of calculating as Total - usedsize

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Storage/OccupancyPlugins/WLCGAccountingJson.py
+++ b/src/DIRAC/Resources/Storage/OccupancyPlugins/WLCGAccountingJson.py
@@ -107,7 +107,9 @@ class WLCGAccountingJson(object):
     sTokenDict['SpaceReservation'] = spaceReservation
     try:
       sTokenDict['Total'] = storageSharesSR['totalsize']
-      sTokenDict['Free'] = sTokenDict['Total'] - storageSharesSR['usedsize']
+      sTokenDict['Free'] = storageSharesSR.get('freesize')
+      if not sTokenDict['Free']:
+        sTokenDict['Free'] = sTokenDict['Total'] - storageSharesSR['usedsize']
     except KeyError as e:
       return S_ERROR(
           errno.ENOMSG, 'Issue finding Total or Free space left. %s in %s storageshares.' % (repr(e), spaceReservation))
--- a/src/DIRAC/Resources/Storage/OccupancyPlugins/WLCGAccountingJson.py
+++ b/src/DIRAC/Resources/Storage/OccupancyPlugins/WLCGAccountingJson.py
@@ -107,9 +107,7 @@ class WLCGAccountingJson(object):
     sTokenDict['SpaceReservation'] = spaceReservation
     try:
       sTokenDict['Total'] = storageSharesSR['totalsize']
-      sTokenDict['Free'] = storageSharesSR.get('freesize')
-      if not sTokenDict['Free']:
-        sTokenDict['Free'] = sTokenDict['Total'] - storageSharesSR['usedsize']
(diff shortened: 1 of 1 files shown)


src/DIRAC/Resources/Storage/OccupancyPlugins/WLCGAccountingJson.py
```diff
@@ -107,7 +107,9 @@ def getOccupancy(self, **kwargs):
     sTokenDict['SpaceReservation'] = spaceReservation
     try:
       sTokenDict['Total'] = storageSharesSR['totalsize']
-      sTokenDict['Free'] = sTokenDict['Total'] - storageSharesSR['usedsize']
+      sTokenDict['Free'] = storageSharesSR.get('freesize')
+      if not sTokenDict['Free']:
+        sTokenDict['Free'] = sTokenDict['Total'] - storageSharesSR['usedsize']
```
chaen, line 112:
```suggestion
      sTokenDict['Free'] = storageSharesSR.get('freesize', sTokenDict['Total'] - storageSharesSR['usedsize'])
```

chaen:
This is of course fine, but I am a bit surprised, since this WLCG task force explicitly did not want free space value. Do you have a site that publishes it ?

iueda:
> This is of course fine, but I am a bit surprised, since this WLCG task force explicitly did not want free space value. Do you have a site that publishes it ?

Yes, here is a "customized" version of WLCGAccounting json, where totalsize - usedsize is not useful and freesize is provided as the actual value for the free space:  
```
gfal-cat srm://dcache-se-desy.desy.de:8443/srm/managerv2?SFN=/pnfs/desy.de/belle/belle2/TMP/occupancy.json
{
    "storageservice":
    {
        "storageshares":
        [
            {
                "name": "BELLE2DATA",
                "path":
                [
                    "/pnfs/desy.de/belle/belle2/DATA/"
                ],
                "numberoffiles": 370532,
                "totalsize": 628926799383072,
                "usedsize": 179148132275499,
                "freesize": 16166144760942,
                "timestamp": 1614168985
            }
        ]
    }
}
```

chaen:
but that seems like the accounting is wrong on this storage. Did you discuss the matter with them ? I am very curious to understand these strange values

iueda:
The accounting is not "wrong". We requested a standard accounting at the top level `/pnfs/desy.de/belle/belle2`, where it should be total = used + free.
Then, what they provided is a json per subdirectory, each corresponding to our SE endpoint, `DATA` and `TMP`, where
* totalsize = the total size of the SE, common to `DATA` and `TMP`
* freesize = the free size of the SE, common to `DATA` and `TMP`
* usedsize = the used size under each subdirectory, for they don't want to make "double-counting".
I told them that we are ok with double-counting, but they seem to believe their solution is the best.

Thus;
```
total = used(DATA) + used(TMP) + free
```
and we use only `freesize` for RSS for now, `usedsize` can be utilized in future when we compare the value against the accounting by the catalog.
There are subdirectories other than the above, with a separate disk pool for each, with different totalsize. Therefore, I guess they prefer to put a json per subdirectory.

chaen:
I am happy merging this PR, as it does not do any harm, so I will approve it.

However, this json accounting clearly does not match the standard agreed upon by all the storage providers (https://eoslhcb.cern.ch/eos/lhcb/proc/accounting). 
In the case of different pool, each pool should be an entry in the global json file. When you say 

> they seem to believe their solution is the best.

is it DESY admins or dCache people ?

chaen (approved):

iueda:
> > is it DESY admins or dCache people ?

my understanding is that the DESY people implemented their own solution instead of activating what is implemented for dCache.

Issue #5114, opened 2021-04-21
Payload proxy renewal mechanism is screwed when using M2Crypto


atsareg:
The payload proxy renewal mechanism inside the pilot is screwed when using M2Crypto. The payload proxy is renewed in the thread coded in the [ComputingElement.py ](https://github.com/DIRACGrid/DIRAC/blob/54d89885089ba29205d50bee9c3bff068e1aed2d/src/DIRAC/Resources/Computing/ComputingElement.py#L388). However, the call to the ProxyManager in this case fails with RuntimeError "Proxy string no longer supported." from [here](https://github.com/DIRACGrid/DIRAC/blob/54d89885089ba29205d50bee9c3bff068e1aed2d/src/DIRAC/Core/DISET/private/Transports/SSL/M2Utils.py#L106). I guess that the problem is due to this [piece of code](https://github.com/DIRACGrid/DIRAC/blob/54d89885089ba29205d50bee9c3bff068e1aed2d/src/DIRAC/Core/DISET/private/BaseClient.py#L236). 

  What would be the right way to fix this ?

chaen:
Do you have the stack trace of the `ProxyManager` to pinpoint exactly where the call fails ?

chaen:
I think I found it, but I am not sure of the logic of this whole stuff. The fix can be either to pass that info differently (providing I understand what it actually is trying to do), or to implement the deprecated feature in the M2crypto implementation.

atsareg:
The exception is in the ProxyManager client in the pilot and I have difficulties to get the output right now (the jobs are in the ARC CEs with problems to download the pilot output). When I managed, I have seen the exception reported and the origin. I guess the problem is that the pilot proxy location is passed as a parameter to the service call explicitly, while the M2Crypto is only allowing the proxy discovery in standard locations. So, may be it is enough to just not to pass the pilot proxy location.

chaen:
Well in fact this is what should be done: pass the proxy location rather than passing it as a string. I'll take a quick look, but to validate it I would need you to come with a reproducer once you have the callstack

atsareg:
These are lines that reproduce the problem:

    from DIRAC.Core.Base import Script
    from DIRAC.Core.DISET.RPCClient import RPCClient
    from DIRAC.Core.Security.ProxyInfo import getProxyInfo
    
    Script.parseCommandLine()
    
    result = getProxyInfo()
    if not result['OK']:
      print "failed getProxyInfo()"
    
[...]

PR #5411, opened 2021-09-16
[v7r2] Reintroduce proxyChain parameter

chaen:
Addresses https://github.com/DIRACGrid/DIRAC/issues/5114

Running the following code

```python
from DIRAC.Core.Base.Client import Client
from DIRAC.Core.Security.X509Chain import X509Chain

Script.parseCommandLine()

#service = 'Framework/ProxyManager'
service = 'WorkloadManagement/JobMonitoring'
print("With standard proxy")
c = Client()
c.setServer(service)
res = c.whoami()
print( res['Value']['DN'] if res['OK'] else res)

print("With a stolen proxy")
proxy = X509Chain()
proxy.loadProxyFromFile('/tmp/chaen/proxy.fstagni.lhcb_user')
c = Client(proxyChain=proxy)
c.setServer(service)
res = c.whoami()
print( res['Value']['DN'] if res['OK'] else res)


```

Without the fix
```bash
With standard proxy
/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=chaen/CN=705305/CN=Christophe Haen
With a stolen proxy

  File "/tmp/chaen/CC7py3/10.3.0a3/diracos/lib/python3.9/site-packages/DIRAC/Core/DISET/private/BaseClient.py", line 469, in _connect
    transport = gProtocolDict[self.__URLTuple[0]]['transport'](self.__URLTuple[1:3], **self.kwargs)
  File "/tmp/chaen/CC7py3/10.3.0a3/diracos/lib/python3.9/site-packages/DIRAC/Core/DISET/private/Transports/M2SSLTransport.py", line 94, in __init__
    self.__ctx = getM2SSLContext(**kwargs)
  File "/tmp/chaen/CC7py3/10.3.0a3/diracos/lib/python3.9/site-packages/DIRAC/Core/DISET/private/Transports/SSL/M2Utils.py", line 111, in getM2SSLContext
    raise RuntimeError("Proxy string no longer supported.")
RuntimeError: Proxy string no longer supported.

```

With the fix
```bash
With standard proxy
/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=chaen/CN=705305/CN=Christophe Haen
With a stolen proxy
/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=fstagni/CN=693025/CN=Federico Stagni
```

I also fixed it for https

BEGINRELEASENOTES

*Core
FIX: RPCClient and TornadoClient support proxyChain 

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/DISET/private/Transports/SSLTransport.py
+++ b/src/DIRAC/Core/DISET/private/Transports/SSLTransport.py
@@ -54,7 +54,7 @@ def checkSanity(urlTuple, kwargs):
     certFile = certTuple[0]
     useCerts = True
   elif "proxyString" in kwargs:
-    if not isinstance(kwargs['proxyString'], six.string_types):
+    if not isinstance(kwargs['proxyString'], six.string_types if six.PY2 else bytes):
       gLogger.error("proxyString parameter is not a valid type", str(type(kwargs['proxyString'])))
       return S_ERROR("proxyString parameter is not a valid type")
   else:
--- a/src/DIRAC/Core/Tornado/Client/private/TornadoBaseClient.py
+++ b/src/DIRAC/Core/Tornado/Client/private/TornadoBaseClient.py
@@ -32,8 +32,10 @@ __RCSID__ = "$Id$"
 
 from io import open
 import errno
+import os
 import requests
 import six
+import tempfile
 from six.moves import http_client
 
 
@@ -505,6 +507,12 @@ class TornadoBaseClient(object):
     # Do we use the server certificate ?
     if self.kwargs[self.KW_USE_CERTIFICATES]:
       cert = Locations.getHostCertificateAndKeyLocation()
+    elif self.kwargs.get(self.KW_PROXY_STRING):
+      tmpHandle, cert = tempfile.mkstemp()
+      fp = os.fdopen(tmpHandle, "wb")
+      fp.write(self.kwargs[self.KW_PROXY_STRING])
+      fp.close()
+
     # CHRIS 04.02.21
     # TODO: add proxyLocation check ?
     else:
--- a/src/DIRAC/Core/DISET/private/Transports/SSL/M2Utils.py
+++ b/src/DIRAC/Core/DISET/private/Transports/SSL/M2Utils.py
@@ -7,6 +7,7 @@ from __future__ import division
 from __future__ import print_function
 
 import os
+import tempfile
 from M2Crypto import SSL, m2, X509
 
 from DIRAC.Core.DISET import DEFAULT_SSL_CIPHERS, DEFAULT_SSL_METHODS
@@ -72,8 +73,7 @@ def getM2SSLContext(ctx=None, **kwargs):
                                a proxy is used unless other flags are set.
         - useCertificates: Boolean, Set to true to use hostcerts in client
                            mode.
-        - proxyString: String, no-longer supported, used to allow a literal
-                               proxy string to be provided.
+        - proxyString: String, allow a literal proxy string to be provided.
         - proxyLocation: String, Path to file to use as proxy, defaults to
                                  usual location(s) if not set.
         - skipCACheck: Boolean, if True, don't verify peer certificates.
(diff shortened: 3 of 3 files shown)


src/DIRAC/Core/DISET/private/Transports/SSLTransport.py
```diff
@@ -54,7 +54,13 @@ def checkSanity(urlTuple, kwargs):
...
     useCerts = True
   elif "proxyString" in kwargs:
-    if not isinstance(kwargs['proxyString'], six.string_types):
+    if (
+        six.PY2 and not isinstance(
+            kwargs['proxyString'],
+            six.string_types)) or (
+        six.PY3 and not isinstance(
+            kwargs['proxyString'],
+            bytes)):
```
chrisburr, line 63: Please check this is the same but I think this can be simpler:
```suggestion
    if not isinstance(kwargs['proxyString'], six.string_types if six.PY2 else bytes):
```

src/DIRAC/Core/Tornado/Client/private/TornadoBaseClient.py
```diff
@@ -505,6 +507,12 @@ def _request(self, retry=0, outputFile=None, **kwargs):
     # Do we use the server certificate ?
     if self.kwargs[self.KW_USE_CERTIFICATES]:
       cert = Locations.getHostCertificateAndKeyLocation()
+    elif self.kwargs.get(self.KW_PROXY_STRING):
+      tmpHandle, cert = tempfile.mkstemp()
```
chrisburr, line 511: Does this ever get cleaned up again?

chrisburr: I think you can make `cert` a tuple to avoid writing anything to disk?

src/DIRAC/Core/DISET/private/Transports/SSLTransport.py
chaen: yes, better thanks !

src/DIRAC/Core/Tornado/Client/private/TornadoBaseClient.py
chaen: No it does not, and I don't really like it, but in fact we never clean them up... I don't see an esthetic way of cleaning it without ugly flags and `if`.

I did hope being able to pass certs as strings, but it is not supported unfortunately

https://github.com/psf/requests/blob/10968357a19a146332c8dfd8e6d17006d746e580/requests/adapters.py#L242-L254

chrisburr: I don't think the file needs to exist outside of this function's scope?

If so I think you can used `NamedTemporaryFile` and let it delete it as part of the garbage collection sometime after the function exits?

chaen: I thought of using `NamedTemporaryFile`, but
* There is that slightly bothering statement in the doc (https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) 
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).
* I still need to close the file eventually. Where do I close it ?

chrisburr: You don't need to close it, you can write then flush and the the GC take care of closing it. It's an ugly hack that doesn't work on Windows but it's better than leaving the files lying around and we're already very broken on the latter.

atsareg (approved):

src/DIRAC/Core/Tornado/Client/private/TornadoBaseClient.py
chaen: Damn you ! Your last comment lost me in `cpython` implementation and posix specification for a while :D 

> let Python delete it as part of the garbage collection sometime after the function exits

There is actually no guarantee that the file will ever be closed. The `inputoutput` tutorial used to say so up to 3.7, but was then changed to reflect the reality:

https://docs.python.org/3.7/tutorial/inputoutput.html#reading-and-writing-files
> If you’re not using the with keyword, then you should call f.close() to close the file and immediately free up any system resources used by it. If you don’t explicitly close a file, Python’s garbage collector will eventually destroy the object and close the open file for you, but the file may stay open for a while. Another risk is that different Python implementations will do this clean-up at different times.

https://docs.python.org/3.8/tutorial/inputoutput.html#reading-and-writing-files
> If you’re not using the with keyword, then you should call f.close() to close the file and immediately free up any system resources used by it.
Warning
Calling f.write() without using the with keyword or calling f.close() might result in the arguments of f.write() not being completely written to the disk, even if the program exits successfully. 

Full discussion about the matter:
https://bugs.python.org/issue17852
https://github.com/python/cpython/pull/23528/files

And since `NamedTemporaryFile` relies on close being called (https://github.com/python/cpython/blob/064464fc38269e70f7e3a34cb25fc9085ab85782/Lib/tempfile.py#L501-L505)....

Now, I can add a `finally` in the big `try/except` and do some cleanup there, based on ugly flag. I agree with you that leaving the file behind sucks, but the use case is so seldom, it can only be performed by someone with a lot of karma already, and finally, it is safe from the file permission point of view.

PR #5418, opened 2021-09-19
Validate v7r3p2

atsareg:


atsareg:
MariaDB query syntax error due to recent fixes. What is the MariaDB version used in the integration tests ?

PR #5427, opened 2021-09-22
[v7r2]csv library expects bytes in py2, str in py3

chaen:
BEGINRELEASENOTES

*DMS
FIX: adapt csv use to py3

*Test
FIX: adapt csv use to py3

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
+++ b/src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
@@ -43,6 +43,7 @@ from __future__ import division
 from __future__ import print_function
 
 import csv
+import six
 from collections import defaultdict
 
 from DIRAC.Core.Utilities.DIRACScript import DIRACScript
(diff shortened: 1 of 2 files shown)


src/DIRAC/DataManagementSystem/scripts/dirac_dms_protocol_matrix.py
```diff
@@ -183,7 +184,7 @@ def main():
     gLogger.verbose("%s -> %s: %s" % (src, dst, proto))
 
   # Write the matrix in the file
-  with open(outputFile, 'wb') as csvfile:
+  with open(outputFile, 'wb') if six.PY2 else open(outputFile, 'w') as csvfile:
```
chrisburr, line 187:
```suggestion
  with open(outputFile, 'wb' if six.PY2 else 'w') as csvfile:
```

chaen: :+1:

chrisburr (approved):

PR #5455, opened 2021-10-04
[v7r2] CI: add sweep as keyword and case insensitive

chaen:
BEGINRELEASENOTES
*Tests
CHANGE: add sweep as keyword and case insensitive

ENDRELEASENOTES


chrisburr:
The labeler action doesn't consider files/directories starting with a `.`: https://github.com/actions/labeler/issues/135

As it's rare to have a PR which only modifies these files I won't worry too much for now.

PR #5460, opened 2021-10-04
[integration] Reduce number of CI jobs

chrisburr:
To reduce the load on the CI I think we can remove some CI jobs without reducing the effective coverage.

Any thoughts?


chaen:
makes a lot of sense to me

fstagni (dismissed):

fstagni (approved):

PR #5461, opened 2021-10-04
[v7r2] Sleep instead of dropping requests when services are overloaded

chrisburr:
In LHCb the changes in #5394 have turned out to occasionally cause issues with operations which need to interactions with the master CS (e.g. making edits).

After discussing with @chaen we've come to the conclusion that sleeping is better than dropping requests directly. This also allows the kernel TCP backlog to be effective.

BEGINRELEASENOTES

*Core
CHANGE: Sleep instead of dropping requests when services are overloaded

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/DISET/ServiceReactor.py
+++ b/src/DIRAC/Core/DISET/ServiceReactor.py
@@ -43,6 +43,10 @@ from DIRAC.ConfigurationSystem.Client import PathFinder
 
 __RCSID__ = "$Id$"
 
+#: Time during which the service does not accept new requests and handles those in the queue, if the backlog is too large
+#: This sleep is repeated for as long as Service.wantsThrottle is truthy
+THROTTLE_SERVICE_SLEEP_SECONDS = 0.25
+
 
 class ServiceReactor(object):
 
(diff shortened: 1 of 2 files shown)


chaen (approved):

src/DIRAC/Core/DISET/ServiceReactor.py
```diff
@@ -43,6 +43,8 @@
 
 __RCSID__ = "$Id$"
 
+THROTTLE_SERVICE_SLEEP_SECONDS = 0.25
```
andresailer, line 46: Could you add a comment what the effect of this sleep is here?
```
#: Time during which the service does not accept new requests and handles those in the queue, if the backlog is too large
```
(Did I understand the effect of this sleep correctly?)

chrisburr:
> Why labels were not added?

I misunderstood the configuration syntax, it was fixed by: https://github.com/DIRACGrid/DIRAC/commit/6a16ca18278ae03261124f7a9de9220c71f53a86

src/DIRAC/Core/DISET/ServiceReactor.py
chrisburr: As it is currently implemented it's slightly different but I'll let you read the comment to check if it's understandable.

andresailer: Yes I understand :+1: 
But please keep the `#:` so the comment shows up in the documentation page.

chrisburr: Ah I wasn't aware of that convention. Fixed!

fstagni (approved):

PR #5470, opened 2021-10-05
WIP - First version of using REST interface for ARC Computing Element

rajanandakumar:
This is a WIP to update the ARC Computing element to use the REST interface provided. This is a jumbo change for this file and I have tried to be as thorough as possible - but all normal disclaimers apply of course. So, it will be very useful to have someone look it over.

Constraints : 
1. ARC REST interface is documented in https://www.nordugrid.org/arc/arc6/tech/rest/rest.html
2. Not all sites have the REST interface turned on properly - the interface has only become "stable" fairly recently.
3. A given pilot job is identified by its long string, like for example "JyIMDm45XpznpOq3sm8tVnomAyEPUmABFKDm0DJKDmABFKDm5zDIXn".
This is exactly the identifier in the REST interface. However, in the API (and the CLI before that), it comes bundled with the CE and protocol added in, like for example
gsiftp://arcce02.esc.qmul.ac.uk:2811/jobs/JyIMDm45XpznpOq3sm8tVnomAyEPUmABFKDm0DJKDmABFKDm5zDIXn
For reasons of history and that it is easier to get information from a quick look, I decided to retain this second format and provide helper conversion functions to move between the formats in the code.

Development philosophy:
1. Keep the ARC api in place for now as a fall back option
2. For every CE, start with testing if its REST interface is properly setup
3. If the REST interface is available, use it. Otherwise fall back to the api
4. Try to make it as compatible to the API format as possible.

In the long run, if all sites enable the REST api, we will not have to use the API interface and can drop the ARC libraries from the DIRACOS.

Technology choices:
1. Python3 "requests" library for talking to the interface
2. json as the format in which we provide data to and request response from the CEs (where possible)
-  There are a few places - delegations, downloading outputs - where json is obviously not possible
3. We use ldapsearch to test if the REST interface is available for a given CE. It is assumed to be supported indefinitely (email discussion with nordugrid developers).

### DIFF
--- a/src/DIRAC/Resources/Computing/ARCComputingElement.py
+++ b/src/DIRAC/Resources/Computing/ARCComputingElement.py
@@ -2,10 +2,11 @@
 # File :   ARCComputingElement.py
 # Author : A.T.
 # Update to use ARC API : Raja Nandakumar
+# Update to use ARC REST interface keeping the API in place : Raja Nandakumar
 ########################################################################
 
 """ ARC Computing Element
-    Using the ARC API now
+    Using the REST interface now and fall back to using the ARC API if REST interface is not available.
 """
 from __future__ import absolute_import
 from __future__ import division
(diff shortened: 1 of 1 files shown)


rajanandakumar:
Going to close this PR and resubmit it with the black and other silly stuff.

PR #5480, opened 2021-10-07
[v7r2] Increased failover request delay

phicharp:
BEGINRELEASENOTES
* JobWrapper
CHANGE: We have race conditions between the termination of jobs and the failover request. There is also a problem with RAL storage as a failed upload would remove the file recovered by failover if this is too soon. Hence delay to 45 minutes the execution of failover requests

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
+++ b/src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
@@ -1288,7 +1288,7 @@ class JobWrapper(object):
     def sendFailoverRequest(self):
         """Create and send a combined job failover request if any"""
         request = Request()
-        # Forbid the request to be executed within the requested delay (default 45 minutes)
+        # Forbid the request to be executed within the requested delay
         request.delayNextExecution(self.failoverRequestDelay)
 
         requestName = "job_%s" % self.jobID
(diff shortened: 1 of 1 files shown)


src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
```diff
@@ -119,6 +119,7 @@ def __init__(self, jobID=None, jobReport=None):
         self.defaultErrorFile = gConfig.getValue(self.section + "/DefaultErrorFile", "std.err")
         self.diskSE = gConfig.getValue(self.section + "/DiskSE", ["-disk", "-DST", "-USER"])
         self.tapeSE = gConfig.getValue(self.section + "/TapeSE", ["-tape", "-RDST", "-RAW"])
+        self.failoverRequestDelay = gConfig.getValue(self.section + "/FailoverRequestDelay", 45) * 60
```
chaen, line 122: Watch out, for once the doc is correct ! 
https://github.com/DIRACGrid/DIRAC/blob/591db8ccc25bfbfd34e2c18fd4ab4f6160cdce0e/src/DIRAC/RequestManagementSystem/Client/Request.py#L334

This is minutes already

phicharp: OK, I'll fix

PR #5500, opened 2021-10-17
[v7r3] Fixes for VMDIRAC 

atsareg:
BEGINRELEASENOTES

*Resources
FIX: OpenStackCE - removed buggy line
NEW: Utilities - added possibility to specify extra yum installable packages in the configuration
NEW: Utilities - added possibility to specify ssh connection to the VM in the configuration 
FIX: cloudinit.template - use VMType as the Queue analog
FIX: dirac_resource_get_parameters - fixed bugs in the VMType getting logic

*WMS
FIX: CloudDirector - create only one pilot reference

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/ConfigurationSystem/Client/Helpers/Resources.py
+++ b/src/DIRAC/ConfigurationSystem/Client/Helpers/Resources.py
@@ -205,17 +205,16 @@ def getSiteGrid(site):
         return S_ERROR("Wrong Site Name format")
     return S_OK(sitetuple[0])
 
-
-def getQueue(site, ce, queue):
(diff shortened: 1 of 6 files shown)


sfayer (approved):
Looks fine to me... It removes some of the logic that we're using in GridPP (VMType settings in "Images" rather than "VMTypes"), but we should just update our config for that. I suggest that the v7r3 wiki release notes should mention:
 * VM settings should no longer be included in Images section (use VMTypes instead).
 * MaxCycles default for VM pilot increased to 100 (configurable in VMTypes).

PR #5526, opened 2021-10-21
[v7r2] In OperationHandlerBase: Use request id instead of request name in logger

andresailer:
RequestID is unique, and shorter than RequestName, also using getLocalSubLogger makes sense here I think.

BEGINRELEASENOTES

*RMS
CHANGE: OperationHandlerBase: use the requestID instead of the requestName in the logger
CHANGE: OperationHandlerBase: use getLocalSubLogger

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/RequestManagementSystem/private/OperationHandlerBase.py
+++ b/src/DIRAC/RequestManagementSystem/private/OperationHandlerBase.py
@@ -141,8 +141,8 @@ class OperationHandlerBase(object):
             raise TypeError("expecting Operation instance")
         self.operation = operation
         self.request = operation._parent
-        self.log = gLogger.getSubLogger(
-            "pid_%s/%s/%s/%s" % (os.getpid(), self.request.RequestName, self.request.Order, self.operation.Type)
+        self.log = gLogger.getLocalSubLogger(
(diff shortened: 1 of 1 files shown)


src/DIRAC/RequestManagementSystem/private/OperationHandlerBase.py
```diff
@@ -141,8 +141,8 @@ def setOperation(self, operation):
             raise TypeError("expecting Operation instance")
         self.operation = operation
         self.request = operation._parent
-        self.log = gLogger.getSubLogger(
-            "pid_%s/%s/%s/%s" % (os.getpid(), self.request.RequestName, self.request.Order, self.operation.Type)
+        self.log = gLogger.getLocalSubLogger(
```
chaen, line 144: Are you sure that nowhere in operations we try to get a sublogger from that?

andresailer: Actually there is one place T_T
https://github.com/DIRACGrid/DIRAC/blob/ed32a3c5a6cbf8fabab1f6f1f780267acfb2148d/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py#L63

PR #5542, opened 2021-10-25
[rel-v7r2]: WIP : First version of AREX computing element

rajanandakumar:
This is a first release of the AREX computing element. The documentation for this PR is essentially the same as that in

https://github.com/DIRACGrid/DIRAC/pull/5471

Hopefully I have addressed most of the comments in the above PR. However, if I have missed something, please do comment again here.

Some of the parts of the code need to wait for testing for the following reasons.

1. The LHCbDirac certification system is currently not accepting jobs
2. The "delegation" part of the ARC REST interface involving adding delegation ID to XRSL string at time of job submission is awaiting the next release of ARC computing software - supposedly imminent. So, this is done as per the documentation, but not yet tested.

### DIFF
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -1,0 +1,632 @@
+""" AREX Computing Element (ARC REST interface)
+    Using the REST interface now and fail if REST interface is not available.
+    A lot of the features are common with the API interface. In particular, the XRSL
+    language is used in both cases. So, we retain the xrslExtraString and xrslMPExtraString strings.
+"""
+
+__RCSID__ = "$Id$"
+
+import os
+import json
+import requests
+
+from DIRAC import S_OK, S_ERROR
+from DIRAC.Core.Security import Locations
+from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
+from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
+from DIRAC.Core.Utilities.File import makeGuid
+from DIRAC.Resources.Computing.ARCComputingElement import ARCComputingElement
+from DIRAC.Resources.Computing.ComputingElement import ComputingElement
+
+# Note : interiting from ARCComputingElement. See https://github.com/DIRACGrid/DIRAC/pull/5330#discussion_r740907255
+class AREXComputingElement(ARCComputingElement):
+    def __init__(self, ceUniqueID):
+        """Standard constructor."""
+        super(AREXComputingElement, self).__init__(ceUniqueID)
+
+        # Default REST port
+        self.port = "443"
+        # REST version to adopt
+        self.restVersion = "1.0"
+        # Time left before proxy renewal: 3 hours is a good default
+        self.proxyTimeLeftBeforeRenewal = 10800
+        # Timeout
+        self.arcRESTTimeout = 5.0
+        # Request session
+        self.session = None
+        self.headers = {}
+        # URL used to communicate with the REST interface
+        self.base_url = ""
+
+    #############################################################################
+
+    def _reset(self):
+        """Configure the Request Session to interact with the AREX REST interface.
+        Specification : https://www.nordugrid.org/arc/arc6/tech/rest/rest.html
+
+        The following needed variables are obtained from the CS. If not available, some hopefully
+        sensible defaults are set.
+        "RESTEndpoint"      - CE is Queried if not available in CS
+               - The endpoint we talk to
+        "XRSLExtraString" - Default = ""
+               - Any CE specific string with additional parameters
+        "XRSLMPExtraString" - Default = ""
+               - Any CE specific string with additional parameters for MP jobs
+        "ARCRESTTimeout"    - DEfault = 1.0 (seconds)
+               - Timeout for the rest query
+        "proxyTimeLeftBeforeRenewal" - Default = 10000 (seconds)
+               - As the name says
+
+        Note : This is not run from __init__ as the design of DIRAC means that ceParameters is
+        filled with CEDefaults only at the time this class is initialised for the given CE
+        """
+        # super()._reset()
+        self.log.debug("Testing if the REST interface is available", "for %s" % self.ceName)
+
+        # Get options from the ceParameters dictionary
+        self.port = self.ceParameters.get("Port", self.port)
+        self.restVersion = self.ceParameters.get("RESTVersion", self.restVersion)
+        self.queue = self.ceParameters.get("Queue", self.queue)
+
+        self.proxyTimeLeftBeforeRenewal = self.ceParameters.get(
+            "proxyTimeLeftBeforeRenewal", self.proxyTimeLeftBeforeRenewal
+        )
+        self.arcRESTTimeout = self.ceParameters.get("ARCRESTTimeout", self.arcRESTTimeout)
+
+        # Build the URL based on the CEName, the port and the REST version
+        service_url = "https://" + self.ceName + ":" + self.port
+        self.base_url = service_url + "/arex/rest/" + self.restVersion + "/"
+
+        # Set up the request framework
+        self.session = requests.Session()
+        self.session.verify = Locations.getCAsLocation()
+        self.headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+        }
+        # Attach the token to the headers if present
+        if os.environ.get("BEARER_TOKEN"):
+            self.headers["Authorization"] = "Bearer " + os.environ["BEARER_TOKEN"]
+
+        return S_OK()
+
+    #############################################################################
+
+    def _arcToDiracID(self, arcJobID):
+        """Convert an ARC jobID into a DIRAC jobID.
+        Example: 1234 becomes https://<ce>:<port>/arex/1234
+
+        :param str: ARC jobID
+        :return: DIRAC jobID
+        """
+        # Add CE and protocol information to pilot ID
+        if "://" in arcJobID:
+            self.log.warn("Identifier already in ARC format", arcJobID)
+            return arcJobID
+
+        diracJobID = "https://" + self.ceHost + ":" + self.port + "/arex/" + arcJobID
+        return diracJobID
+
+    def _DiracToArcID(self, diracJobID):
+        """Convert a DIRAC jobID into an ARC jobID.
+        Example: https://<ce>:<port>/arex/1234 becomes 1234
+
+        :param str: DIRAC jobID
+        :return: ARC jobID
+        """
+        # Remove CE and protocol information from pilot ID
+        if "://" in diracJobID:
+            arcJobID = diracJobID.split("arex/")[-1]
+            return arcJobID
+        self.log.warn("Identifier already in REST format?", diracJobID)
+        return diracJobID
+
+    def _UrlJoin(self, words):
+        # Return a full URL. The base_url is already defined.
+        if type(words) != type([]):
+            return "Unknown input : %s" % words
+        b_url = self.base_url.strip()
+        q_url = b_url if b_url.endswith("/") else b_url + "/"
+        for word in words:
+            w = str(word).strip()
+            w = w if w.endswith("/") else w + "/"
+            q_url = q_url + w
+        return q_url
+
+    #############################################################################
+
+    def _getDelegation(self, jobID):
+        """Here we handle the delegations (Nordugrid language) = Proxy (Dirac language)
+
+        If the jobID is empty:
+            Create and upload a new delegation to the CE and return the delegation ID.
+            This happens when the call is from the job submission function (self.submitJob).
+            We want to attach a delegation to the XRSL strings we submit for each job, so that
+            we can update this later if needed.
+            More info at
+            https://www.nordugrid.org/arc/arc6/users/xrsl.html#delegationid
+            https://www.nordugrid.org/arc/arc6/tech/rest/rest.html#delegation-functionality
+
+        If the jobID is not empty:
+            Query and return the delegation ID of the given job.
+            This happens when the call is from self.renewJobs. This function needs to know the
+            delegation associated to the job
+            More info at
+            https://www.nordugrid.org/arc/arc6/tech/rest/rest.html#jobs-management
+            https://www.nordugrid.org/arc/arc6/tech/rest/rest.html#delegations-management
+
+        :param str: job ID
+        :return: delegation ID
+        """
+        # Create a delegation
+        if not jobID:
+            # Prepare the command
+            command = "delegations"
+            params = {"action": "new"}
+            query = self._UrlJoin([command])
+            if query.startswith("Unknown"):
+                return S_ERROR("Problem creating REST query %s" % query)
+
+            # Get a proxy
+            proxy = X509Chain()
+            result = proxy.loadProxyFromFile(self.session.cert)
+            if not result["OK"]:
+                return S_ERROR("Can't load %s: %s " % (self.session.cert, result["Message"]))
+
+            # Submit a POST request
+            response = self.session.post(
+                query,
+                data=proxy.dumpAllToString(),
+                headers=self.headers,
+                params=params,
+                timeout=self.arcRESTTimeout,
+            )
+            delegationID = ""
+            if response.ok:
+                delegationURL = response.headers.get("location", "")
+                if delegationURL:
+                    delegationID = delegationURL.split("new/")[-1]
+                    # Prepare the command
+                    command = "delegations/" + delegationID
+                    query = self._UrlJoin([command])
+                    if query.startswith("Unknown"):
+                        return S_ERROR("Problem creating REST query %s" % query)
+
+                    # Submit the proxy
+                    response = self.session.put(
+                        query,
+                        data=response.text,
+                        headers=self.headers,
+                        timeout=self.arcRESTTimeout,
+                    )
+                    if not response.ok:
+                        self.log.warn(
+                            "Issue while interacting with the delegation",
+                            "%s - %s" % (response.status_code, response.reason),
+                        )
+                        delegationID = ""
+
+            return S_OK(delegationID)
+
+        # Retrieve delegation for existing job
+        else:
+            # Prepare the command
+            command = "jobs"
+            params = {"action": "delegations"}
+            query = self._UrlJoin([command])
+            if query.startswith("Unknown"):
+                return S_ERROR("Problem creating REST query %s" % query)
+
+            # Submit the POST request to get the delegation
+            jobsJson = {"job": [{"id": jobID}]}
+            response = self.session.post(
+                query,
+                data=json.dumps(jobsJson),
+                headers=self.headers,
+                timeout=self.arcRESTTimeout,
+            )
+            delegationID = ""
+            if response.ok:  # Check if the job has a delegation
+                p = response.json()
+                if "delegation_id" in p["job"]:
+                    delegationID = p["job"]["delegation_id"][0]
+            return S_OK(delegationID)
+
+    #############################################################################
+
+    def submitJob(self, executableFile, proxy, numberOfJobs=1):
+        """Method to submit job
+        Assume that the ARC queues are always of the format nordugrid-<batchSystem>-<queue>
+        And none of our supported batch systems have a "-" in their name
+        """
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot submit jobs.")
+        self.log.verbose("Executable file path: %s" % executableFile)
+
+        # Get the name of the queue: nordugrid-<batchsystem>-<queue>
+        self.arcQueue = self.queue.split("-", 2)[2]
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("Failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # Prepare the command
+        command = "jobs"
+        params = {"action": "new"}
+        query = self._UrlJoin([command])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+
+        # Get a "delegation" and use the same delegation for all the jobs
+        delegation = ""
+        result = self._getDelegation("")
+        if not result["OK"]:
+            self.log.warn("Could not get a delegation", "For CE %s" % self.ceHost)
+            self.log.warn("Continue without a delegation")
+        else:
+            delegation = "(delegationid=%s)" % result["Value"]
+
+        # Submit multiple jobs sequentially.
+        # Bulk submission would not be significantly faster than multiple single submission.
+        # https://www.nordugrid.org/arc/arc6/tech/rest/rest.html#job-submission-create-a-new-job
+        # Also : https://bugzilla.nordugrid.org/show_bug.cgi?id=4069
+        batchIDList = []
+        stampDict = {}
+        for _ in range(numberOfJobs):
+            # Get the job into the ARC way
+            xrslString, diracStamp = self._writeXRSL(executableFile)
+            if len(delegation) > 5:
+                xrslString = xrslString + delegation
+            self.log.debug("XRSL string submitted", "is %s" % xrslString)
+            self.log.debug("DIRAC stamp for job", "is %s" % diracStamp)
+
+            # Submit the POST request
+            response = self.session.post(
+                query,
+                data=xrslString,
+                headers=self.headers,
+                params=params,
+                timeout=self.arcRESTTimeout,
+            )
+            if not response.ok:
+                self.log.warn(
+                    "Failed to submit job",
+                    "to CE %s with error - %s - and messages : %s"
+                    % (self.ceHost, response.status_code, response.reason),
+                )
+                break
+
+            responseJob = response.json()["job"]
+            if responseJob["status-code"] > "400":
+                self.log.warn(
+                    "Failed to submit job",
+                    "to CE %s with error - %s - and messages: %s"
+                    % (self.ceHost, responseJob["status-code"], responseJob["reason"]),
+                )
+                break
+
+            jobID = responseJob["id"]
+            pilotJobReference = self._arcToDiracID(jobID)
+            batchIDList.append(pilotJobReference)
+            stampDict[pilotJobReference] = diracStamp
+            self.log.debug(
+                "Successfully submitted job",
+                "%s to CE %s" % (pilotJobReference, self.ceHost),
+            )
+
+            # At this point, only the XRSL job has been submitted to AREX services
+            # Here we also upload the executable.
+            command = "jobs/" + jobID + "/session/" + os.path.basename(executableFile)
+            query = self._UrlJoin([command])
+            if query.startswith("Unknown"):
+                return S_ERROR("Problem creating REST query %s" % query)
+
+            # Extract the content of the file
+            with open(executableFile, "r") as f:
+                content = f.read()
+
+            # Submit the PUT request
+            response = self.session.put(query, data=content, headers=self.headers, timeout=self.arcRESTTimeout)
+            if response.ok:
+                self.log.info("Executable correctly uploaded")
+
+        if batchIDList:
+            result = S_OK(batchIDList)
+            result["PilotStampDict"] = stampDict
+        else:
+            result = S_ERROR("No pilot references obtained from the ARC job submission")
+        return result
+
+    #############################################################################
+
+    def killJob(self, jobIDList):
+        """Kill the specified jobs
+
+        :param list: list of DIRAC Job IDs
+        """
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot kill jobs.")
+        self.log.debug("Killing jobs", ",".join(jobIDList))
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("Failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # List of jobs in json format for the REST query
+        jList = [self._DiracToArcID(job) for job in jobIDList]
+        jobsJson = {"job": [{"id": job} for job in jList]}
+
+        # Prepare the command
+        command = "jobs"
+        params = {"action": "kill"}
+        query = self._UrlJoin([command])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+
+        # Killing jobs should be fast
+        response = self.session.post(
+            query,
+            data=json.dumps(jobsJson),
+            headers=self.headers,
+            params=params,
+            timeout=self.arcRESTTimeout,
+        )
+        if not response.ok:
+            return S_ERROR("Failed to kill all these jobs: %s %s" % (response.status_code, response.reason))
+
+        self.log.debug("Successfully deleted jobs %s " % (response.json()))
+        return S_OK()
+
+    #############################################################################
+
+    def getCEStatus(self):
+        """Method to return information on running and pending jobs.
+        Query the CE directly to get the number of waiting and running jobs for the given
+        VO and queue.
+        The specification is apparently in glue2 and if you do a raw print out of the information
+        it goes on and on as it dumps the full configuration of the CE for all VOs, queues,
+        states and parameters. Hopefully this function weeds out everything except the relevant
+        queue.
+        """
+
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot get CE status.")
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("Failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # Try to find out which VO we are running for.
+        # Essential now for REST interface.
+        res = getVOfromProxyGroup()
+        vo = res["Value"] if res["OK"] else ""
+
+        # Prepare the command
+        command = "info"
+        params = {"schema": "glue2"}
+        query = self._UrlJoin([command])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+
+        # Submit the GET request
+        response = self.session.get(query, headers=self.headers, params=params, timeout=self.arcRESTTimeout)
+        if not response.ok:
+            res = S_ERROR("Unknown failure for CE %s. Is the CE down?" % self.ceHost)
+            return res
+        ceData = response.json()
+
+        # Look only in the relevant section out of the headache
+        queueInfo = ceData["Domains"]["AdminDomain"]["Services"]["ComputingService"]["ComputingShare"]
+        if not isinstance(queueInfo, list):
+            queueInfo = [queueInfo]
+
+        # I have only seen the VO published in lower case ...
+        result = S_OK()
+        result["SubmittedJobs"] = 0
+
+        magic = self.queue + "_" + vo.lower()
+        for i in range(len(queueInfo)):
+            if queueInfo[i]["ID"].endswith(magic):
+                result["RunningJobs"] = queueInfo[i]["RunningJobs"]
+                result["WaitingJobs"] = queueInfo[i]["WaitingJobs"]
+                break  # Pick the first (should be only ...) matching queue + VO
+
+        return result
+
+    #############################################################################
+
+    def _renewJobs(self, jobList):
+        """Written for the REST interface - jobList is already in the ARC format
+        This function is called only by this class, NOT by the SiteDirector
+        """
+        # Renew the jobs
+        for job in jobList:
+            # First get the delegation (proxy)
+            result = self._getDelegation(job)
+            if not result["OK"]:
+                self.log.warn("Could not get a delegation from", "Job %s" % job)
+                continue
+            delegationID = result["Value"]
+
+            # Prepare the command
+            command = "delegations/" + delegationID
+            params = {"action": "get"}
+            query = self._UrlJoin([command])
+            if query.startswith("Unknown"):
+                return S_ERROR("Problem creating REST query %s" % query)
+
+            # Submit the POST request to get the proxy
+            response = self.session.post(query, headers=self.headers, params=params, timeout=self.arcRESTTimeout)
+            proxy = X509Chain()
+            res = proxy.loadChainFromString(response.text)
+
+            # Now test and renew the proxy
+            if not res["OK"]:
+                continue
+
+            timeLeft = proxy.getRemainingSecs()
+            if timeLeft < self.proxyTimeLeftBeforeRenewal:
+                self.log.debug(
+                    "Renewing proxy for job",
+                    "%s whose proxy expires at %s" % (job, timeLeft),
+                )
+                # Proxy needs to be renewed - try to renew it
+                command = "delegations/" + delegationID
+                params = {"action": "renew"}
+                query = self._UrlJoin([command])
+                if query.startswith("Unknown"):
+                    return S_ERROR("Problem creating REST query %s" % query)
+                response = self.session.post(
+                    query,
+                    headers=self.headers,
+                    params=params,
+                    timeout=self.arcRESTTimeout,
+                )
+                if response.ok:
+                    self.log.debug("Proxy successfully renewed", "for job %s" % job)
+                else:
+                    self.log.debug(
+                        "Proxy not renewed",
+                        "for job %s with delegation %s" % (job, delegationID),
+                    )
+            else:  # No need to renew. Proxy is long enough
+                continue
+        return S_OK()
+
+    #############################################################################
+
+    def getJobStatus(self, jobIDList):
+        """Get the status information for the given list of jobs"""
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot get job status.")
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("AREXComputingElement: failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        if type(jobIDList) != type([]):
+            jobTmpList = [jobIDList]
+
+        # Pilots are stored with a DIRAC stamp (":::XXXXX") appended
+        jobList = []
+        for j in jobTmpList:
+            job = j.split(":::")[0]
+            jobList.append(job)
+
+        self.log.debug("Getting status of jobs : %s" % jobList)
+        jobsJson = {"job": [{"id": self._DiracToArcID(job)} for job in jobList]}
+
+        # Prepare the command
+        command = "jobs"
+        params = {"action": "status"}
+        query = self._UrlJoin([command])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+
+        # Submit the POST request to get status of the pilots
+        response = self.session.post(
+            query,
+            data=json.dumps(jobsJson),
+            headers=self.headers,
+            params=params,
+            timeout=self.arcRESTTimeout,
+        )
+        if not response.ok:
+            self.log.info(
+                "Failed getting the status of the jobs",
+                "%s - %s" % (response.status_code, response.reason),
+            )
+            return S_ERROR("Failed getting the status of the jobs")
+
+        resultDict = {}
+        jobsToRenew = []
+        jobsToCancel = []
+        for job in response.json()["job"]:
+            jobID = self._arcToDiracID(job["id"])
+            # ARC REST interface returns hyperbole
+            arcState = job["state"].capitalize()
+            self.log.debug("REST ARC status", "for job %s is %s" % (jobID, arcState))
+            resultDict[jobID] = self.mapStates[arcState]
+
+            # Renew proxy only of jobs which are running or queuing
+            if arcState in ("Running", "Queuing"):
+                jobsToRenew.append(job["id"])
+            # Cancel held jobs so they don't sit in the queue forever
+            if arcState == "Hold":
+                jobsToCancel.append(job["id"])
+                self.log.debug("Killing held job %s" % jobID)
+
+        # Renew jobs to be renewed
+        # Does not work at present - wait for a new release of ARC CEs for this.
+        result = self._renewJobs(jobsToRenew)
+        if not result["OK"]:
+            return result
+
+        # Kill jobs to be killed
+        result = self.killJob(jobsToCancel)
+        if not result["OK"]:
+            return result
+
+        return S_OK(resultDict)
+
+    #############################################################################
+
+    def getJobOutput(self, jobID, localDir=None):
+        """Get the specified job standard output and error files.
+        The output is returned as strings.
+        """
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot get job output.")
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("AREXComputingElement: failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # Extract stamp from the Job ID
+        if ":::" in jobID:
+            pilotRef, stamp = jobID.split(":::")
+        else:
+            pilotRef = jobID
+            stamp = ""
+        if not stamp:
+            return S_ERROR("Pilot stamp not defined for %s" % pilotRef)
+
+        # Prepare the command
+        command = "jobs/"
+        job = self._DiracToArcID(pilotRef)
+        query = self._UrlJoin([command, job, "session", stamp, ".out"])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+
+        # Submit the GET request to retrieve outputs
+        response = self.session.get(query, headers=self.headers, timeout=self.arcRESTTimeout)
+        if not response.ok:
+            self.log.error("Error downloading stdout", "for %s: %s" % (job, response.text))
+            return S_ERROR("Failed to retrieve at least some output for %s" % jobID)
+        output = response.text
+
+        query = self._UrlJoin([command, job, "session", stamp, ".err"])
+        if query.startswith("Unknown"):
+            return S_ERROR("Problem creating REST query %s" % query)
+        response = self.session.get(query, headers=self.headers, timeout=self.arcRESTTimeout)
+        if not response.ok:
+            self.log.error("Error downloading stderr", "for %s: %s" % (job, response.text))
+            return S_ERROR("Failed to retrieve at least some output for %s" % jobID)
+        error = response.text
+
+        return S_OK((output, error))
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -516,8 +516,8 @@ class AREXComputingElement(ARCComputingElement):
             return result
         self.session.cert = Locations.getProxyLocation()
 
-        if type(jobIDList) != type([]):
-            jobTmpList = [jobIDList]
+        if not isinstance(jobIDList, list):
+            jobIDList = [jobIDList]
 
         # Pilots are stored with a DIRAC stamp (":::XXXXX") appended
         jobList = []
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -4,6 +4,10 @@
     language is used in both cases. So, we retain the xrslExtraString and xrslMPExtraString strings.
 """
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
 __RCSID__ = "$Id$"
 
 import os
@@ -14,9 +18,7 @@ from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Security import Locations
 from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
 from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
-from DIRAC.Core.Utilities.File import makeGuid
 from DIRAC.Resources.Computing.ARCComputingElement import ARCComputingElement
-from DIRAC.Resources.Computing.ComputingElement import ComputingElement
 
 # Note : interiting from ARCComputingElement. See https://github.com/DIRACGrid/DIRAC/pull/5330#discussion_r740907255
 class AREXComputingElement(ARCComputingElement):
@@ -123,7 +125,7 @@ class AREXComputingElement(ARCComputingElement):
 
     def _UrlJoin(self, words):
         # Return a full URL. The base_url is already defined.
-        if type(words) != type([]):
+        if not isinstance(words, list):
             return "Unknown input : %s" % words
         b_url = self.base_url.strip()
         q_url = b_url if b_url.endswith("/") else b_url + "/"
@@ -278,8 +280,7 @@ class AREXComputingElement(ARCComputingElement):
         for _ in range(numberOfJobs):
             # Get the job into the ARC way
             xrslString, diracStamp = self._writeXRSL(executableFile)
-            if len(delegation) > 5:
-                xrslString = xrslString + delegation
+            xrslString += delegation
             self.log.debug("XRSL string submitted", "is %s" % xrslString)
             self.log.debug("DIRAC stamp for job", "is %s" % diracStamp)
 
@@ -521,7 +522,7 @@ class AREXComputingElement(ARCComputingElement):
 
         # Pilots are stored with a DIRAC stamp (":::XXXXX") appended
         jobList = []
-        for j in jobTmpList:
+        for j in jobIDList:
             job = j.split(":::")[0]
             jobList.append(job)
 
@@ -583,7 +584,7 @@ class AREXComputingElement(ARCComputingElement):
 
     #############################################################################
 
-    def getJobOutput(self, jobID, localDir=None):
+    def getJobOutput(self, jobID, _localDir=None):
         """Get the specified job standard output and error files.
         The output is returned as strings.
         """
--- a/src/DIRAC/Resources/Computing/ARCComputingElement.py
+++ b/src/DIRAC/Resources/Computing/ARCComputingElement.py
@@ -1,9 +1,3 @@
-########################################################################
-# File :   ARCComputingElement.py
-# Author : A.T.
-# Update to use ARC API : Raja Nandakumar
-########################################################################
-
 """ ARC Computing Element
     Using the ARC API now
 """
@@ -55,8 +49,7 @@ class ARCComputingElement(ComputingElement):
         self.usercfg = arc.common.UserConfig()
         # set the timeout to the default 20 seconds in case the UserConfig constructor did not
         self.usercfg.Timeout(20)  # pylint: disable=pointless-statement
-        self.ceHost = self.ceParameters.get("Host", self.ceName)
-        self.gridEnv = self.ceParameters.get("GridEnv", self.gridEnv)
+        self.gridEnv = ""
         # Used in getJobStatus
         self.mapStates = {
             "Accepted": "Scheduled",
@@ -73,9 +66,9 @@ class ARCComputingElement(ComputingElement):
             "Finished": "Done",
             "Other": "Done",
         }
-        # Do these after all other initialisations, in case something barks
-        self.xrslExtraString = self.__getXRSLExtraString()
-        self.xrslMPExtraString = self.__getXRSLExtraString(multiprocessor=True)
+        # Extra XRSL info
+        self.xrslExtraString = ""
+        self.xrslMPExtraString = ""
 
     #############################################################################
 
@@ -95,59 +88,6 @@ class ARCComputingElement(ComputingElement):
         j.ServiceInformationInterfaceName = "org.nordugrid.ldapng"
         j.PrepareHandler(self.usercfg)
         return j
-
-    def __getXRSLExtraString(self, multiprocessor=False):
-        # For the XRSL additional string from configuration - only done at initialisation time
-        # If this string changes, the corresponding (ARC) site directors have to be restarted
-        #
-        # Variable = XRSLExtraString (or XRSLMPExtraString for multi processor mode)
-        # Default value = ''
-        #   If you give a value, I think it should be of the form
-        #          (aaa = "xxx")
-        #   Otherwise the ARC job description parser will have a fit
-        # Locations searched in order :
-        # Top priority    : Resources/Sites/<Grid>/<Site>/CEs/<CE>/XRSLExtraString
-        # Second priority : Resources/Sites/<Grid>/<Site>/XRSLExtraString
-        # Default         : Resources/Computing/CEDefaults/XRSLExtraString
-        #
-        xrslExtraString = ""  # Start with the default value
-        result = getCESiteMapping(self.ceHost)
-        if not result["OK"] or not result["Value"]:
-            self.log.error("Unknown CE ...")
-            return
-        self.site = result["Value"][self.ceHost]
-        # Now we know the site. Get the grid
-        grid = self.site.split(".")[0]
-        # The different possibilities that we have agreed upon
-        if multiprocessor:
-            xtraVariable = "XRSLMPExtraString"
-        else:
-            xtraVariable = "XRSLExtraString"
-        firstOption = "Resources/Sites/%s/%s/CEs/%s/%s" % (grid, self.site, self.ceHost, xtraVariable)
-        secondOption = "Resources/Sites/%s/%s/%s" % (grid, self.site, xtraVariable)
-        defaultOption = "Resources/Computing/CEDefaults/%s" % xtraVariable
-        # Now go about getting the string in the agreed order
-        self.log.debug("Trying to get %s : first option %s" % (xtraVariable, firstOption))
-        result = gConfig.getValue(firstOption, defaultValue="")
-        if result != "":
-            xrslExtraString = result
-            self.log.debug("Found %s : %s" % (xtraVariable, xrslExtraString))
-        else:
-            self.log.debug("Trying to get %s : second option %s" % (xtraVariable, secondOption))
-            result = gConfig.getValue(secondOption, defaultValue="")
-            if result != "":
-                xrslExtraString = result
-                self.log.debug("Found %s : %s" % (xtraVariable, xrslExtraString))
-            else:
-                self.log.debug("Trying to get %s : default option %s" % (xtraVariable, defaultOption))
-                result = gConfig.getValue(defaultOption, defaultValue="")
-                if result != "":
-                    xrslExtraString = result
-                    self.log.debug("Found %s : %s" % (xtraVariable, xrslExtraString))
-        if xrslExtraString:
-            self.log.always("%s : %s" % (xtraVariable, xrslExtraString))
-            self.log.always(" --- to be added to pilots going to CE : %s" % self.ceHost)
-        return xrslExtraString
 
     #############################################################################
     def _addCEConfigDefaults(self):
@@ -197,8 +137,13 @@ class ARCComputingElement(ComputingElement):
     #############################################################################
     def _reset(self):
         self.queue = self.ceParameters.get("CEQueueName", self.ceParameters["Queue"])
-        if "GridEnv" in self.ceParameters:
-            self.gridEnv = self.ceParameters["GridEnv"]
+        self.ceHost = self.ceParameters.get("Host", self.ceHost)
+        self.gridEnv = self.ceParameters.get("GridEnv", self.gridEnv)
+
+        # extra XRSL data (should respect the XRSL format)
+        self.xrslExtraString = self.ceParameters.get("XRSLExtraString", self.xrslExtraString)
+        self.xrslMPExtraString = self.ceParameters.get("XRSLMPExtraString", self.xrslMPExtraString)
+
         return S_OK()
 
     #############################################################################


src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,641 @@
...
+        # Get the REST endpoint (service_url in ARC language) from the CS if available (preferred).
+        # It should be at
+        #   Resources/Sites/<Grid>/<Site>/CEs/<CE>/RESTEndpoint
+        # Otherwise query the CE for this value.
+        result = returnValueOrRaise(getCESiteMapping(self.ceHost))
+        self.site = result[self.ceHost]
+        grid = self.site.split(".")[0]
+
+        # Try getting service_url from the CS
+        ceString = "Resources/Sites/%s/%s/CEs/%s/RESTEndpoint" % (grid, self.site, self.ceHost)
```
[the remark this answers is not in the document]
rajanandakumar: Thanks.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,641 @@
...
+
+        if not self.initialised:  # Something went wrong in the initialisation. Redo it.
+            result = self._configureRestInterface()
+            if not result["OK"]:
+                self.log.debug("CE %s still not properly initialized. Aborting job submission." % self.ceHost)
+                return S_ERROR("REST interface not initialised.")
+        self.s.cert = Locations.getProxyLocation()
+
+        # Will this ever work? Has this ever worked / been tested? See the getJobStatus function below?
+        self.log.debug("Killing jobs", "%s" % jobIDList)
```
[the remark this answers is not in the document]
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,641 @@
...
+                pilotJobReference = self._pilot_toAPI(json.loads(r.text.replace("\n", ""))["job"]["id"])
+                batchIDList.append(pilotJobReference)
+                stampDict[pilotJobReference] = diracStamp
+                self.log.debug("Successfully submitted job", "%s to CE %s" % (pilotJobReference, self.ceHost))
+            else:
+                self.log.warn("Failed to submit job", "to CE %s with error %s." % (self.ceHost, r.status_code))
+                self.log.warn("Message from CE : %s" % json.loads(r.text.replace("\n", "")))
+                self.log.warn("The above message may not help. You may in that case either check")
+                self.log.warn("gocdB (goc.egi.eu) for downtime or contact the site directly.")
+                self.log.warn("Or maybe ask an expert for help ...")
```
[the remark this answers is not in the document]
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,641 @@
...
+            if len(service_url) < 5:  # CE is probably down. Wait for next iteration to test again
+                self.log.debug("CE %s probably down. Please contact the site admin." % self.ceHost)
+                self.initialised = False
+                return S_ERROR("CE probably down. Please contact site admin.")
+
+        # We now have a pointer (service_url) to the REST interface for this CE
+        restVersion = "1.0"  # Will be this value for the forseeable future.
+        self.base_url = service_url + "/rest/" + restVersion + "/"
+        command = "info?schema=glue2"
+        query = self.base_url + command
```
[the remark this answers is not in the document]
rajanandakumar: It is already done. Thanks!

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,641 @@
...
+            # ldapsearch -x -LLL -h grendel.hec.lancs.ac.uk:2135 -b 'o=glue' GLUE2EndpointInterfaceName=org.nordugrid.arcrest GLUE2EndpointURL | grep GLUE2EndpointURL | awk -F ": " '{print $2}'
+            command = """ ldapsearch -x -LLL -h %s:2135 -b 'o=glue' \
+                    GLUE2EndpointInterfaceName=org.nordugrid.arcrest GLUE2EndpointURL \
+                    | grep GLUE2EndpointURL | awk -F ": " '{print $2}' """
+            status = shellCall(1, command % self.ceHost)
+            service_url = ""
+            if status["OK"]:
+                service_url = status["Value"][1].strip()  # All well
+            if len(service_url) < 5:  # CE is probably down. Wait for next iteration to test again
+                self.log.debug("CE %s probably down. Please contact the site admin." % self.ceHost)
```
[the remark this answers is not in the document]
rajanandakumar: Fair enough. Changed to warning.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("AREXComputingElement: failed to set up proxy", result["Message"])
+            return result
+
+        if not self.rest:  # Something went wrong in the initialisation. Redo it.
+            result = self._configureRestInterface()
+            if not result["OK"]:
+                self.log.debug("CE %s still not properly initialized. Aborting job submission." % self.ceHost)
+                return S_ERROR("REST interface not initialised.")
```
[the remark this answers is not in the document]
rajanandakumar: Thanks for this. It took me a bit of time to understand what this code was about. I have implemented it and will test it by and by.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+            # # We need to write the proxy out to get its information
+            # tmpProxyFile = "/tmp/arcRestRenew-" + makeGuid()[:8]
+            # with open(tmpProxyFile, 'w') as outFile: outFile.write(r.text)
+            # proxy = getProxyInfo(tmpProxyFile)
+            # os.unlink(tmpProxyFile) # Cleanup
+
+            # Now test and renew the proxy
+            if proxy["OK"] and "secondsLeft" in proxy["Value"]:
+                timeLeft = int(proxy["Value"]["secondsLeft"])
+                if timeLeft < 10000:  # 2 hours, 46 minutes and 40 seconds
```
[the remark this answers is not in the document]
rajanandakumar: Done ...

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+        params = {"schema": "glue2"}
+        query = self.base_url + command
+        r = self.s.get(query, headers=self.headers, params=params, stream=True, timeout=1.0)
+        r.raw.decode_content = True  # Needed for "lots" of information (like the stream argument above)?
+
+        if not r.ok:
+            res = S_ERROR("Unknown failure for CE %s. Is the CE down?" % self.ceHost)
+            return res
+
+        p = json.loads(r.text.replace("\n", ""))
```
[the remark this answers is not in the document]
chaen: It should not

```python
In [4]: withSpace = '{}\n\n\n\n'

In [5]: json.loads(withSpace)
Out[5]: {}
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+            # First get the delegation (proxy)
+            dID = self._delegation(job)
+            if len(dID) < 2:  # No delegation.
+                continue
+
+            # Get the proxy
+            command = "delegations/" + dID
+            params = {"action": "get"}
+            query = self.base_url + command
+            r = self.s.post(query, headers=self.headers, params=params, timeout=1.0)
```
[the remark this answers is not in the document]
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+        if result != "":
+            service_url = result  # All well
+        else:
+            # Extract the endpoint from the CE directly if it is not in the CS already. The following
+            # command should expand to for example
+            # ldapsearch -x -LLL -h grendel.hec.lancs.ac.uk:2135 -b 'o=glue' GLUE2EndpointInterfaceName=org.nordugrid.arcrest GLUE2EndpointURL | grep GLUE2EndpointURL | awk -F ": " '{print $2}'
+            command = """ ldapsearch -x -LLL -h %s:2135 -b 'o=glue' \
+                    GLUE2EndpointInterfaceName=org.nordugrid.arcrest GLUE2EndpointURL \
+                    | grep GLUE2EndpointURL | awk -F ": " '{print $2}' """
+            status = shellCall(1, command % self.ceHost)
```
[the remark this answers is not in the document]
rajanandakumar: Done. Thanks!

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
chrisburr: The only way I can imagine getting that error is if you have invalid JSON like:

```python
In [1]: import json
   ...: json.loads('{"job_name": "hello\nworld"}')
JSONDecodeError                           Traceback (most recent call last)
<ipython-input-1-4ada5a6af7b6> in <module>
      1 import json
----> 2 json.loads('{"job_name": "hello\nworld"}')

~/miniconda3/envs/dirac-development/lib/python3.9/json/__init__.py in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    344             parse_int is None and parse_float is None and
    345             parse_constant is None and object_pairs_hook is None and not kw):
--> 346         return _default_decoder.decode(s)
    347     if cls is None:
    348         cls = JSONDecoder

~/miniconda3/envs/dirac-development/lib/python3.9/json/decoder.py in decode(self, s, _w)
    335
    336         """
--> 337         obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    338         end = _w(s, end).end()
    339         if end != len(s):

~/miniconda3/envs/dirac-development/lib/python3.9/json/decoder.py in raw_decode(self, s, idx)
    351         """
    352         try:
--> 353             obj, end = self.scan_once(s, idx)
    354         except StopIteration as err:
    355             raise JSONDecodeError("Expecting value", s, err.value) from None

JSONDecodeError: Invalid control character at: line 1 column 20 (char 19)

In [2]: import json
   ...: json.loads('{"job_name": "hello\\nworld"}')
Out[2]: {'job_name': 'hello\nworld'}
```

If so that's something the ARC developers should fix as it's explictly forbidden in the [section 9 of the JSON spec](https://www.ecma-international.org/publications-and-standards/standards/ecma-404/):

> A string is a sequence of Unicode code points wrapped with quotation marks (U+0022). All code points may be placed within the quotation marks except for the code points that must be escaped: quotation mark (U+0022), reverse solidus (U+005C), and the control characters U+0000 to U+001F. There are two-character escape sequence representations of some characters.
> * \\" represents the quotation mark character (U+0022).
> * \\\\ represents the reverse solidus character (U+005C).
> * \\/ represents the solidus character (U+002F).
> * \\b represents the backspace character (U+0008).
> * \\f represents the form feed character (U+000C).
> * \\n represents the line feed character (U+000A).
> * \\r represents the carriage return character (U+000D).
> * \\t represents the character tabulation character (U+0009).

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,650 @@
...
+
+        # Try getting service_url from the CS
+        ceString = cfgPath("Resources", "Sites", self.grid, self.site, "CEs", self.ceHost, "RESTEndpoint")
+        result = self._getFromCfgSvc(ceString)
+        if result["OK"]:
+            service_url = result["Value"]  # All well
+        if len(service_url) < 5:  # There is no endpoint in the CS. Discover it.
+            # The following command should expand to for example
+            # ldapsearch -x -LLL -h grendel.hec.lancs.ac.uk:2135 -b 'o=glue' GLUE2EndpointInterfaceName=org.nordugrid.arcrest GLUE2EndpointURL | grep GLUE2EndpointURL | awk -F ": " '{print $2}'
+            server = ldap3.Server("ldap://self.ceHost:2135")
```
chrisburr, line 146: I guess this is a typo?
```suggestion
            server = ldap3.Server("ldap://" + self.ceHost + "2135")
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,650 @@
...
+            try:
+                connection = ldap3.Connection(server, client_strategy=ldap3.SAFE_SYNC, auto_bind=True)
+            except ldap3.core.exceptions.LDAPSocketOpenError:
+                self.log.error("Could not connect to server", "CE : %s:2135" % self.ceHost)
+                return False
+            status, result, response, _ = connection.search(
+                "o=glue", "(GLUE2EndpointInterfaceName=org.nordugrid.arcrest)", attributes="GLUE2EndpointURL"
+            )
+            if not status:
```
chrisburr, line 155: Is it possible for this to return no results?
```suggestion
            if not status or len(response) == 0:
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Thanks! Committed as it is.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Good point! Thanks!

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: I will point it out again and get back on this.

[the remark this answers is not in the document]
rajanandakumar: ARC developers have finally agreed that there is a problem - and fixed it! To come in the next release of ARC software(v6.14) as I understand it. So, hopefully in the next 2 weeks or so. Will remove this hack (and other similar lines) once the fix within ARC is confirmed to work. Keeping open till then.

rajanandakumar:
Following a series of updates / fixes based on comments / suggestions and lots of help (many thanks!), this WIP has to pause for a few weeks with the following outstanding points. The points we are waiting for are :

1. New release of ARC
2. More documentation, tests in the DIRAC project
3. Fixes to the LHCb certification system.

In principle the code is in an "almost final" state, only pending more testing. The actual points from the comments that are waiting for resolution are

1. Variable name "jj"
https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r736571741
Variable name seems okay to me. But happy to listen to alternative suggestions.

2. json decoding
https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r737278510
ARC developers have finally agreed that there is a problem - and fixed it! To come in the next release of ARC software(v6.14) as I understand it.

3. Proxy management
https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r737313124
Waiting for testing ability after next release of ARC software (v6.14) and its installation on a few test CEs.

4. Coding style
https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r737908770
Wait for resolution of point 2 above.

5. Changing of working directory during output download
https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r737297807
Waiting for ability to submit jobs to LHCb certification instance to test this part.

6. More detailed testing of the module
Waiting for ability to submit jobs to LHCb certification instance to test this part.

7. tests
Waiting for example tests from other modules

8. Documentation
Similar to other CEs. Waiting for example to be available

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+        "Undefined": "Unknown",
+        "Running": "Running",
+        "Finishing": "Running",
+        "Deleted": "Killed",
+        "Killed": "Killed",
+        "Failed": "Failed",
+        "Hold": "Failed",
+        "Finished": "Done",
+        "Done": "Done",
+        "Other": "Done",
```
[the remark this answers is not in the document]
rajanandakumar: No idea - obviously one of the many possible options for the "Done" status that we found when running our tests in 2015(!) from the git logs. This is apparently dependent on the batch system used by the site - mostly HTCondor, but  sites are known to use other batch systems than this with ARC.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+                return S_OK(result)
+        # Likely none of the paths were valid in the CS
+        return S_OK()
+
+    def _pilot_toAPI(self, pilot):
+        # Add CE and protocol information to pilot ID
+        if "://" in pilot:
+            self.log.warning("Pilot already in API format", "%s" % pilot)
+            return pilot
+        pilotAPI = "gsiftp://" + self.ceHost + "/" + pilot
```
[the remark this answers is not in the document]
rajanandakumar: It is a string representing the pilot which is stored in the pilotDB and communicated with the CE. We are not using the urllib in this module. What benefit do we get from using urllib?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+            r = self.s.post(query, data=json.dumps(jj), headers=self.headers, timeout=self.arcRESTTimeout)
+            dID = ""
+            if r.ok:  # Check if the job has a delegation
+                p = json.loads(r.text.replace("\n", ""))
+                if "delegation_id" in p["job"]:
+                    dID = p["job"]["delegation_id"][0]
+            return dID
+
+    #############################################################################
+    def submitJob(self, executableFile, proxy, numberOfJobs=1):
```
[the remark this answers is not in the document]
rajanandakumar: Yikes. What should I do?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+            return pilot
+        pilotAPI = "gsiftp://" + self.ceHost + "/" + pilot
+        return pilotAPI
+
+    def _pilot_toREST(self, pilot):
+        # Remove CE and protocol information from pilot ID
+        if "://" in pilot:
+            pilotREST = pilot.split("jobs/")[-1]
+            return pilotREST
+        else:
```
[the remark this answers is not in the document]
rajanandakumar: Does the same comment apply also to the function _pilot_toAPI?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+from DIRAC.Core.Utilities.File import makeGuid
+from DIRAC.Core.Utilities.List import breakListIntoChunks
+from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
+from DIRAC.Resources.Computing.ARCComputingElement import ARCComputingElement
+from DIRAC.Core.Security.ProxyInfo import getProxyInfo
+from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
+from DIRAC.Core.Security import Locations
+from DIRAC.Core.Utilities.ReturnValues import returnValueOrRaise
+from DIRAC.ConfigurationSystem.Client.Helpers.Path import cfgPath
+import ldap3
```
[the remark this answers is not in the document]
rajanandakumar: Will do at next commit.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Maybe it is time to update this set of states after we are able to run tests.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,648 @@
...
+        query = self.base_url + command
+        # First get a "delegation" and use the same delegation for all the jobs
+        deleg = self._delegation("")
+        deleText = ""
+        if len(deleg) < 2:
+            self.log.warning("Could not get a delegation", "For CE %s", self.ceHost)
+            self.log.warning("Continue without a delegation")
+        else:
+            deleText = "(delegationid=%s)" % deleg
+        # It is fairly simple to change to bulk submission. Should I do so?
```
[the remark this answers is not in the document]
rajanandakumar: As an explanation : XRSL uses "+" to merge jobs. So, if we want to submit multiple jobs, we need a long string along the lines of

+(&(...))(&(...))(&(...))

where each "&(...)" is a pilot job for us.

I do not know how big each job in DIRAC is in terms of the length of the string used. As a result, if we put together too many jobs for multiple submission, I worry the following two points may affect the submission 

1. Internal string length within python
2. How much "data" can the requests module of python support (https://docs.python-requests.org/en/master/index.html)?

The above questions are for experts in the field to answer. I would naively guess that they are not an issue, but wanted to get a basic version into production without having to debug this as a possible problem.

If neither of the above is a problem, we can happily use the above prescription to conflate multiple jobs and submit them in one interaction with the CE.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: yes, please do if you can!

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: It would make sure that the string is correctly constructed (e.g. "correct" number of "/").

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: It is already correct there I see.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: Nothing specifically to do (changes in the PR above will come in v7r3). Just beware of that, in case of calls to `submitJob`.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: I don't think your worries are an issue, so you can conflate multiple jobs in one go.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: So I do not understand and need some handholding please. What is incorrect here?

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Thanks. This code does not call the function submitJob. It is happily left in the hands of the siteDirector.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Okay - thanks. Once I am able to submit to the certification system again, I will test this and implement it.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: nothing is incorrect, but this structure:

```
if x:
    return y
else:
    return z
```
is equivalent to:
```
if x:
    return y
return z
```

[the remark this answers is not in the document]
rajanandakumar: Ah - I am an old programmer 😄

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
chrisburr: > Internal string length within python

However much RAM you have

> How much "data" can the requests module of python support (https://docs.python-requests.org/en/master/index.html)?

IIRC there is a 2GB limit here but if the payload size is anywhere near that big then we'll hit other issues.

One thing to be careful of is that `xrslString = xrslString + "something"` can cause performance issues when the string gets long as you have to copy the string every time and the runtime becomes quadratic. It's better to use something like this:

```python
xrslStrings = []
for job in jobs:
    xrslString = ...
    xrslStrings.append(xrslString)
xrslString = "&".join(xrslString)
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Considering what we are working with, what exactly are we worried about? I mean, I can do it, but it seems unnecessary to use an external library to do something that we know exactly what to do and is simple.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Thanks. Will keep in mind for when I am able to test on the certification instance. I really have no feel about the size of the payload, but hope that it is much less than 2GB - unless of course we are doing something mad.

As a comment though, I suppose in our case it will have to simply be n copies of the same string added together. How would you suggest I do this efficiently? Will something like the following work (note the structure in https://github.com/DIRACGrid/DIRAC/pull/5542#discussion_r744189927)?

```
multipleJobs = "+(" + ")(".join(n*[xrslString]) + ")"
```
Where n is of course the number of jobs I want to submit.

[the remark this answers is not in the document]
rajanandakumar: It turns out that there is yet another bug in ARC which prevented multiple job submission. This is also being fixed (hopefully) in the next version.

I have also had a comment from the developer who does not think that submitting multiple jobs would be faster than our iterative submission of single jobs. I have asked for clarification on why.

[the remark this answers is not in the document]
rajanandakumar: And from response of the ARC developer as to "why will this *not* be faster?"
-----
It will take same amount (actually slightly more due to additional parsing of catenated job descriptions) of CPU power to process job descriptions. You can save on request/response time in case of low latency network. But as nothing prevents client form sending multiple HTTP POSTs without waiting for answer I expect the difference to be negligible. Of course it may vary due too many variables.
-----
Full thread in nordugrid mailing list in https://mail.nordugrid.org/mailman/private/nordugrid-discuss/2021q4/055051.html
-----
So, there is a question as to whether we want to do this, and how much of a benefit will DIRAC get from multiple job submission.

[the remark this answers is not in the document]
fstagni: Right. So, don't bother with bulk.

fstagni:
Just "ping" to know what's the status (e.g. of ARC release)

chrisburr:
It looks like this was fixed in v6.14.0 (December 2, 2021): https://source.coderefinery.org/nordugrid/arc/-/commit/e13a6146e7f31ece3c2287b6093f633243951ff7

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Finally confirmed that it is fixed in the newest ARC v6.14+.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+            query = self.base_url + command
+            r = self.s.post(query, headers=self.headers, params=params, timeout=1.0)
+            proxy = X509Chain.loadChainFromString(r.text)
+
+            # # We need to write the proxy out to get its information
+            # tmpProxyFile = "/tmp/arcRestRenew-" + makeGuid()[:8]
+            # with open(tmpProxyFile, 'w') as outFile: outFile.write(r.text)
+            # proxy = getProxyInfo(tmpProxyFile)
+            # os.unlink(tmpProxyFile) # Cleanup
+
```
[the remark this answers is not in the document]
rajanandakumar: Tested (offline) to work. Resolving for now as a new review will be needed anyway after my next set of commits.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+        r = self.s.post(query, data=jj, headers=self.headers, params=params, timeout=float(len(jobList)))
+        if not r.ok:
+            self.log.info("Failed getting the status of the jobs")
+            return S_ERROR("Failed getting the status of the jobs")
+
+        p = json.loads(r.text.replace("\n", ""))
+        resultDict = {}
+        jobsToRenew = []
+        jobsToCancel = []
+        for job in p["job"]:
```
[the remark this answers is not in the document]
rajanandakumar: Done. Thanks!

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+            if r.ok:  # Get the delegation and "PUT" it in the CE ...
+                dID = r.headers.get("location", "").split("new/")[-1]
+                command = "delegations/" + dID
+                query = self.base_url + command
+                r1 = self.s.put(query, data=r.text, headers=self.headers, timeout=1.0)
+                if not r1.ok:
+                    dID = ""
+            return dID
+        else:  # Retrieve delegation for existing job
+            jj = {}
```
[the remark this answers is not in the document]
rajanandakumar: Anyway replaced jj with jobsJson.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+
+        result = returnValueOrRaise(getCESiteMapping(self.ceHost))
+        self.site = result[self.ceHost]
+        self.grid = self.site.split(".")[0]
+
+        # For the REST interface
+        if self.initialised:  # a bunch of the local variables are set up here.
+            self.log.info("REST interface successfully configured", "for %s" % self.ceHost)
+        else:
+            self.log.info("REST interface not yet properly configured", "for %s" % self.ceHost)
```
fstagni, line 87: "not yet"? Is it going to be configured eventually? Otherwise it's better to stop here.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+        self.xrslExtraString = self.__getXRSLExtraString()
+        self.xrslMPExtraString = self.__getXRSLExtraString(multiprocessor=True)
+
+        # Get from the CS the time left (should be in seconds) for proxy before renewal
+        self.proxyTimeLeftBeforeRenewal = 10000  # 2 hours, 46 minutes and 40 seconds
+        variable = "proxyTimeLeftBeforeRenewal"
+        firstOption = cfgPath("Resources", "Sites", self.grid, self.site, "CEs", self.ceHost, variable)
+        secondOption = cfgPath("Resources", "Sites", self.grid, self.site, variable)
+        defaultOption = cfgPath("Resources", "Computing" "CEDefaults", variable)
+        result = self._getFromCfgSvc(firstOption, secondOption, defaultOption)
```
fstagni, line 99: Shouldn't this process of finding options already be done for you? The `self.ceParameters` dictionary should have them already, IIUC.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+
+    #############################################################################
+    @property
+    def initialised(self):
+        """Check if the REST interface is available and ready, configuring it if required.
+        Hopefully this is executed just once per CE, though it should not harm to run it more often.
+        If all is well the following variables are set, and a few others on which they (especially self.s) depend.
+        self.initialised : Static variable to say if the REST interface is available and ready
+        self.base_url: The basic URL for the REST interface
+        self.s       : The python "requests" module session
```
fstagni, line 120: Can you have a more explicit name?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+            )
+            if not status or len(response) == 0:
+                self.log.error("Bad status from LDAP search", result)
+                return False
+            if len(response) != 1:
+                self.log.warning("Expected one endpoint, got %s. Using the first." % len(response))
+            # For some reason the response is of class "bytes"
+            service_url = response[0]["raw_attributes"]["GLUE2EndpointURL"][  # pylint: disable=unsubscriptable-object
+                0
+            ].decode()
```
fstagni, line 165: Is this really the indentation you want?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+    def __init__(self, ceUniqueID):
+        """Standard constructor."""
+        super(AREXComputingElement, self).__init__(ceUniqueID)
+
+        self.ceType = CE_NAME
+        self.submittedJobs = 0
+        self.mandatoryParameters = MANDATORY_PARAMETERS
+        self.pilotProxy = ""
+        self.queue = ""
+        self.outputURL = "gsiftp://localhost"
```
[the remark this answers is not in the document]
rajanandakumar: Good point. More usefully, this is not being used anywhere that I can see. Happy to remove this variable.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Thanks - to be checked by me and updated / fixed.

It is the same algorithm also used in the ARCComputingElement.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: No - this is the indentation black wants.

If you however allow me to use more cryptic variables ... :)

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+from DIRAC.Core.Utilities.List import breakListIntoChunks
+from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
+from DIRAC.Resources.Computing.ARCComputingElement import ARCComputingElement
+from DIRAC.Core.Security.ProxyInfo import getProxyInfo
+from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
+from DIRAC.Core.Security import Locations
+from DIRAC.Core.Utilities.ReturnValues import returnValueOrRaise
+from DIRAC.ConfigurationSystem.Client.Helpers.Path import cfgPath
+
+CE_NAME = "AREX"
```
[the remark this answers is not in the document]
rajanandakumar: Oh! Do I understand that I need to just remove this line?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+        secondOption = cfgPath("Resources", "Sites", self.grid, self.site, xtraVariable)
+        defaultOption = cfgPath("Resources", "Computing" "CEDefaults", xtraVariable)
+        # Now go about getting the string in the agreed order
+        result = self._getFromCfgSvc(firstOption, secondOption, defaultOption)
+        if result["OK"]:
+            xrslExtraString = result["Value"]
+            self.log.debug("Found %s : %s" % (xtraVariable, xrslExtraString))
+        if xrslExtraString:
+            self.log.always("%s : %s" % (xtraVariable, xrslExtraString))
+            self.log.always(" --- to be added to pilots going to CE : %s" % self.ceHost)
```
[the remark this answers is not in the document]
rajanandakumar: Accepted.

But why? Is it better practice this way?

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: I think so

src/DIRAC/Resources/Computing/AREXComputingElement.py
fstagni: I think that if you remove the `# pylint: disable` it would indent it "nicely"

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: Having the "variable" part of a log message separated helps classifying the log messages.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
fstagni: BTW, gsiftp used here?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,662 @@
...
+        # base_url = "gsiftp://" + self.ceHost
+        # pilotAPI = urljoin(base_url, pilot)
+        return pilotAPI
+
+    def _pilot_toREST(self, pilot):
+        # Remove CE and protocol information from pilot ID
+        if "://" in pilot:
+            pilotREST = pilot.split("jobs/")[-1]
+            return pilotREST
+        self.log.warning("Pilot already in REST format?", "%s" % pilot)
```
fstagni, line 283:
```suggestion
        self.log.warning("Pilot already in REST format?", pilot)
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,644 @@
...
+                self.log.warn("The above message may not help. You may in that case either check")
+                self.log.warn("gocdB (goc.egi.eu) for downtime or contact the site directly.")
+                self.log.warn("Or maybe ask an expert for help ...")
+                self.log.debug("ARC job", "%s" % xrslString)
+                self.log.debug("DIRAC stamp", "%s" % diracStamp)
+                break  # Boo hoo *sniff*
+
+        if batchIDList:
+            result = S_OK(batchIDList)
+            result["PilotStampDict"] = stampDict
```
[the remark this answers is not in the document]
fstagni: It is (unfortunately) correct.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Seems to work.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: What do I do here if the init fails?

Apparently I should not do an explicit return of S_ERROR.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,668 @@
...
+        self.site = result[self.ceHost]
+        self.grid = self.site.split(".")[0]
+
+        # For the REST interface
+        if self.initialised:  # a bunch of the local variables are set up here.
+            self.log.info("REST interface successfully configured", "for %s" % self.ceHost)
+        else:
+            self.log.info("REST interface not yet properly configured", "for %s" % self.ceHost)
+            self.log.info("Try again in next iteration", "Bye Bye")
+            return S_ERROR("Try try try initialisation again")
```
chrisburr, line 86: You can't return a result from `__init__`. The only option is to raise an exception:

```suggestion
            raise RuntimeError("REST interface not yet properly configured")
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,668 @@
...
+            status, result, response, _ = connection.search(
+                "o=glue", "(GLUE2EndpointInterfaceName=org.nordugrid.arcrest)", attributes="GLUE2EndpointURL"
+            )
+            if not status or len(response) == 0:
+                self.log.error("Bad status from LDAP search", result)
+                return False
+            if len(response) != 1:
+                self.log.warning("Expected one endpoint, got %s. Using the first." % len(response))
+            # For some reason the response is of class "bytes"
+            service_url = response[0]["raw_attributes"]["GLUE2EndpointURL"][0].decode()
```
chrisburr, line 163:
```suggestion
            attributes = response[0]["raw_attributes"]  # pylint: disable=unsubscriptable-object
            service_url = attributes["GLUE2EndpointURL"][0].decode()
```

chrisburr, line 163:
```suggestion
            attributes = response[0]["raw_attributes"]  # pylint: disable=unsubscriptable-object
            service_url = attributes["GLUE2EndpointURL"][0].decode()
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,668 @@
...
+        Hopefully this is executed just once per CE, though it should not harm to run it more often.
+        If all is well the following variables are set, and a few others on which they (especially self.session)
+        depend.
+        self.initialised : Static variable to say if the REST interface is available and ready
+        self.base_url : The basic URL for the REST interface
+        self.session  : The python "requests" module session
+        self.headers  : The format of information we prefer : json of course (except during delegations)
+        Specification : https://www.nordugrid.org/arc/arc6/tech/rest/rest.html
+        """
+        if hasattr(self, "s"):
```
chrisburr, line 124: I guess this should be:
```suggestion
        if hasattr(self, "session"):
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Fixed by splitting the line.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Yes - this is kind of the solution I went for.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Thanks!

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,668 @@
...
+            service_url = response[0]["raw_attributes"]["GLUE2EndpointURL"][0].decode()
+
+        # We now have a pointer (service_url) to the REST interface for this CE
+        restVersion = "1.0"  # Will be this value for the forseeable future.
+        self.base_url = service_url + "/rest/" + restVersion + "/"
+
+        # Set up the request framework
+        self.session = requests.Session()
+        self.session.verify = Locations.getCAsLocation()
+        self.headers = {"accept": "application/json", "Content-Type": "application/json"}
```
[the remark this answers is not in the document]
rajanandakumar: Done - thanks.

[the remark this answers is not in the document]
rajanandakumar: One possible point is that the "self.session" is now defined in the function "initialised" instead of "__init__". 

From formally pure python, I am guessing this is not very elegant. Any thoughts as to how I can tell that the initialisation has successfully run without looking for "session" in "self"?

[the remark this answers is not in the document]
fstagni: `__init__` will only be called once.

marianne013:
As Federico noted, there should be a queue specific XRSLExtraString. Otherwise we cannot use e.g. high memory queues, as we have to set it by CE and if the CE has more than one queue, it gets really really messy.

rajanandakumar:
> As Federico noted, there should be a queue specific XRSLExtraString. Otherwise we cannot use e.g. high memory queues, as we have to set it by CE and if the CE has more than one queue, it gets really really messy.

Noted - thanks.

aldbr:
A few fixes from the ARC/token hackathon that took place in September: https://github.com/rajanandakumar/DIRAC/pull/2

rajanandakumar:
> A few fixes from the ARC/token hackathon that took place in September: [rajanandakumar#2](https://github.com/rajanandakumar/DIRAC/pull/2)

Hi @aldbr 

Thanks for this. I will get to it next as soon as I am done with the ceParameters which is working now for me.

Cheers,
Raja.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -66,86 +67,98 @@ class AREXComputingElement(ARCComputingElement):
...
+            self.proxyTimeLeftBeforeRenewal = self.ceParameters[variable]
+        self.log.debug(f"Setting {variable} to {self.proxyTimeLeftBeforeRenewal} for CE", self.ceHost )
+
+        variable = "ARCRESTTimeout"
+        self.arcRESTTimeout = 1.0
+        if variable in self.ceParameters.keys():
+            self.arcRESTTimeout = self.ceParameters[variable]
+        self.log.debug(f"Setting {variable} to {self.arcRESTTimeout} for CE", self.ceHost )
+
+        variable = "XRSLExtraString"
```
aldbr: I guess it would be better to handle `XRSLExtraString` and `XRSLMPExtraString` in `ARCComputingElement`.
`self.ceParameters` is generally used in `_reset()`.
Renaming the function as  `_reset()` should help `ComputingElement` to initialize the CE properly.
This is what we do in the other CEs. 
Please take a look at my PR, this is implemented.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -66,86 +67,98 @@ class AREXComputingElement(ARCComputingElement):
...
         Hopefully this is executed just once per CE, though it should not harm to run it more often.
-        If all is well the following variables are set, and a few others on which they (especially self.session)
-        depend.
+        If all is well the following variables are set, and a few others on which they (especially self.s) depend.
         self.initialised : Static variable to say if the REST interface is available and ready
-        self.base_url : The basic URL for the REST interface
-        self.session  : The python "requests" module session
-        self.headers  : The format of information we prefer : json of course (except during delegations)
+        self.base_url: The basic URL for the REST interface
+        self.s       : The python "requests" module session
```
aldbr: Why did you decide to rename `session` as `s`?
`session` was much more explicit IMHO.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -511,41 +462,37 @@ def getCEStatus(self):
...
 
             # Get the proxy
             command = "delegations/" + dID
             params = {"action": "get"}
             query = self.base_url + command
-            r = self.session.post(query, headers=self.headers, params=params, timeout=self.arcRESTTimeout)
-            proxy = X509Chain()
-            proxy.loadChainFromString(r.text)
+            r = self.s.post(query, headers=self.headers, params=params, timeout=self.arcRESTTimeout)
+            proxy = X509Chain.loadChainFromString(r.text)
```
aldbr: Are you sure about this line?
`loadChainFromString` does not seem to be a class method + it seems to return `S_OK()`, not a proxy.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -511,41 +462,37 @@ def getCEStatus(self):
...
         This function is called only by this class, NOT by the SiteDirector"""
-
-        # Get the new proxy once for all the jobs ...
-        newProxy = X509Chain()
-        res = newProxy.loadProxyFromFile(self.session.cert)
-
-        # Renew the jobs
         for job in jobList:
             # First get the delegation (proxy)
             dID = self._delegation(job)
```
aldbr: Have you managed to get a delegation ID from a Job ID?
I received `Action not implemented` when I tried.

From what I understand, a delegation ID is bound to many jobs, am I wrong?
If it is the case, it would be probably more efficient to get a list of delegation IDs first, and then work with them instead of the job IDs to renew the proxies.
What do you think?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -511,41 +462,37 @@ def getCEStatus(self):
...
+            # tmpProxyFile = "/tmp/arcRestRenew-" + makeGuid()[:8]
+            # with open(tmpProxyFile, 'w') as outFile: outFile.write(r.text)
+            # proxy = getProxyInfo(tmpProxyFile)
+            # os.unlink(tmpProxyFile) # Cleanup
 
             # Now test and renew the proxy
-            if not proxy["OK"]:
-                continue
-            timeLeft = proxy.getRemainingSecs()
+            if not proxy["OK"] or "secondsLeft" not in proxy["Value"]:
```
aldbr: Also, I wonder whether it is really necessary to make a request to get the proxy from the delegation ID + check whether it should be renewed + renew it if needed VS make a request to renew the proxy regardless of the time left.
What do you think?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -486,13 +437,13 @@ def getCEStatus(self):
...
         query = self.base_url + command
-        r = self.session.get(query, headers=self.headers, params=params, timeout=5.0 * self.arcRESTTimeout)
+        r = self.s.get(query, headers=self.headers, params=params, timeout=5.0 * self.arcRESTTimeout)
 
         if not r.ok:
             res = S_ERROR("Unknown failure for CE %s. Is the CE down?" % self.ceHost)
             return res
 
-        p = r.json()
+        p = json.loads(r.text)
```
aldbr: Why did you change that?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -323,33 +278,32 @@ def _delegation(self, jobID):
...
+            jj = {"job": [{"id": jobID}]}  # job in ARC REST json format
             command = "jobs"
             params = {"action": "delegations"}
             query = self.base_url + command
-            r = self.session.post(query, data=json.dumps(jobsJson), headers=self.headers, timeout=self.arcRESTTimeout)
+            r = self.s.post(query, data=json.dumps(jj), headers=self.headers, timeout=self.arcRESTTimeout)
             dID = ""
             if r.ok:  # Check if the job has a delegation
-                p = r.json()
+                p = json.loads(r.text.replace("\n", ""))
```
aldbr: Why is this better than `r.json()`?

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: I did not want to type in "session" each time. "s" in this context seemed clear enough.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: I believe it was following a suggestion in one of the earlier reviews. As I say in the comment just below that, it is to be tested.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Hi @aldbr 

About the delegations, which CE did you test against? For me, it worked with lancaster

ce = "grendel.hec.lancs.ac.uk"

However, I did not have luck with any of the RAL CEs. So, it depends on the version of the ARC CE used I suppose.

About the job renewal, I am not sure I understand. Is your suggestion to simply renew all the delegations at the given CE? This may mean renewing delegations without any jobs, which are best left to expire ...

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Hmmm - good point. It does sound logical, but can I ask @sfayer  or @fstagni to also comment on this please. We will save 1 REST api query.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: What are you comparing with? I don't think I had r.json anywhere until now?

If it is with your version, I will get to it after going over your comments here!

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Old history - until we got the ARC developers to fix issues with their response strings, the replace was essential to prevent crashes (see earlier thread of my discussion with Chris Burr and bug report on nordugrid bugzilla).

src/DIRAC/Resources/Computing/AREXComputingElement.py
chrisburr: You should use `session` for the name.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,625 @@
+########################################################################
+# File :   Originally ARCComputingElement.py
+# Author : Raja Nandakumar
+# Use the ARC REST interface and retire api libraries when complete
+########################################################################
+
```
chrisburr: This is all duplicated with the git history and module docstring:
```suggestion
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,625 @@
...
+    A lot of the features are common with the API interface. In particular, the XRSL
+    language is used in both cases. So, we retain the xrslExtraString and xrslMPExtraString strings.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+__RCSID__ = "$Id$"
+
+import six
```
chrisburr: `six` belongs in the non-standard library imports and each list should be sorted alphabetically, i.e.
```python
import os
import stat
import sys

import json
import ldap3
import requests
import six
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,625 @@
...
+import os
+import stat
+import sys
+
+import requests
+import json
+import ldap3
+
+# from urllib.parse import urljoin
+
```
chrisburr:
```suggestion
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
aldbr: > About the job renewal, I am not sure I understand. Is your suggestion to simply renew all the delegations at the given CE? This may mean renewing delegations without any jobs, which are best left to expire ...

Let's say jobs A and B are submitted at the same time, and C is submitted later. We would have:
- delegation1 is bound to jobs A, B
- delegation2 is bound to job C

Your method does:
- get delegation of A: delegation1; renew it
- get delegation of B: delegation1; renew it
- get delegation of C: delegation2; renew it

In this context, delegation1 is renewed twice.

Whereas you could first make a list of delegations. You would have:
- get delegation of A: delegation1
- get delegation of B: delegation1
- get delegation of C: delegation2
- list of delegations [delegation1, delegation2]
- renew delegation1, delegation2

Do I understand correctly the code?
What do you think?

src/DIRAC/Resources/Computing/AREXComputingElement.py
aldbr: Well, you wrote `r.json()` in the previous commit and then you changed that for `json.loads(r.text)`, I didn't change anything here.

src/DIRAC/Resources/Computing/AREXComputingElement.py
aldbr: Ok, could you add a comment about that above this line then, please?

chrisburr: This bug was fixed so it should be using `r.json()`.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,545 @@
+""" AREX Computing Element (ARC REST interface)
+    Using the REST interface now and fail if REST interface is not available.
+    A lot of the features are common with the API interface. In particular, the XRSL
+    language is used in both cases. So, we retain the xrslExtraString and xrslMPExtraString strings.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
fstagni, line 9: I think this could be coded directly in py3 (only, so remove also the `import six` below) as I don't see the point of using this within a py2 installation. It will also avoid further modifications in later branches.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,545 @@
...
+    def __init__(self, ceUniqueID):
+        """Standard constructor."""
+        super(AREXComputingElement, self).__init__(ceUniqueID)
+
+        # Default REST port
+        self.port = "443"
+        # REST version to adopt
+        self.restVersion = "1.0"
+        # Time left before proxy renewal: 2 hours, 46 minutes and 40 seconds
+        self.proxyTimeLeftBeforeRenewal = 10000
```
fstagni: What is the rational behind this number?

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,545 @@
...
+        else:
+            pilotRef = jobID
+            stamp = ""
+        if not stamp:
+            return S_ERROR("Pilot stamp not defined for %s" % pilotRef)
+
+        # Prepare the command
+        command = "jobs/"
+        job = self._DiracToArcID(pilotRef)
+        query = self.base_url + command + job + "/session/" + stamp + ".out"
```
fstagni: I find these string manipulations inconvenient: there will always be the case when a double `/` or a double `.` is inserted. There are better libraries to avoid that, care to use them?

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: It is a round number of seconds :)

The aim is that a period of around 3 hours of proxy remaining time was considered a good amount during a discussion when this feature was originally requested by a (non-LHC) VO.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: I chose this specifically, as these strings are definite here. So, we any know exactly what we are doing and can avoid loading yet another library.

If it is a strong requirement, it can be done.

fstagni: > we any know exactly what we are doing

I like your positivity!

Loading another library is OK, anyway mine is not a strong requirement.

rajanandakumar: Okay :smile:  - can be done in the next update. I am sure there will be a requirement after we test it against CSCS in LHCb.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Superseded by the changes made by @aldbr

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: PR accepted. Resolving.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done in the patch suggested by @aldbr

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Good point. I was trying to avoid having two loops. This also will work I agree - when you have some time, will you be able to provide a patch, or shall I do it?

It is not something that is going to happen often. So, I am happy for testing this module to go ahead with the code as it exists and update this in the next iteration.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Hmmm - git is not a strong point for me, but it does not matter. I was anyway planning to move everything to r.json() later.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Yes - agreed.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Oh - okay. I believe this is done with the PR of @aldbr .

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Will do as I mentioned to @fstagni a short while ago (at another point)

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: This comment is beyond me :smile: . @aldbr can I hand over this comment to you please as the lines were added by you if I am correct.

src/DIRAC/Resources/Computing/AREXComputingElement.py
fstagni: 10800 (3 hours) is not too bad 😉

rajanandakumar: Will do in next round. Happy for this version to be tested as requested by Vladimir!

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done - I had to write an internal function, as I did not find anything appropriate in the python libraries.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Turns out, this will join at most one part to a URL.

With my prehistoric programming ability, I decided to write an internal function to do this.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Done.

rajanandakumar:
I note that there are changes to .pre-commit-config.yaml and to pyproject.toml which I needed to get the black working properly.

pyproject.toml
```diff
@@ -10,4 +10,4 @@ git_describe_command = "git describe --dirty --tags --long --match *[0-9].[0-9]*
 
 [tool.black]
 line-length = 120
-target-version = ['py27']
+target-version = ['py36']
```
chrisburr: This definitely shouldn't be changed

chrisburr (changes requested):

pyproject.toml
rajanandakumar: Hi @chrisburr 

Just curious, why? :)

More seriously, black (v22.8) on my machine (arch linux) refused to work with the given target version of py27. And once I made the change to get black working, git refused to commit my code, unless I also committed pyproject.toml also.

So, I was left with a catch-22 situation. I am happy for you to ignore the commit to this file if this change is such a breaking change (is it?).

Cheers,
Raja.

fstagni: Well as you can see the tests on Github Actions did not run, so yes this is breaking.

chrisburr: > Just curious, why? :)

Some DIRAC installations are still using Python 2.7 and using a newer black version causes the code to be reformatted in a way which causes a `SyntaxError` to be raised. This is why we still use the old `black` version in `rel-v7r2`/`rel-v7r3`. It's already been updated to be `py39` for `rel-v8r0`/`integration`.

rajanandakumar: Hi @fstagni and @chrisburr ,

Thanks for the explanations. With the latest two commits, I hope I have reset the two files - pyproject and pre-commit-config. The AREXComputingElement remains unchanged.

Let me know if there is something else that is going wrong here, as I still do not see all the checks running.

Edit : Fixing the pre-commit config on the web seems to have resolved the issue with the checks.

Cheers,
Raja.

rajanandakumar: And on my machine, I am not able to downgrade black to the version required as the relevant gpg signatures for the package have expired apparently.

.pre-commit-config.yaml
```diff
@@ -22,4 +22,4 @@ repos:
     rev: 21.9b0
     hooks:
       - id: black
-        additional_dependencies: [".[python2]", "click==8.0.4"]
+        additional_dependencies: [".[python2]"]
```
chrisburr: Please revert this as well

rajanandakumar: Apologies - I am not sure I understand. Do we want click or not?

In the original file I had, click was not there...

chrisburr: We want this file to be untouched by this pull request. This branch was originally made a year ago and have been changes to `rel-v7r2` since then (in this case to workaround an issue with the Python 2 compatible `black` release no longer working with newer `click` releases).

Would you like me to rebase this branch for you and squash all the commits into one?

rajanandakumar: Yes please. In hindsight, this is the fastest!

chrisburr: Done :+1:

You'll have conflicts when you try to pull locally, you can get the latest version with:

```bash
git fetch --all
# DANGER: This is destructive, make sure you're on the right branch
# and you have no local changes
git reset --hard FETCH_HEAD
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,632 @@
...
+        :return: ARC jobID
+        """
+        # Remove CE and protocol information from pilot ID
+        if "://" in diracJobID:
+            arcJobID = diracJobID.split("arex/")[-1]
+            return arcJobID
+        self.log.warn("Identifier already in REST format?", diracJobID)
+        return diracJobID
+
+    def _UrlJoin(self, words):
```
aldbr, line 126: I don't know if this was really needed but it doesn't hurt.

On the one hand, as you said in a previous comment, we control most of the content of the url.
On the other hand, I can understand that building a long url (like in `getJobOutput()`) can be source of errors.
We could also use the `urllib.parse` library to build the url in an automated way: https://docs.python.org/3.9/library/urllib.parse.html

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,632 @@
...
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("AREXComputingElement: failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        if type(jobIDList) != type([]):
+            jobTmpList = [jobIDList]
```
aldbr:
```suggestion
        if not isinstance(jobIDList, list):
            jobIDList = [jobIDList]
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -0,0 +1,632 @@
...
+        # Bulk submission would not be significantly faster than multiple single submission.
+        # https://www.nordugrid.org/arc/arc6/tech/rest/rest.html#job-submission-create-a-new-job
+        # Also : https://bugzilla.nordugrid.org/show_bug.cgi?id=4069
+        batchIDList = []
+        stampDict = {}
+        for _ in range(numberOfJobs):
+            # Get the job into the ARC way
+            xrslString, diracStamp = self._writeXRSL(executableFile)
+            if len(delegation) > 5:
+                xrslString = xrslString + delegation
```
aldbr:
```suggestion
            xrslString += delegation
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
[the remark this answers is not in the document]
rajanandakumar: Apologies for missing this question until now.

No - gsiftp is not used here. It is used in a dummy fashion, only to get the pilot into the "old-style DIRAC" format. The protocol is chosen as the same protocol previously used in the ARC API, primarily so that a shifter looking at the pilot monitoring will not be too confused.

Resolving this conversation as the request is implemented in the code.

src/DIRAC/Resources/Computing/AREXComputingElement.py
rajanandakumar: Indeed, we control the content. However, it has been a repeated request and so has been implemented this way.

I did take a close look at urlparse. However, it can only join exactly two strings. There is also urlunsplit which does something very different with exactly 6 arguments. And os.path.join does something very different.

So, after googling a bit, I found that writing a function gave me the best solution. However, if you can implement a similar feature with a standard library, I am happy to use it!

.pre-commit-config.yaml
rajanandakumar: Thanks a lot! :+1:

fstagni:
I pushed some minor fixes, now it is ready I think. If the checks don't fail I will merge.

PR #5578, opened 2021-11-08
Fix cloudinit

atsareg:
BEGINRELEASENOTES

*Resources
FIX: cloudinit-template - allow time in the monitor for the pilot to start 
FIX: cloudinit-template - set DIRACSYSCONFIG to point to the pilot.cfg in order to allow its use in the user applications

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Cloud/cloudinit.template
+++ b/src/DIRAC/Resources/Cloud/cloudinit.template
@@ -143,6 +143,7 @@ write_files:
                      -o /Cloud/%(running-pod)s/JobWrappersLocation=/mnt/dirac \
                      --UseServerCertificate
      mkdir -p runit/VirtualMachineMonitorAgent/log
+     sleep 180
      exec dirac-agent WorkloadManagement/VirtualMachineMonitorAgent
 
 yum_repos:
--- a/src/DIRAC/Resources/Cloud/cloudinit.template
+++ b/src/DIRAC/Resources/Cloud/cloudinit.template
@@ -106,6 +106,7 @@ write_files:
             --name "%(dirac-site)s" \
             -Q "%(vmtype)s" \
             --cert \
+            --userEnvVariables DIRACSYSCONFIG:::/mnt/dirac/pilot.cfg \
             -o /LocalSite/VMID=%(vm-uuid)s \
             -o /LocalSite/LocalCE=%(ce-type)s \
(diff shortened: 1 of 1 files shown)


fstagni (approved):

sfayer:
I have a couple of questions:
 * Do you see problems where the monitor is starting up too quickly and stopping the node? It should have a built-in wait before it does anything, so perhaps we could modify that to be slower instead?
 * (Possibly more a question for @fstagni) We seem to be adding that environment variable for the pilot.cfg in lots of places: Perhaps the pilot could do it by default instead, or does that break something else?
 * I thought were were making a dirac.cfg to pilot.cfg symlink somewhere for the cloud pilots anyway, is that not sufficient in this case?

atsareg:
@sfayer : 
1. Indeed I saw problems when the monitor starts before the JobAgent and halts the VM because of that. This was a simple fix. In the monitor agent there is a possibility to slow down the reaction which is defined by "running pod" in the configuration. Are you actually using running pods ? We should probably scratch this concept and define these parameters in the /Operations/.../Cloud section directly.   
3. As far as I can see the link etc/dirac.cfg -> pilot.cfg is defined only when using the SingularityComputingElement inner CE. I am using PoolCE/InProcess CE. But even when using the SingularityComputingElement the way the link is created seems not to be compatible with Python3 pilots. To be checked.

sfayer:
>     1. Indeed I saw problems when the monitor starts before the JobAgent and halts the VM because of that. This was a simple fix. In the monitor agent there is a possibility to slow down the reaction which is defined by "running pod" in the configuration. Are you actually using running pods ? We should probably scratch this concept and define these parameters in the /Operations/.../Cloud section directly.

No, we don't use (Running)Pods at all: I think I removed a lot of the pod code when we merged VMDIRAC as it was mostly broken anyway. I'm happy for that to be removed and merged into the main config section.

>     2. As far as I can see the link etc/dirac.cfg -> pilot.cfg is defined only when using the SingularityComputingElement inner CE. I am using PoolCE/InProcess CE. But even when using the SingularityComputingElement the way the link is created seems not to be compatible with Python3 pilots. To be checked.

Yes, it's certainly possible it was in SingularityCE and I'm misremembering... Perhaps we could make a symlink in the cloudinit.template so that it works for all InnerCE types? We'd still need to find out how to make it compatible with python3 pilots.

atsareg:
I will address the delay in the monitor start in the next PR where the RunningPods will be cleaned away

PR #5606, opened 2021-11-17
[v7r3] Return error if JobPolicy.getControlledUsers returns an empty list

chrisburr:
⚠️ I'm not certain this is the right solution but somehow we have to prevent it going further and trying to do ``WHERE (`Owner`, `OwnerGroup`) IN (  ) AND``:

```
ERROR  _query (SELECT `Status`, COUNT(*) FROM `Jobs`    WHERE (`Owner`, `OwnerGroup`) IN (  ) AND `LastUpdateTime` < "2021-11-17 13:11" GROUP BY `Status...
```

BEGINRELEASENOTES

*WorkloadManagement
FIX: Return error if JobPolicy.getControlledUsers returns an empty list

ENDRELEASENOTES


fstagni (approved):
I think it's fine. Alternatively `JobPolicy.getControlledUsers` could be changed, but here it's more correct.

PR #5623, opened 2021-11-19
[integration] do not use dirac-proxy-init in dirac-configure

TaykYoku:
The creation of this PR was inspired by [discussion](https://github.com/DIRACGrid/DIRAC/pull/5397#discussion_r752345622).

It looks like `dirac-proxy-init` contains special flags for use inside `dirac-configure` command. IMHO, this algorithm can be simplified by using the `X509Chain` class at once. There is a case when certificates in a standard place are not found, this change will allow the user at desire to define a non-standard place of the certificate.

BEGINRELEASENOTES

*Core
CHANGE: dirac-configure: use X509Chain instead of dirac-proxy-init, allow user to define a non-standard place of the certificate

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/scripts/dirac_configure.py
+++ b/src/DIRAC/Core/scripts/dirac_configure.py
@@ -47,14 +47,25 @@ from __future__ import division
 from __future__ import print_function
 import sys
 import os
+from prompt_toolkit import prompt, print_formatted_text, HTML
+from prompt_toolkit.completion import FuzzyWordCompleter
 
 import DIRAC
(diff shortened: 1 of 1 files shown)


chaen:
I don't think it is a good idea to duplicate the logic. but let's continue the [discussion](https://github.com/DIRACGrid/DIRAC/pull/5397#discussion_r752345622).

chrisburr:
Closing as agreed in the BiLD.

At some point this will need to be modified to use `dirac-login` instead of `dirac-proxy-init` but we need to finalise the details of `dirac-login` first.

PR #5638, opened 2021-11-26
[v7r2] More executor fixes

chrisburr:
BEGINRELEASENOTES

*Core
FIX: Tasks never get removed from the ExecutorState if an UnrecoverableTaskException is raised

ENDRELEASENOTES


chrisburr:
This has been running in LHCb's production setup for several days and seems to have fixed the issues with the optimisers getting completely stuck.

There is still a bunch of race conditions that cause jobs to get stuck in "Recieved" until the mind is restarted but I don't have the time to look into those any further.

PR #5710, opened 2021-12-08
[integration] feat (core): MultiVO shifter enhancements

martynia:
BEGINRELEASENOTES

*Core
NEW: A multi VO shifter enhancement when used by agents. Functions in `shifter.py` get a new argument, `vo`, which defaults to False and is used by `Operations()` constructor to access a VO-specific section in the CS. The feature is activated by an agent defining a `self.vo `attribute.  

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Base/AgentModule.py
+++ b/src/DIRAC/Core/Base/AgentModule.py
@@ -351,7 +351,10 @@ class AgentModule(object):
 
     def _setShifterProxy(self):
         if self.__moduleProperties["shifterProxy"]:
-            result = setupShifterProxyInEnv(self.__moduleProperties["shifterProxy"], self.am_getShifterProxyLocation())
+            vo = getattr(self, "vo", False)
+            result = setupShifterProxyInEnv(
+                self.__moduleProperties["shifterProxy"], self.am_getShifterProxyLocation(), vo=vo
+            )
             if not result["OK"]:
                 self.log.error("Failed to set shifter proxy", result["Message"])
                 return result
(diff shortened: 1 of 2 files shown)


chaen:
In practice, that means having 1 agent per VO. I don't care, but isn't it exactly what was deemed as uncool for services ? 

Also, I am not sure to understand how you expect `self.vo` to be defined for an agent ?

martynia:
The similar way is is defined by a SiteDirector. The derived class will have to define it, otherwise you will get a VO-less shifter, as before.
Cheers, JM

martynia:
We have a use case for this. In a multi-VO environment we would have an agent dealing with pilot logs for all jobs/VOs but using a VO specific shifter since we have to upload the logs to an SE, which is VO-aware. Does it make sense ?
JM

andresailer:
I agree that it is really uncool to have one agent per VO, and ideally the VO should be figured out from some other information, and a required proxy should be created then, similar to what I have done for the TransformationSystem agents.

 * You can achieve the same result by having VO specific shifters in the Default section of Operations/Shifters. Shifters are defined by user and group, so they are not VO-less per se. (adding another way would require updates to the documentation in this PR beyond the release notes.)

* If you anyway have to specify the VO for the agent, maybe it would be more convenient to define the user and group in the agent configuration and not using the shifterProxy, or making the agent configurable in a way where it can get a list of VO+user+group tuples to deal with multiple VOs

Then using the `executeWithUserProxy` function
https://github.com/DIRACGrid/DIRAC/blob/0b28e274f8a0e5f2dce7467277469a58476558da/src/DIRAC/TransformationSystem/Agent/TransformationCleaningAgent.py#L156-L158
(also as a decorator) or  the `UserProxy` context manager
https://github.com/DIRACGrid/DIRAC/blob/0ebab9f0b93212a5ac42cb72a6982b958708ba98/src/DIRAC/TransformationSystem/Utilities/TransformationInfo.py#L149

To run the functionality that requires the proxy.
See also https://dirac.readthedocs.io/en/latest/CodeDocumentation/Core/Utilities/Proxy.html

PR #5713, opened 2021-12-09
[integration] use DIRACScript as Script

TaykYoku:
No need to use both `DIRACScript` and `Script`(should deprecated). Noticed looking at #5694.

BEGINRELEASENOTES

*Core
CHANGE: add deprecation message for Script module

*WMS
FIX: use DIRACScript as Script

ENDRELEASENOTES


chaen:
Is `DIRACScript` totally backward compatible with `Script` ? In which case why not changing the code of `Script` for the one of `DIRACScript` ? it would be less of a burden for everybody

TaykYoku:
> Is DIRACScript totally backward compatible with Script ?

yes

> In which case why not changing the code of Script for the one of DIRACScript ?

Didn't quite understand do you want to move `DIRAC.Core.Utilities.DIRACScript.DIRACScript` -> `DIRAC.Core.Base.Script.Script`?

src/DIRAC/Core/Base/Script.py
```diff
@@ -55,6 +56,9 @@
 gIsAlreadyInitialized = False
 
 
+@deprecated(
+    "Do\nfrom DIRAC.Core.Utilities.DIRACScript import DIRACScript as Script\ninstead of\nfrom DIRAC.Core.Base import Script"
+)
 def parseCommandLine(script=False, ignoreErrors=False, initializeMonitor=False):
```
chrisburr, line 62: How are we supposed to use DIRAC interactively in Python if we remove this? There is no equivalent to:

```python
parseCommandLine()
```

I'm very much in favour of having something more findable but `DIRACScript` isn't for use outside of DIRAC+extensions.

TaykYoku: I shared another method as we agreed at the last rally. Please, to use DIRAC interactively, use:
```python
from DIRAC import initialize
initialize()
```

andresailer:
I think there needs to be a lot better motivation to make every user change their python scripts just to replace one magic method (`parseCommandLine`) with another (`initialize`).

src/DIRAC/Core/Base/Script.py
```diff
@@ -48,6 +48,11 @@
 gIsAlreadySetUsageMsg = False
 gIsAlreadyInitialized = False
 
+gLogger.warn(
```
andresailer, line 51: is `warn` now printed in scripts by default?
This is printed at import time, and maybe from somewhere that the user doesn't control.
It is entirely unclear what "custom configuration" is supposed to mean. `parseCommandLine` also loads the global configuration.

TaykYoku: Yes, better to use decorator, fixed.

TaykYoku:
>I think there needs to be a lot better motivation to make every user change their python scripts just to replace one magic method (parseCommandLine) with another (initialize).

If we are talking about scripts, when you should not use `initialize`, keep working with `Script.parseCommandLine`. `initialize` is useful when you use DIRAC interactively in Python.

andresailer:
I am talking about users writing their own little scripts to do whatever, not "commands" that are part of DIRAC or extensions.

TaykYoku:
>I am talking about users writing their own little scripts to do whatever, not "commands" that are part of DIRAC or extensions.

Well, I understand you don't want to bother users about updating their scripts. Honestly, it's hard for me to estimate the scale of this case and how difficult it will be for users to update their scripts.
This change just tell about obsolescence of use `from DIRAC.Core.Base.Script import parseCommandLine`. And these methods will continue to work. We can delete them later at any time.

TaykYoku:
Well, I tried to take into account the latest wishes and maintain backward compatibility without duplicating the code. And left the changes approved at the build meeting.
Now it can be reviewed/merged it, I hope

TaykYoku:
To be clear, what was done:
- I move `DIRACScript` logic to `DIRAC.Core.Base.Script`
- I remove all deprecation messages

Now users can continue to use:
```
Script.parseCommandLine()
```
OR
```
parseCommandLine()
```

Commands created with DIRACScript Decorator (as described in previous release instructions) will also work(do I need to change this in the instructions and mark DIRACScript as obsolete?):
```
from DIRAC.Core.Utilities.DIRACScript import DIRACScript as Script

@Script
def main(...
```

Recommended way:
- for commands use:
```

@Script
def main(...
```
- in other cases:
```
import DIRAC
DIRAC.initialize()  # load configuration
```

Are there any additional comments or opinions? Can it be merged?

src/DIRAC/ConfigurationSystem/Client/LocalConfiguration.py
```diff
@@ -386,7 +372,6 @@ def __parseCommandLine(self):
         for opt, val in opts:
             if opt in ("-h", "--help"):
                 self.showHelp()
-                sys.exit(2)
```
chrisburr, line 389: Why? Running `dirac-something --help` shouldn't ever do anything (though I think it should probably exit with `0`).

src/DIRAC/Core/Base/__init__.py
```diff
@@ -0,0 +1,2 @@
+""" For backward compatability """
+from DIRAC.Core.Base.Script import Script
```
chrisburr, line 2: I struggle to see what this is doing?

src/DIRAC/Core/Utilities/DIRACScript.py
```diff
@@ -1,185 +1,2 @@
...
-    def disableCS(cls):
-        """See :func:`~DIRAC.ConfigurationSystem.Client.LocalConfiguration.LocalConfiguration.disableCS`."""
-        cls.localCfg.disableCS()
-
-    @classmethod
-    def enableCS(cls):
-        """See :func:`~DIRAC.ConfigurationSystem.Client.LocalConfiguration.LocalConfiguration.enableCS`."""
-        return cls.localCfg.enableCS()
+""" For backward compatability """
+from DIRAC.Core.Base.Script import Script as DIRACScript
```
chrisburr, line 2: Importing this should show a deprecation message

src/DIRAC/TransformationSystem/Client/TransformationCLI.py
```diff
@@ -1,13 +1,8 @@
...
+import DIRAC
 
-from __future__ import print_function
-from __future__ import absolute_import
-from __future__ import division
-
-from DIRAC.Core.Base.Script import parseCommandLine
-
-parseCommandLine()
+DIRAC.initialize()  # Initialize configuration
```
chrisburr, line 5: I realise this was a pre-existing issue but can this be refactored to make it safe to import?

src/DIRAC/__init__.py
```diff
@@ -153,6 +153,22 @@ def _computeRootPath(rootPath):
 from DIRAC.ConfigurationSystem.Client.Config import gConfig
 
 
+def initialize():
```
chrisburr, line 156: I think there are some arguments that would be useful for this. Perhaps:

```suggestion
def initialize(*, log_level=None, auth_required=True, cert=None, key=None):
```

(The `(*, ` makes them [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/).)

chrisburr: Also I'd be tempted to call this `init` rather than `initialize`, but I might be biased by the presence of the `z` 😉

src/DIRAC/ConfigurationSystem/Client/LocalConfiguration.py
TaykYoku: `showHelp` will do DIRAC.exit()

src/DIRAC/Core/Base/__init__.py
TaykYoku: you can do:
```python
Script.parseCommandLine()
```

src/DIRAC/Core/Utilities/DIRACScript.py
TaykYoku: Ill create separate PR with it

TaykYoku: #5870

src/DIRAC/__init__.py
TaykYoku: >I think there are some arguments that would be useful for this

I personally do not think it is worth burdening this method with arguments, at least for now. But it may make sense whether you could do the appropriate PR or issue, if you have an idea of what should be there.

PR #5755, opened 2021-12-17
[v7r3] Allow to specify input protocol in the CS 

chaen:
There are cases when specific storage can input/output more protocols with a given plugin than other storages (e.g. some tape are SRM+HTTPs ready, some aren't, `dCache` can return `xroot` instead of `root`, `Storm` can return `file`, etc). 
This PR allows to overwrite the hardcoded default with a CS configuration. This was meant to be possible from the beginning when I introduced multi protocol, but somehow, it did not make it into the code. Now, it's here

(triggered by https://github.com/DIRACGrid/DIRAC/issues/5727)

BEGINRELEASENOTES
*Resources
NEW: allow to overwrite input/output  storage plugin parameter with the CS
CHANGE: add "xroot" to the list of input protocol for the XROOT plugin

ENDRELEASENOTES

### DIFF
--- a/dirac.cfg
+++ b/dirac.cfg
@@ -548,6 +548,8 @@ Resources
         Access = remote
         SpaceToken = LHCb-EOS
         WSUrl = /srm/v2/server?SFN=
+        InputProtocols = file, https, root, srm, gsiftp # Allow to overwrite the list of protocols understood as input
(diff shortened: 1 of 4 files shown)


sfayer (approved):
Looks fine to me.

PR #5771, opened 2022-01-10
[7.3] add tornado IO loop handling to ThreadScheduler

TaykYoku:
As far as I understand we are moving from **DISET** to **Tornado**, but the services (and not only) are dependent(e.g.: `SecurityLogClient`) on:
```
from DIRAC.Core.Utilities.ThreadScheduler import gThreadScheduler
```

During the transition period both **DISET**/**TornadoService** will use `gThreadScheduler`.

The goal of PR will make `gThreadScheduler` universal for both cases and use tornado IO loop if `DIRAC_USE_TORNADO_IOLOOP` is set.

- [ ] testing

BEGINRELEASENOTES

*Core
CHANGE: add tornado IO loop handling to ThreadScheduler

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/ThreadScheduler.py
+++ b/src/DIRAC/Core/Utilities/ThreadScheduler.py
@@ -10,6 +10,11 @@ import hashlib
 import threading
 import time
 
+# For IOLoopScheduler
+import os
+import tornado.ioloop
+from functools import partial
+
 from DIRAC import S_ERROR, S_OK, gLogger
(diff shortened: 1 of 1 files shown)


chaen:
The only Client making use of the `ThreadSceduler` is the `SecurityLog` system, that we want to remove (https://github.com/DIRACGrid/DIRAC/pull/5760), so I wouldn't bother doing that really.

TaykYoku:
This PR is considered unnecessary based on the comment above.

PR #5796, opened 2022-01-13
[integration] [hackathon] fix(transformation): convert task/file ID to integer to fix type mismatch

andresailer:
The fix in TaksManager fixes the original problem discovered during the hackathon, the fix in transformationDB is a followup issue.

I guess this comes from the JSON encode/decode where integer keys are no longer supported?

BEGINRELEASENOTES

*TS
FIX: Convert fileID / taskID to fix issue when task status is updated in transformation treatment

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/TransformationSystem/Client/TaskManager.py
+++ b/src/DIRAC/TransformationSystem/Client/TaskManager.py
@@ -68,7 +68,7 @@ class TaskBase(TransformationAgentsUtilities):
         for taskID, task in taskDict.items():
             transID = task["TransformationID"]
             if task["Success"]:
-                res = self.transClient.setTaskStatusAndWmsID(transID, taskID, "Submitted", str(task["ExternalID"]))
+                res = self.transClient.setTaskStatusAndWmsID(transID, int(taskID), "Submitted", str(task["ExternalID"]))
                 if not res["OK"]:
                     self._logWarn(
                         "Failed to update task status after submission",
--- a/src/DIRAC/TransformationSystem/DB/TransformationDB.py
+++ b/src/DIRAC/TransformationSystem/DB/TransformationDB.py
@@ -705,7 +705,8 @@ class TransformationDB(DB):
 
         for error, fileIDStatusList in statusFileDict.items():
             req = reqBase + ",".join(
-                "(%d, %d, '%s', 0, UTC_TIMESTAMP())" % (transID, fileID, status) for fileID, status in fileIDStatusList
+                "(%d, %d, '%s', 0, UTC_TIMESTAMP())" % (transID, int(fileID), status)
+                for fileID, status in fileIDStatusList
             )
             if error:
                 # Increment the error counter when we requested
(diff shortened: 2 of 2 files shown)


fstagni (dismissed):

src/DIRAC/TransformationSystem/Client/TaskManager.py
```diff
@@ -68,7 +68,7 @@ def updateDBAfterTaskSubmission(self, taskDict):
         for taskID, task in taskDict.items():
             transID = task["TransformationID"]
             if task["Success"]:
-                res = self.transClient.setTaskStatusAndWmsID(transID, taskID, "Submitted", str(task["ExternalID"]))
+                res = self.transClient.setTaskStatusAndWmsID(transID, int(taskID), "Submitted", str(task["ExternalID"]))
```
chaen, line 71: You are correct that it very likely comes from `JEncode`, but I do not think this is the correct fix. I assume that `taskID` should already be an `int`.
The only place I can see `updateDBAfterTaskSubmission` is 
https://github.com/DIRACGrid/DIRAC/blob/1a6fc08cf01e1fa2431eef5da5209efd848ffc3e/src/DIRAC/TransformationSystem/Agent/TaskManagerAgentBase.py#L663

Looking up the chain, my unntested guess is that we should implement `getTasksToSubmit` in the client with the proper conversion. See for example:

https://github.com/DIRACGrid/DIRAC/blob/1a6fc08cf01e1fa2431eef5da5209efd848ffc3e/src/DIRAC/DataManagementSystem/Service/FTS3ManagerHandler.py#L156-L164

andresailer: You could test your assumption by adding the conversion here, I guess

https://github.com/DIRACGrid/DIRAC/blob/1a6fc08cf01e1fa2431eef5da5209efd848ffc3e/src/DIRAC/TransformationSystem/Agent/TaskManagerAgentBase.py#L607

chaen: I don't have an issue trying the proper fix as a hotfix, but I do I trigger the problem on the certif instance ?

andresailer: The test from here https://trello.com/c/Rk7HHlxy/6-submit-replication-transformation

src/DIRAC/TransformationSystem/DB/TransformationDB.py
```diff
@@ -705,7 +705,8 @@ def setFileStatusForTransformation(self, transID, fileStatusDict=None, connectio
 
         for error, fileIDStatusList in statusFileDict.items():
             req = reqBase + ",".join(
-                "(%d, %d, '%s', 0, UTC_TIMESTAMP())" % (transID, fileID, status) for fileID, status in fileIDStatusList
+                "(%d, %d, '%s', 0, UTC_TIMESTAMP())" % (transID, int(fileID), status)
```
chrisburr, line 708:
```suggestion
                f"({transID}, {fileID}, '{status}', 0, UTC_TIMESTAMP())"
```

PR #5798, opened 2022-01-14
[8.0] fix (JEncode): do NOT blindy decode bytes

chaen:
This was triggered by this issue during the hackaton (https://trello.com/c/GQo62UoN/15-verify-monitoring-application)

```python
2022-01-13 10:07:54 UTC Monitoring/Monitoring VERBOSE: Time for WMSHistory:NumberOfJobs - Report 0.00 Plot 0.00 (99.34% r/p)
2022-01-13 10:07:54 UTC Monitoring/Monitoring INFO: Returning plot file: 9a889123620787924c02fcbd8b36b24b.png
2022-01-13 10:07:54 UTC Monitoring/Monitoring ERROR: Uncaught exception when serving Transfer toClient
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/RequestHandler.py", line 198, in __doFileTransfer
    uRetVal = self.transfer_toClient(fileInfo[0], fileInfo[1], fileHelper)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/MonitoringSystem/Service/MonitoringHandler.py", line 137, in transfer_toClient
    retVal = fileHelper.sendData(retVal["Value"])
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/private/FileHelper.py", line 65, in sendData
    retVal = self.oTransport.sendData(S_OK([True, sBuffer]))
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/private/Transports/BaseTransport.py", line 169, in sendData
    sCodedData = MixedEncode.encode(uData)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/MixedEncode.py", line 22, in encode
    return JEncode.encode(inData)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/JEncode.py", line 183, in encode
    return json.dumps(inData, cls=DJSONEncoder)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/JEncode.py", line 110, in default
    return obj.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```
The `bytes` decoding was added in `JEncode` by this commit https://github.com/DIRACGrid/DIRAC/pull/5149/commits/64910e5e8e5bf0368081b5376a9750baaabe9fca  so this is basically reverting it.

First of all, changing something as low level and fundamental as `JEncode` without updating the unit tests is borderline criminal. 

Secondly, we should not blindly try to decode, since what you are given is not necessarily `utf-8` compatible (and above is such an example). What reaches `JEncode` should be serializable over `json` so it is up to the caller to do what is needed, for example converting in `base64`. 

This PR will not fix the issue that the `Monitoring` has with `JEncode`, the error will just be different. 

The adoption of `JEncode` comes again 5 years too late, since the original idea was to make the adoption of `https` easier. So I think we should let `Monitoring` be as it is, just disable `JEncode` for it. If we do an `https` version, we can use `base64`. That is exactly what the CS does. 
Now, ultimately, as I already said, we should not be sending files with Tornado and instead of wasting even more time in trying to force that to happen, we should just run with what we have and focus on a clean solution

BEGINRELEASENOTES
*Core
FIX: JEncode do not transparently decode bytes

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/JEncode.py
+++ b/src/DIRAC/Core/Utilities/JEncode.py
@@ -105,9 +105,6 @@ class DJSONEncoder(json.JSONEncoder):
         # if the object inherits from JSJerializable, try to serialize it
         elif isinstance(obj, JSerializable):
             return obj._toJSON()  # pylint: disable=protected-access
-        # if the object a bytes, decode it
-        elif isinstance(obj, bytes):
-            return obj.decode()
 
         # otherwise, let the parent do
         return super(DJSONEncoder, self).default(obj)


aldbr:
Something to take into account: just reverting https://github.com/DIRACGrid/DIRAC/commit/64910e5e8e5bf0368081b5376a9750baaabe9fca triggers a bug in `SiteDirector` when it tries to download proxies:

```python
2022-01-14 09:54:06 UTC WorkloadManagement/SiteDirector ERROR: Agent exception while calling method <bound method SiteDirector.execute of <DIRAC.WorkloadManagementSystem.Agent.SiteDirector.SiteDirector object at 0x7efcf57e8e20>>
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Base/AgentModule.py", line 341, in am_secureCall
    result = functor(*args)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py", line 324, in execute
    result = self.updatePilotStatus()
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py", line 1042, in updatePilotStatus
    result = gProxyManager.getPilotProxyFromDIRACGroup(self.pilotDN, self.pilotGroup, 23400)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/FrameworkSystem/Client/ProxyManagerClient.py", line 407, in getPilotProxyFromDIRACGroup
    return self.downloadVOMSProxy(
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/ThreadSafe.py", line 26, in lockedFunc
    return funcToCall(*args, **kwargs)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/FrameworkSystem/Client/ProxyManagerClient.py", line 339, in downloadVOMSProxy
    retVal = rpcClient.getVOMSProxy(
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Base/Client.py", line 107, in executeRPC
    return getattr(rpcClient, toExecute)(*parms)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/RPCClient.py", line 43, in __call__
    return self.__doRPCFunc(self.__remoteFuncName, args, **kwargs)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/RPCClient.py", line 89, in __doRPC
    return self.__innerRPCClient.executeRPC(sFunctionName, args, **kwargs)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/private/InnerRPCClient.py", line 69, in executeRPC
    retVal = transport.sendData(S_OK(list(args)))
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/DISET/private/Transports/BaseTransport.py", line 169, in sendData
    sCodedData = MixedEncode.encode(uData)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/MixedEncode.py", line 22, in encode
    return JEncode.encode(inData)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/JEncode.py", line 183, in encode
    return json.dumps(inData, cls=DJSONEncoder)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/site-packages/DIRAC/Core/Utilities/JEncode.py", line 113, in default
    return super(DJSONEncoder, self).default(obj)
  File "/opt/dirac/versions/v8.0.0a12-1642062208/Linux-x86_64/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable

```

chaen:
I believe @TaykYoku has more insight into it, since he introduced the `decode`

TaykYoku:
Hi @aldbr, your exception is that `JEncod.encode` gets bytes type value and tries to process them with a standard JSON encoder. This is to be expected, as the proposed PR changes do not fix the hackathon problem.

How I understood the situation:
`DIRAC_USE_JSON_ENCODE` env variable was defined on the hackathon, which changed the behavior of the **DISET** protocol, namely `DIRAC.Core.DISET.private.Transports.BaseTransport.sendData` -> `MixedEncode.encode`
```
    if os.getenv("DIRAC_USE_JSON_ENCODE", "NO").lower() in ("yes", "true"):
        return JEncode.encode(inData)
    return DEncode.encode(inData)
```

`DEncode.encode` return **bytes**(so it just skip bytes encoding) while `JEncode.encode` - **str**(so it should decode bytes, otherwise - a exception).

@chaen thank you for the timely reminder of the tests and kind words
this is decoding for the standard case(utf-8), I didn't think anyone would need to decode the image this way

chaen:
Well, no need to write new tests, as this commit must be reverted anyway :-) What needs doing is fixing what it breaks

TaykYoku:
bytes are transferred to the encoder anyway, in DEncode, in JEncode, so IMHO tests should reflect this case (to be honest, I expected that this is already taken into account).  And I did not understand that these lines are violated?  json always returns str, so everything in JEncode must be converted to str, see, for example, datetime.  the reason for the hackathon error is an attempt to transfer the file bytes via DISET using JEncode, I'm still trying to understand why?  file transfer for http services should be different, and DISET transmits bytes, so why pass them through JEncode?

chaen:
It showed up in the hackathon because `DIRAC_USE_JSON_ENCODE` was set. 

The service should NOT be passing `bytes` if we want to be JSON compatible. 

And from my point of view, it is a waste of time to try and fix the DISET monitoring service to make it JSON compatible. We should either:
* just have the https implementation do what it must (i.e. `base64` like we did for the CS) and forget about the exact interface compatibility with the DISET version. But that will be crap (Tornado is not meant to send files)
* Just keep running the problematic services (like `Monitoring`) with DISET until we have a real clean replacement, and disable the use of `JEncode` for these specific services

TaykYoku:
> The service should NOT be passing bytes if we want to be JSON compatible.

- I think `datetime.date` and `datetime.datetime` is not as compatible with JSON as `bytes`, but it is taken into account, no?
- If services are to be written in full JSON compatibility, then we should probably use `json` instead of `JEncode`, no?
- Looking at my changes(last discussion) you probably noticed that I use `JEncode` to convert python dict to json string, more understandable for tornado (BaseRequestHandler). So if this is not a valid function of JEncode class, then I have to write another.

> Tornado is not meant to send files

Well, we still have the practice of using `nginx`, in my opinion, not the best option, but an option. And **tornado** has an [issue](https://github.com/tornadoweb/tornado/issues/1448) regarding the use of `sendfile`, which should significantly improve its performance.

------------------
IMHO:
I have a hunch that we need to rethink our attitude to the argument types handling, it seems to me that a well-optimized argument type handler, combined with the use of annotation and typing, will be fast and allow us to use all the python data types we are used to.

chaen:
ok, let's not embark into complex discussions, and strip it down to the minimum:
This is how standard json compatible data (i.e. not `bytes) work, in a very simplified way

`myService -> JEncode -> json encode ------------- NETWORK ---------- json decode -> JEncode -> myClient`

Without `bytes.encode` that you added, what happens is 

`myService -> JEncode -> XX boom`. `json` cannot serialize bytes

So you added `if bytes, encode`

`myService -> JEncode -> "\x69fdfsd".encode() -> json -------- NETWORK ------ json -> JEncode -> myClient`

Now, two things:
* this call to encode will sometimes fail (that's the origin of this PR)
* You have to decode on the other side

Indeed, `datetime` is not json standard, and we added support to it. And this works because the format is strictly defined (https://github.com/DIRACGrid/DIRAC/blob/integration/src/DIRAC/Core/Utilities/JEncode.py#L13) , the conversion to it cannot fail (as opposed to the `bytes.encode`) and the decoding is done on the receiving end. 

Now, by all mean, be my guest, and take care of this card
https://trello.com/c/GQo62UoN/15-verify-monitoring-application

PR #5802, opened 2022-01-18
[8.0] solve the problem with bytes on the hackathon (wait #5811)

TaykYoku:
The essence of the problem:
when checking compatibility with **JSON**, we are faced with the reality that bytes can also be in a result, and `JEncode` does not accept them.

`JEncode` already has primitive logic that converts bytes to string (**_utf-8_**), but as it turned out, this is not enough if we shove it into `DISET`, so I added `base64` here and now the `bytes` on the input remain `bytes` on the output.

This is the best solution I see at the moment, while we are thinking about an ideal/clean approach.

#5798 

BEGINRELEASENOTES

*Core
FIX:  handle bytes in JEncode 

*Framework
FIX: use int instead of long

*Resources
FIX: use int instead of long

*StorageManagement
FIX: use int instead of long

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/DEncode.py
+++ b/src/DIRAC/Core/Utilities/DEncode.py
@@ -17,7 +17,6 @@ from __future__ import division
 
 __RCSID__ = "$Id$"
 
-from past.builtins import long
 import six
 import datetime
 import os
@@ -44,11 +43,9 @@ def _ord(char):
 # part of the HTTPS transition.
 class types(object):
     IntType = int
-    LongType = long
     FloatType = float
     BooleanType = bool
     StringType = str
-    UnicodeType = type(u"")
     NoneType = type(None)
     ListType = list
     TupleType = tuple
@@ -248,20 +245,6 @@ def decodeInt(data, i):
 
 g_dEncodeFunctions[types.IntType] = encodeInt
 g_dDecodeFunctions[_ord("i")] = decodeInt
-
-
-def encodeLong(iValue, eList):
-    """Encoding longs"""
-    raise TypeError("Data type long is no longer supported, please use int or float.")
-
-
-def decodeLong(data, i):
-    """Decoding longs"""
-    raise TypeError("Data type long is no longer supported, please use int or float.")
-
-
-g_dEncodeFunctions[types.LongType] = encodeLong
-g_dDecodeFunctions[_ord("I")] = decodeLong
 
 
 def encodeFloat(iValue, eList):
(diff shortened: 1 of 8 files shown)


src/DIRAC/StorageManagementSystem/DB/StorageManagementDB.py
```diff
@@ -1343,7 +1342,7 @@ def getCacheReplicasSummary(self, connection=False):
         resSummary = {}
         i = 1
         for status, se, numFiles, sumFiles in res["Value"]:
-            resSummary[i] = {"Status": status, "SE": se, "NumFiles": long(numFiles), "SumFiles": float(sumFiles)}
+            resSummary[i] = {"Status": status, "SE": se, int("NumFiles"): numFiles, "SumFiles": float(sumFiles)}
```
chrisburr, line 1345: I guess this was supposed to be:
```suggestion
            resSummary[i] = {"Status": status, "SE": se, "NumFiles": int(numFiles), "SumFiles": float(sumFiles)}
```

TaykYoku:
I checked on lbcertifdirac70.cern.ch and it looks like it works as expected.
Moreover, as far as I understand it is possible to remove "base64 logic" and `CSJSONClient` for a configuration with this change without loss of performance. The same goes for `BundleDeliveryJSONClient`.

src/DIRAC/Core/Utilities/DEncode.py
```diff
@@ -44,11 +43,9 @@ def _ord(char):
 # part of the HTTPS transition.
 class types(object):
     IntType = int
-    LongType = int
     FloatType = float
     BooleanType = bool
     StringType = str
-    UnicodeType = type(u"")
```
chrisburr, line 51: I think this is still needed until v8.1 so that Python 2 based v7r3 clients can still communicate with the server.

src/DIRAC/Core/Utilities/JEncode.py
```diff
@@ -107,7 +108,7 @@ def default(self, obj):  # pylint: disable=method-hidden
             return obj._toJSON()  # pylint: disable=protected-access
         # if the object a bytes, decode it
         elif isinstance(obj, bytes):
-            return obj.decode()
+            return {"__dCls": "b64", "obj": b64encode(obj).decode()}
```
chrisburr, line 111: Can you update the encoding tests to include one that gets hypothesis to test using `bytes`

chrisburr (changes requested):
This should be rebased on rel-v7r2

src/DIRAC/Core/Utilities/JEncode.py
TaykYoku: done in #5811

PR #5854, opened 2022-02-03
[v8r0] Treat X509 data as str rather than bytes and add DISET+JEncode CI

chrisburr:
(This explanation is rather rushed but it's all I have time for right now, I can discuss it further later if needed.)

The issue is that DEncode doesn't have a good distinction between `bytes` and `str` so when moving to Python 3 I made it treat everything as strings. In the handful of cases that bytes are truly needed I used the [surrogate escapes](https://www.python.org/dev/peps/pep-0383/). (This is a horrible hack but keeps DEncode working the way it always has.) With `JEncode` this isn't an option as the actual wire format is now defined to be text so it can't pass around the invalid unicode sequences.

The biggest break is that we intentionally decided to treat PEM-encoded x509 as `bytes`. I'd forgotten that we sent them over the network therefore `DEncode` could quietly comvert them to strings (because the PEM encoding is base64+ASCII based). This PR bites the bullet and moves the PEM certificate handling to use strings, thankfully the test coverage is decent so it should be mostly working.

This PR also fixes a bunch of other cases where bytes were accidentally used and adds CI that uses JEncode so we can avoid future breaks.

BEGINRELEASENOTES

*Core
CHANGE: Treat X509 data as str instead of bytes

*Framework
FIX: UserProfileManager usage with JEncode

*WorkloadManagement
FIX: JobDB usage with JEncode

*Production
FIX: ProductionManager use with JEncode

ENDRELEASENOTES


chrisburr:
Thanks for the reminder :) I've updated the PR description.

.github/workflows/integration.yml
```diff
@@ -37,6 +37,8 @@ jobs:
             ARGS: MYSQL_VER=mariadb:10.6.3 ES_VER=opensearchproject/opensearch:1.0.0
           - TEST_NAME: "HTTPS"
             ARGS: TEST_HTTPS=Yes
+          - TEST_NAME: "Force JEncode"
+            ARGS: DIRAC_USE_JSON_ENCODE=Yes DIRAC_USE_JSON_DECODE=Yes
```
fstagni, line 41: ...just to say that this was the way to go! I don't know why we didn't think to do this before!

fstagni (approved):
This looks to me like the correct approach and I would give it a go in tomorrow's hackathon

src/DIRAC/FrameworkSystem/Client/UserProfileClient.py
```diff
@@ -17,7 +17,7 @@ def __getRPCClient(self):
 
     def storeVar(self, varName, data, perms={}):
         try:
-            stub = DEncode.encode(data)
+            stub = DEncode.encode(data).decode()
```
TaykYoku, line 20: this class I think better inherit `Client`

TaykYoku (commented):
Overall, it looks good to me

TaykYoku:
I'm a little worried that such a change could affect extensions

src/DIRAC/FrameworkSystem/DB/ProxyDB.py
```diff
@@ -427,7 +427,7 @@ def __storeProxy(self, userDN, chain, proxyProvider=None):
         sqlInsert = True
         if data:
             sqlInsert = False
-            pem = data[0][1]
+            pem = data[0][1].decode("ascii")
```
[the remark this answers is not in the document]
chrisburr: Not really but I think it's safer to make clear that this should be decoded (or better yet, changed to `TEXT` in the database).

tests/Integration/Framework/Test_UserProfileClient.py
```diff
@@ -41,7 +41,7 @@ def test_storeRetrieveDelete():
 
     retVal = up.listAvailableVars()
     assert retVal["OK"], retVal
-    assert retVal["Value"] == (("adminusername", currentGroup, userVO or "undefined", key),)
+    assert retVal["Value"] == [["adminusername", currentGroup, userVO or "undefined", key]]
```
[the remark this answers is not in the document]
chrisburr: I fixed this in the UserProfileDB to be consistent.

(Search for `# Convert returned tuples to lists to appease JEncode`)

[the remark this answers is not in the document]
TaykYoku: Ah, sorry, seems I missed that

src/DIRAC/FrameworkSystem/Client/UserProfileClient.py
chrisburr: I don't understand what you mean here?

TaykYoku: I mean that `DEncode` or `JEncode` encoding should be inside `Client` logic, IIUC

chrisburr: The `UserProfileClient` stores `DEncode`-ed objects in the database. Maybe that should be changed but definitely not in this PR.

chrisburr:
> I'm a little worried that such a change could affect extensions

How exactly? Are there any active extensions to the proxy manager?

TaykYoku:
This for example https://github.com/DIRACGrid/WebAppDIRAC/pull/630

TaykYoku (approved):

PR #5917, opened 2022-03-03
[v7r2] Add SourceSE filter to RMS ReplicateAndRegister

### LINKED ISSUE #5893: ReplicateAndRegister Operation doesn't consider SourceSE in FTS3 Mode?
Hi,

We've finally set-up our RMS system for data transfers with all of the FTS3 stuff in production and it's working quite well, however we've run in to one issue... An end user has submitted some ReplicateAndRegister tasks with about 50 files each but the RequestExecutingAgent is processing them _really_ slowly (the total throughput is far less than 1 request per minute, even with 32 threads in the pool).

I've been investigating this and found the following things:
 * Each LFN has around 5 replicas.
 * The RMS operations have a sourceSE set by the user to a preferred SE that is accessible.
 * One of the SEs that also has most of the files (but is not the sourceSE) is completely offline (connection timed-out).
 * We don't have the RSS system enabled.

I've found that while the classic dmTransfer function did used to filter based on the requested source:
https://github.com/DIRACGrid/DIRAC/blob/e9b59389e7362c85c2f19e2f67089c168d5bacfd/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py#L501

The newer FTS3 stuff seems to have a loop that queries all of the replicas of each LFN directly and doesn't consider the sourceSE at this point at all:
https://github.com/DIRACGrid/DIRAC/blob/e9b59389e7362c85c2f19e2f67089c168d5bacfd/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py#L109-L110

This seems to be quite a lot of extra work to be doing considering we know the sourceSE we'd like to use at this point. Have I misunderstood how this all fits together? If not, would it be possible to consider only querying the operation's sourceSE here if it's set?
(I think the main surprise is that the user has asked to copy from site A to site B and suddenly we've found that we're also sending a lot of potentially slow query traffic to Sites X, Y & Z!).

Regards,
Simon

sfayer:
Hi,

This is the patch to filter the replicas by the SourceSE when preparing a R&R request to avoid contacting SEs unnecessarily.

Closes #5893.

Regards,
Simon

BEGINRELEASENOTES
*RequestManagementSystem
FIX: filter by SourceSE in ReplicateAndRegister preparation
FIX: Change RMS.Operation sourceSEList and targtSEList default to empty list
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
+++ b/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
@@ -52,7 +52,7 @@ from DIRAC.ConfigurationSystem.Client.Helpers import Registry
 from DIRAC.MonitoringSystem.Client.MonitoringReporter import MonitoringReporter
 
 
-def filterReplicas(opFile, logger=None, dataManager=None):
+def filterReplicas(opFile, logger=None, dataManager=None, opSources=None):
     """filter out banned/invalid source SEs"""
 
     if logger is None:
@@ -76,6 +76,11 @@ def filterReplicas(opFile, logger=None, dataManager=None):
         return S_ERROR(failed)
 
     replicas = replicas["Successful"].get(opFile.LFN, {})
+
+    # If user set sourceSEs, only consider those replicas
+    if opSources:
+        replicas = {x: y for (x, y) in replicas.items() if x in opSources}
+
     noReplicas = False
     if not replicas:
         allReplicas = dataManager.getReplicas(opFile.LFN, getUrl=False)
@@ -291,7 +296,7 @@ class ReplicateAndRegister(DMSRequestOperationsBase):
 
     def _filterReplicas(self, opFile):
         """filter out banned/invalid source SEs"""
-        return filterReplicas(opFile, logger=self.log, dataManager=self.dm)
+        return filterReplicas(opFile, logger=self.log, dataManager=self.dm, opSources=self.operation.sourceSEList)
 
     def _checkExistingFTS3Operations(self):
         """
--- a/src/DIRAC/RequestManagementSystem/Client/Operation.py
+++ b/src/DIRAC/RequestManagementSystem/Client/Operation.py
@@ -227,7 +227,7 @@ class Operation(object):
     @property
     def targetSEList(self):
         """helper property returning target SEs as a list"""
-        return self.TargetSE.split(",") if self.TargetSE else [""]
+        return self.TargetSE.split(",") if self.TargetSE else []
 
     @property
     def Arguments(self):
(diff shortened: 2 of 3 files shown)


andresailer (dismissed):
Looks good to me

src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
```diff
@@ -52,13 +52,20 @@
...
         logger = gLogger
     if dataManager is None:
         dataManager = DataManager()
+    # Create a set of sourceSEs specificed by user
+    userSources = None
+    if opSources:
+        if isinstance(opSources, list):
+            userSources = set(opSources)
+        else:
+            userSources = set([x.strip() for x in opSources.split(",")])
```
chaen, line 68: you could use
 https://github.com/DIRACGrid/DIRAC/blob/3476d79c6c2ec439c90b0da733a80e889042596c/src/DIRAC/RequestManagementSystem/Client/Operation.py#L211
 
 instead of just `sourceSE` when calling `filterReplicas`, so you don't even need to do the strip and split yourself
 
Though I wonder why the hell is the default value `[""]` That clearly seems wrong to me :thinking:

sfayer: I'm happy to use the sourceSEList but that default is a problem... I'd have to do something horrible like `if len(sources) == 1 and not sources[0]: ...` to check if the user has set anything or not!

Would you be happy for the default to be changed to an empty list? At least then it'll evaluate to False if the user hasn't set anything.

chaen: I believe the current default does not make any sense !

src/DIRAC/RequestManagementSystem/Client/Operation.py
```diff
@@ -222,7 +222,7 @@ def __len__(self):
     @property
     def sourceSEList(self):
         """helper property returning source SEs as a list"""
-        return self.SourceSE.split(",") if self.SourceSE else [""]
+        return self.SourceSE.split(",") if self.SourceSE else []
```
andresailer, line 225: Same for targetSE couple lines below?

sfayer: Sure, I was a bit dubious about changing that one (my specific worry being if someone had a `for x in targetSEList` loop somewhere relying on the default containing a single element), but changing it doesn't seem to break any of my tests.

sfayer: It seems some of the modules (CheckMigration) do rely on the "empty string" target list to effectively act as a wildcard... I've reverted it back to only changing the source default for now.

andresailer: I think it is just the tests expecting that for some reason, I will try to have a look later today.
Fine for me to merge this regardless, but having these two properties with the same default sounds saner to me.

andresailer: I adapted the test by explicitly setting targetSE

chaen (approved):

PR #5946, opened 2022-03-14
[v7r3] Escape Time stamp in JobDB

phicharp:
BEGINRELEASENOTES

* WMS
FIX: escape time stamp in setHeartBeatData

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
+++ b/src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
@@ -1967,7 +1967,11 @@ class JobDB(DB):
         # If HeartBeatTime is being set, set it...
         timeStamp = dynamicDataDict.pop("HeartBeatTime", None)
         if timeStamp:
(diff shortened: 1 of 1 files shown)


src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
```diff
@@ -1967,7 +1967,11 @@ def setHeartBeatData(self, jobID, dynamicDataDict):
         # If HeartBeatTime is being set, set it...
         timeStamp = dynamicDataDict.pop("HeartBeatTime", None)
         if timeStamp:
-            req = "UPDATE Jobs SET HeartBeatTime=%s WHERE JobID=%s" % (timeStamp, e_jobID)
+            result = self._escapeString(timeStamp)
+            if not result["OK"]:
+                self.log.warn("Failed to escape string ", timeStamp)
```
andresailer, line 1972: Need to leave the function here, otherwise `req` is no defined and 1981 gives an exception, right?

PR #5947, opened 2022-03-14
[v7r3]fix (RequestTask): fix error referenced before assignment

chaen:
Although not obvious, this does not work

```python
err=None
try:
  raise ValueError('boom')
except Exception as err:
  print("Caught!")
if err:
  print("there is err")
```

```bash
python /tmp/testErr.py 
Caught!
  File "/tmp/testErr.py", line 6, in <module>
    if err:
NameError: name 'err' is not defined
```

The second fix is a follow up to https://github.com/DIRACGrid/DIRAC/pull/5917  This PR did the correct thing by making the default an empty list, but now we understand why it wasn't an empty list :laughing: 

BEGINRELEASENOTES
*RMS
FIX: fix variable referenced before assignment in RequestTask
FIX: take into account that targetSEList can be empty in the operations

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Agent/RequestOperations/RegisterFile.py
+++ b/src/DIRAC/DataManagementSystem/Agent/RequestOperations/RegisterFile.py
@@ -95,7 +95,13 @@ class RegisterFile(OperationHandlerBase):
             # # get LFN
             lfn = opFile.LFN
             # # and others
-            fileTuple = (lfn, opFile.PFN, opFile.Size, self.operation.targetSEList[0], opFile.GUID, opFile.Checksum)
+
+            # CHRIS: for whatever reason, we only take the first one
+            targetSE = ""
+            if self.operation.targetSEList:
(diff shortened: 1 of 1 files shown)


src/DIRAC/DataManagementSystem/Agent/RequestOperations/RegisterFile.py
```diff
@@ -95,7 +95,14 @@ def __call__(self):
             # # get LFN
             lfn = opFile.LFN
             # # and others
-            fileTuple = (lfn, opFile.PFN, opFile.Size, self.operation.targetSEList[0], opFile.GUID, opFile.Checksum)
+
+            # CHRIS 22.03.14
```
andresailer, line 99: We have he git log, so we don't need the date.

chaen: done

andresailer (approved):

PR #5967, opened 2022-03-17
[v7r3] Fix race condition in FTS3DB.getNonFinishedOperations

chaen:
The reasons and explanations for this change are described in detail in `test_raceCondition`

This has been running under very heavy load in LHCb for a week
BEGINRELEASENOTES
*DMS

FIX: fix a race condition when running multiple FTS3Agents in parallel

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Agent/FTS3Agent.py
+++ b/src/DIRAC/DataManagementSystem/Agent/FTS3Agent.py
@@ -54,6 +54,11 @@ AGENT_NAME = "DataManagement/FTS3Agent"
 # Lifetime in seconds of the proxy we download for submission
 PROXY_LIFETIME = 43200  # 12 hours
 
+# Instead of querying many jobs at once,
+# which maximizes the possibility of race condition
+# when running multiple agents, we rather do it in steps
+JOB_MONITORING_BATCH_SIZE = 20
+
 
 class FTS3Agent(AgentModule):
     """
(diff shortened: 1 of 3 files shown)


atsareg:
Looks fine to me, but I do not feel expert here

src/DIRAC/DataManagementSystem/DB/FTS3DB.py
```diff
@@ -537,12 +544,13 @@ def getNonFinishedOperations(self, limit=20, operationAssignmentTag="Assigned"):
...
             # We need to do the select in two times because the join clause that makes the limit difficult
+            # We get the list of operations ID that have associated jobs assigned
+            opIDsWithJobAssigned = session.query(FTS3Job.operationID).filter(~FTS3Job.assignment.is_(None)).subquery()
             operationIDsQuery = (
                 session.query(FTS3Operation.operationID)
-                .outerjoin(FTS3Job)
                 .filter(FTS3Operation.status.in_(["Active", "Processed"]))
                 .filter(FTS3Operation.assignment.is_(None))
-                .filter(FTS3Job.assignment.is_(None))
+                .filter(~FTS3Operation.operationID.in_(opIDsWithJobAssigned))
```
andresailer, line 553: This fixes the race condition? Is the splitting of monitoring into batches still needed to fix the problem?

andresailer (commented):
Just a typo

src/DIRAC/DataManagementSystem/DB/FTS3DB.py
chaen: This fixes the race condition.
Splitting into batches optimizes even further because we do not block at once many jobs/operations.

src/DIRAC/DataManagementSystem/Agent/FTS3Agent.py
```diff
@@ -54,6 +54,11 @@
 # Lifetime in seconds of the proxy we download for submission
 PROXY_LIFETIME = 43200  # 12 hours
 
+# Instead of querying many jobs at once,
+# which maximizes the possibility of race condition
```
andresailer, line 58: Just wondering if you want to rephrase this now that the race condition is fixed

PR #5971, opened 2022-03-19
[v7r2] fix: Limit reoccurrences of subprocess unicode error

sfayer:
Hi,

Here is the patch to limit the number of times the unicode error message will be printed for #5968.

Regards,
Simon

BEGINRELEASENOTES
*Core
FIX: Limit reoccurrences of subprocess unicode error
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/Subprocess.py
+++ b/src/DIRAC/Core/Utilities/Subprocess.py
@@ -372,6 +372,7 @@ class Subprocess(object):
     def __readFromFile(self, fd, baseLength):
         """read from file descriptor :fd: and save it to the dedicated buffer"""
         try:
+            numErrors = 0
             dataString = ""
             fn = fd.fileno()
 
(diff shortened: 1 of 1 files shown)


fstagni (dismissed):

src/DIRAC/Core/Utilities/Subprocess.py
```diff
@@ -389,7 +390,11 @@ def __readFromFile(self, fd, baseLength):
                 try:
                     dataString += nB.decode()
                 except UnicodeDecodeError as e:
-                    self.log.warn("Unicode decode error in readFromFile", "(%r): %r" % (e, dataString))
+                    if numErrors < 5:
+                        self.log.warn("Unicode decode error in readFromFile", "(%r): %r" % (e, dataString))
```
chrisburr, line 394:
```suggestion
                        self.log.warn("Unicode decode error in readFromFile", "(%r): %r" % (e, dataString[max(0, e.start-10):e.end+10]))
```

fstagni (approved):

PR #5996, opened 2022-04-01
[8.0] Upgrade Stomp version and implement simpler interface

chaen:
The existing support of MQ in DIRAC was designed with the idea of abstracting as much as possible the type of queues and the protocol used. It is all very well, but at the cost of:
* extremely complex code
* very inefficient because way too many abstraction layers
* Has bugs (can't listen to an alias with multiple host behind for example)
* Was done with the idea of exposing each and every single possibility of usage

It is a very theoretical work, with no actual understanding of how you really want to use MQ. It was also design with some wrong understanding of the different roles in a MQ system. In practice, there is ALWAYS a relation between what you actually want to do  with your messages, and how you configure it. That's why so much abstraction is useless. 

This PR builds on the LHCb experience, and make some compromises on generality:
* We want to write and listen to some destinations behind an alias
* We have only ever seen `stomp`
* We only authenticate with username/password
* We want reconnection when the broker reboots or whatever

Maybe one day we see RabbitMQ in the landscape, and then it will still be time to factorize.

In the meantime, I have created this new code under `Resources/MessageQueue/Simple` not to disturb the existing.  We will see if we can little by little replace it, but with the few tests I have done, I have the feeling we can.

BEGINRELEASENOTES
*Resources
CHANGE: release stomp version
NEW: Simple stomp interface

ENDRELEASENOTES

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -40,7 +40,7 @@ dependencies:
   - requests >=2.9.1
   - six >=1.10
   - sqlalchemy
-  - stomp.py =4.1.23
+  - stomp.py
   - suds-jurko >=0.6
   - xmltodict
   - pycurl
--- a/src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
+++ b/src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
@@ -1,0 +1,406 @@
+import json
+import os
+import socket
+import hashlib
+import random
+import stomp
+
+from typing import Optional, Type
+
+from DIRAC import gConfig
+from DIRAC.Core.Utilities.ReturnValues import convertToReturnValue, returnValueOrRaise
+
+
+def _resolve_brokers(alias: str, port: int, ipv4Only: bool = False, ipv6Only: bool = False) -> list[tuple[str, int]]:
+    """
+    To consume all the messages, we need to subscribe to all the hosts behind
+    the DNS alias. In this case though, we should use either IPv4 or IPv6 but not both
+    to avoid double processing in case of topics.
+
+    To have Producers a bit balanced accross machines, we need to resolve the alias and
+    randomize it.
+
+
+    :param alias: The DNS alias
+    :param port: The TCP port
+    :param ipv4Only: Only return IPv4
+    :param ipv6Only: Only return IPv6
+    :return: A list of tuples (resolved ip, port)
+    """
+    assert not (ipv4Only and ipv6Only)
+
+    brokers = list()
+
+    for family, _, _, _, addr in socket.getaddrinfo(alias, port, 0, socket.SOCK_STREAM):
+        ip, port = addr[:2]
+        # Disable IPv6 until worker nodes can use it
+        if (family == socket.AF_INET and not ipv6Only) or (family == socket.AF_INET6 and not ipv4Only):
+            brokers.append((ip, port))
+
+    random.shuffle(brokers)
+    return brokers
+
+
+class ReconnectListener(stomp.ConnectionListener):
+    """Listener that takes care of reconnection"""
+
+    def __init__(self, connectCallback, *args):
+        """
+        :param connectCallback: callback to call for reconnection
+        :param args: all the arguments to pass to the connectCallback
+
+        """
+        # This boolean is to know whether we effectively
+        # want to disconnect or if it is because of a failure
+        self.wantsDisconnect = False
+        self.connectCallback = connectCallback
+        self.args = args
+
+    def on_disconnected(self):
+        """Callback function called after disconnecting from broker."""
+
+        if not self.wantsDisconnect:
+            try:
+                self.connectCallback(*self.args)
+            except Exception as e:
+                print(f"Unexpected error while calling reconnect callback: {e}")
+
+
+def getSubscriptionID(broker: tuple[str, int], dest: str) -> str:
+    """Generate a unique subscribtionID based on the broker host, port and destination
+
+    :param broker: tuple (host,port) to which we connect
+    :param dest: name of the destination (topic or queue)
+
+    """
+    host, port = broker
+    return hashlib.md5((f"{host}_{port}_{dest}").encode()).hexdigest()
+
+
+class StompConsumer:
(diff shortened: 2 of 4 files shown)


src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+        self.args = args
+
+    def on_disconnected(self):
+        """Callback function called after disconnecting from broker."""
+
+        if not self.wantsDisconnect:
+            try:
+                self.connectCallback(*self.args)
+            except Exception as e:
+                print("Unexpected error while calling reconnect callback: %s" % e)
```
chrisburr, line 66:
```suggestion
                print(f"Unexpected error while calling reconnect callback: {e}")
```
or
```suggestion
                print("Unexpected error while calling reconnect callback:", e)
```

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+
+    brokerParams = returnValueOrRaise(gConfig.getOptionsDict("/Resources/MQServices/%s" % mqService))
+    # This is for compatibility reasons with the existing configuration definition
+    # Although there are no reasons to separate queues and topics for stomp
+    topics = [
+        f"/topic/{dest}" for dest in gConfig.getSections("/Resources/MQServices/%s/Topics" % mqService).get("Value", [])
+    ]
+    queues = [
+        f"/queue/{dest}" for dest in gConfig.getSections("/Resources/MQServices/%s/Queues" % mqService).get("Value", [])
+    ]
```
chrisburr, line 339: f-strings

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+    port = int(brokerParams["Port"])
+    username = brokerParams["User"]
+    password = brokerParams["Password"]
+
+    csDestinations = brokerParams.pop("destinations", [])
+
+    if not destination:
+
+        if len(csDestinations) != 1:
+            return S_ERROR("There should be exactly one destination given in parameter or in the CS")
```
chrisburr, line 401:
```suggestion
            raise ValueError("There should be exactly one destination given in parameter or in the CS")
```

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+
+    host = brokerParams["Host"]
+    port = int(brokerParams["Port"])
+    username = brokerParams["User"]
+    password = brokerParams["Password"]
+
+    csDestinations = brokerParams.pop("destinations", None)
+    destinations = destinations or csDestinations
+    if not destinations:
+        return S_ERROR("Destinations should either be given as parameter or defined in the CS")
```
chrisburr, line 369:
```suggestion
        raise ValueError("Destinations should either be given as parameter or defined in the CS")
```

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+import os
+import socket
+import hashlib
+import random
+import stomp
+
+from typing import Optional, Type
+
+from DIRAC import gConfig
+from DIRAC.Core.Utilities.ReturnValues import S_ERROR, convertToReturnValue, returnValueOrRaise
```
chrisburr, line 11:
```suggestion
from DIRAC.Core.Utilities.ReturnValues import convertToReturnValue, returnValueOrRaise
```

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
```diff
@@ -0,0 +1,406 @@
...
+
+    if not destination:
+
+        if len(csDestinations) != 1:
+            return S_ERROR("There should be exactly one destination given in parameter or in the CS")
+        destination = csDestinations[0]
+
+    producer = StompProducer(host, port, username, password, destination)
+
+    return producer
```
chrisburr, line 406: I can't comment on it as it's empty but `src/DIRAC/Resources/MessageQueue/Simple/__init__.py` can be removed now we're Python 3-only.

environment.yml
```diff
@@ -40,7 +40,7 @@ dependencies:
   - requests >=2.9.1
   - six >=1.10
   - sqlalchemy
-  - stomp.py =4.1.23
+  - stomp.py
```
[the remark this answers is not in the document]
chaen: I believe yes, I did not find anything breaking, but I will cross check

src/DIRAC/Resources/MessageQueue/Simple/StompInterface.py
chaen: Not sure I understand what you mean. They are still mandatory as far as I understand:

https://docs.python.org/3/tutorial/modules.html

> The __init__.py files are required to make Python treat directories containing the file as packages. This prevents directories with a common name, such as string, unintentionally hiding valid modules that occur later on the module search path. In the simplest case, __init__.py can just be an empty file, but it can also execute initialization code for the package or set the __all__ variable, described later.

chrisburr: Indeed I've muddled some things up.

(Specificaly namespace packages which do work but have slightly different behaviour:

```bash
$ mkdir -p a/b/c
$ echo 'print("hello world")' > a/b/c/d.py
$ python -c 'import a.b.c.d'
hello world
```
)

environment.yml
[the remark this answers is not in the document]
chrisburr: Then we should relax it in DIRACOS2 as well :) https://github.com/DIRACGrid/DIRACOS2/blob/8b51d79eeda9b8be7255c908fc00fa0d904ebd86/construct.yaml#L58

fstagni (approved):

PR #6013, opened 2022-04-11
[v7r3] fix: use correct environment for condor_rm

marianne013:
Previous version didn't use the proxy set by setproxy, so it always failed with an auth error.
BEGINRELEASENOTES

*Resources
FIX: HTCondorCE: make sure proxy is available when running condor_rm

ENDRELEASENOTES


marianne013:
@andresailer This might also affect condor_status, it might be worth having a look.

marianne013:
@fstagni @andresailer Would you mind having a look at that test failure, please ? I think the test doesn't quite work as it's meant to.

andresailer (commented):
Otherwise looks fine to  me

andresailer:
And it really does not produce errors with local schedd

fstagni (approved):

andresailer (approved):

PR #6017, opened 2022-04-13
[8.0] Propagate loggers from services to DBs

chaen:
One major difference between `dips` and `diset` is that every DISET service was running in it's own process. This mean that the global `gLogger` could be configured with the name of the service. 
With Tornado, the only global name that makes sense is `Tornado/Tornado`. But it is then very difficult to distinguish between the different services. 

In order to do that, every `TornadoService` now declare a class attribute `self.log`, and adds an extra option to the dictionary sent to the centralized logging: `tornadoComponent`.  (3e640c0022b44c0203e4a78c5a9d5202f070c4c4)

This allows to further split down the logs.

The problem is that this logger is only used in the service, and not in the DB classes. So in order to allow for that attribute to be propagated, the services now CAN pass their logger as a parent logger to the DB classes. (9e40bffad19623ce0699d3d33c714703d340643d)
Of course, there will always be some loggers instantiated here and there that will not have that attributes, but we will track them down little by little

All the other commits of that PR are just about using these two previous features everywhere

I first developed it and tested in in LHCb production setup for v7r3. Migrating it to `integration` was a major pain, due to all the tornado rewrite that happened (I tried turning a blind eyes on the many crazy things like https://github.com/DIRACGrid/DIRAC/blob/14956b316cb07d8abb1996d4268106cec338bf47/src/DIRAC/Core/Tornado/Server/HandlerManager.py#L125)

BEGINRELEASENOTES

*Core
NEW: TornadoBaseRequestHandler define a class logger with an extra attribute "tornadoComponent"
NEW: DIRACDB can now use a parentLogger instead of direct gLogger

*All
CHANGE: services propagate their local loggers to the DB
CHANGE: tornado services use local logger

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Base/DIRACDB.py
+++ b/src/DIRAC/Core/Base/DIRACDB.py
@@ -14,7 +14,10 @@ class DIRACDB:
 
         :param self: self reference
         """
-        self.log = gLogger.getSubLogger(self.fullname)  # pylint: disable=no-member
+        parentLogger = kwargs.pop("parentLogger", None)
+        if not parentLogger:
+            parentLogger = gLogger
+        self.log = parentLogger.getSubLogger(self.fullname)  # pylint: disable=no-member
         super(DIRACDB, self).__init__(*args, **kwargs)
 
     def getCSOption(self, optionName, defaultValue=None):
--- a/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
+++ b/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
@@ -335,7 +335,6 @@ class BaseRequestHandler(RequestHandler):
         cls._authManager = AuthManager(cls._getCSAuthorizarionSection(cls._fullComponentName))
 
         if not (urls := cls._pre_initialize()):
-            # cls.log is not yet available
             cls.log.warn("no target method found", f"{cls.__name__}")
         return urls
 
--- a/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
+++ b/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
@@ -29,8 +29,6 @@ from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-er
 from DIRAC.Resources.IdProvider.Utilities import getProvidersForInstance
 from DIRAC.Resources.IdProvider.IdProviderFactory import IdProviderFactory
 
-sLog = gLogger.getSubLogger(__name__.split(".")[-1])
-
 
 def set_attribute(attr, val):
     """Decorator to determine target method settings. Set method attribute.
@@ -247,6 +245,7 @@ class BaseRequestHandler(RequestHandler):
     The first request starts the process of initializing the handler, see the :py:meth:`initialize` method:
 
         - load all registered identity providers for authentication with access token, see :py:meth:`__loadIdPs`.
+        - create a ``cls.log`` logger that should be used in the children classes instead of directly ``gLogger`` (this allows to carry the ``tornadoComponent`` information, crutial for centralized logging)
         - initialization of the monitoring specific to this handler, see :py:meth:`__initMonitoring`.
         - initialization of the target handler that inherit this one, see :py:meth:`initializeHandler`.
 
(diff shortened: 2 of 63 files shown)


src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
```diff
@@ -333,6 +334,7 @@ def __pre_initialize(cls) -> list:
         cls._authManager = AuthManager(cls._getCSAuthorizarionSection(cls._fullComponentName))
 
         if not (urls := cls._pre_initialize()):
+            # cls.log is not yet available
```
chrisburr, line 337: Maybe we should get rid of `sLog` in favour of something like this so `cls.log` always exists:
```python
class BaseRequestHandler(RequestHandler):
    # Subclasses should override this property
    log = gLogger.getSubLogger(__name__.split(".")[-1])
```

src/DIRAC/Core/Base/DIRACDB.py
```diff
@@ -14,7 +14,10 @@ def __init__(self, *args, **kwargs):
 
         :param self: self reference
         """
-        self.log = gLogger.getSubLogger(self.fullname)  # pylint: disable=no-member
+        parentLogger = kwargs.pop("parentLogger", None)
+        if not parentLogger:
+            parentLogger = gLogger
+        self.log = parentLogger.getSubLogger(self.fullname)  # pylint: disable=no-member
```
chrisburr, line 20: Should this emit a warning? (And possibly respect `DIRAC_DEPRECATED_FAIL`.)

src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
chaen: Indeed, doing this way would avoid the very special case of `__pre_initialize`. 
But why should the subclass override it ?

src/DIRAC/Core/Base/DIRACDB.py
chaen: I don't think so, I prefer to keep it backward compatible. Moreover, there are cases where it does not work (the RSS DBs are non sense but I managed to pass it their logger anyway, but the MonitoringDB is a desperate case). 
Also, it would be a real burden in tests

src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
chrisburr: I should have said `This will be overridden in __initialize to be handler specific`.

chaen: works with me, I change it now

chrisburr (dismissed):

src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
```diff
@@ -333,7 +335,8 @@ def __pre_initialize(cls) -> list:
         cls._authManager = AuthManager(cls._getCSAuthorizarionSection(cls._fullComponentName))
 
         if not (urls := cls._pre_initialize()):
-            sLog.warn("no target method found", f"{cls.__name__}")
+            # cls.log is not yet available
```
andresailer, line 338: But `cls.log` is available, because it is used in the next line?

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
```diff
@@ -59,7 +59,7 @@ def initializeHandler(cls, *args):
         try:
             cls.__tokenDB = TokenDB()
```
andresailer, line 60: Shouldn't TokenDB get the logger?

src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
chaen: right, outdated comment, I remove it

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
chaen: Yeah, `TokenDB` is part of a few special DB that do not have a default constructor, so I should modify it properly... I'll do that in this PR

fstagni (approved):

PR #6020, opened 2022-04-19
[v7r3] Fix previous fix in RMS

phicharp:
BEGINRELEASENOTES

* RMS

FIX: if the request ID is an integer the test `is digit()` throws an exception

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/RequestManagementSystem/scripts/dirac_rms_request.py
+++ b/src/DIRAC/RequestManagementSystem/scripts/dirac_rms_request.py
@@ -219,6 +219,7 @@ def main():
             # PEP-515 allows for underscore in numerical literals
             # So a request name 00123_00456
             # is interpreted as a requestID 12300456
+            # Using an exception here for non-string is not an option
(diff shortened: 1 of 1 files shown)


chrisburr (dismissed):

chaen:
I think it just would be nicer to catch  `AttributeError` in the list of exception below, but it does not really matter 

```python
In [2]: a = 3

In [3]: a.isdigit()
Input In [3], in <cell line: 1>()
----> 1 a.isdigit()

AttributeError: 'int' object has no attribute 'isdigit'
```

phicharp:
Thanks Chris 1&2!

phicharp:
Oups! I reverted to the initial fix as adding `AttributeError` wouldn't work as expected if `reqID` is an integer: we want it to not go through the `except` clause...

phicharp:
Sorry for the test failures on the commit messages... I always forget whether one needs a space before the ":" or not! ;-)

chrisburr (approved):

fstagni (approved):

PR #6045, opened 2022-04-28
[v7r2] Don't crash if systemCall fails

chrisburr:
BEGINRELEASENOTES

*WMS
FIX: Don't crash if systemCall fails

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Workflow/Modules/Script.py
+++ b/src/DIRAC/Workflow/Modules/Script.py
@@ -90,9 +90,9 @@ class Script(ModuleBase):
 
         remoteRunner = RemoteRunner()
         if remoteRunner.is_remote_execution():
-            outputDict = remoteRunner.execute(self.command)
+            retVal = remoteRunner.execute(self.command)
         else:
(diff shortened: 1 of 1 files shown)


src/DIRAC/Workflow/Modules/Script.py
```diff
@@ -90,20 +90,21 @@ def _executeCommand(self):
...
                 bufferLimit=self.bufferLimit,
             )
 
-        if not outputDict["OK"]:
+        if not retVal["OK"]:
             failed = True
-            self.log.error("System call execution failed:", "\n" + str(outputDict["Message"]))
-        status, stdout, stderr = outputDict["Value"][0:3]
+            self.log.error("System call execution failed:", "\n" + str(retVal["Message"]))
+            self._exitWithError(1000)
```
chrisburr, line 106: I don't see a nicer way of handling this case without major refactoring but I'm open to ideas.

fstagni (approved):

PR #6046, opened 2022-04-28
[integration] Downloading job sandboxes from the web

chrisburr:
The WebApp assumes Tornado will return the binary content from the sandbox store:

https://github.com/DIRACGrid/WebAppDIRAC/blob/26565c2c4d6368971971757127a0b56581fa7bbd/src/WebAppDIRAC/WebApp/handler/JobMonitorHandler.py#L473-L480

BEGINRELEASENOTES

*Web
FIX: Fix downloading sandboxes from the Job Manager in the WebApp

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
+++ b/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
@@ -942,7 +942,7 @@ class BaseRequestHandler(RequestHandler):
             self.finish(self.__result)
 
         # Return simple text or html
-        elif isinstance(self.__result, str):
+        elif isinstance(self.__result, (str, bytes)):
             self.finish(self.__result)
 
         # JSON


src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
```diff
@@ -942,7 +942,7 @@ async def __execute(self, *args, **kwargs):  # pylint: disable=arguments-differ
             self.finish(self.__result)
 
         # Return simple text or html
-        elif isinstance(self.__result, str):
+        elif isinstance(self.__result, (str, bytes)):
```
chaen, line 945: I think it would be better to use the `rawContent` mechanism above, which is meant for that.

chrisburr: As far as I can tell that's a property of the request not the response which feels completely wrong. This endpoint has no valid `text/*` response. (Tangentially related, isn't this what the `Accept` header is supposed to do?)

I can see why it might be helpful to workaround Python 2's binary and text data being ambigious but I see no reason to let it spread further in Python-3 only releases.

chaen: when I read your response, I almost have the feeling you have hopes to transform the existing Tornado crap into a proper http server :rofl: 

It's true that since the complete refactoring of all that, it makes less sense to have it this way (before, it was always json, except this binary case).

But then if you want to do it properly, let's make it completely. 
We get ride of whole `rawContent` business, we set the `Accept` header, and if we return `bytes`, we set the `ContentType` to `application/octet-stream`

chrisburr: Except `ContentType` to `application/octet-stream` is wrong in this case as this method returns `application/x-tar`. Using the `Accept` header would still be wrong from this method as it's not trying to negotatiate with the server. The server will only ever return a tarball or an error.

> We get ride of whole rawContent business, we set the Accept header, and if we return bytes, we set the ContentType to application/octet-stream

Sure, but we shouldn't pollute a bugfix PR with refactoring.

For that PR I'll need to look into exactly how `rawContent` is being used to see if the `Accept` header is the right thing to do.

fstagni (approved):

PR #6080, opened 2022-05-16
[8.0] fix (StompMQConnector): adapt to stomp 8

chaen:
BEGINRELEASENOTES
*Resources
FIX: Adapt StompMQConnector to Stomp 8

ENDRELEASENOTES


chrisburr:
Shall we backport this to v7r2 to make it so that DIRACOS2 works with all DIRAC versions?

chaen:
that is a relatively easy one to do, but in general I am really not in favor of backporting stuff, so we really need to have a way to freeze the version of DIRACOS2

fstagni (approved):

PR #6102, opened 2022-05-23
[integration] fix: remove decode from proxy in CloudCE

marianne013:
Did the proxy dumpAllToString return type change in integration where it seems to be a str while in v7r3 it seems to be bytes ? 
BEGINRELEASENOTES

*Resources
FIX: remove decode from proxy in CloudCE

ENDRELEASENOTES


chrisburr:
> Did the proxy dumpAllToString return type change in integration where it seems to be a str while in v7r3 it seems to be bytes ?

Yes: https://github.com/DIRACGrid/DIRAC/pull/5854

chrisburr (approved):

PR #6105, opened 2022-05-23
[v7r3] CloudEndpoint fix:  Error in the condition to get FlavorName

zhangxiaomei:
BEGINRELEASENOTES
*Resources
FIX: Error in the condition to get FlavorName
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Cloud/CloudEndpoint.py
+++ b/src/DIRAC/Resources/Cloud/CloudEndpoint.py
@@ -237,7 +237,7 @@ class CloudEndpoint(Endpoint):
         createNodeDict["image"] = image
 
         # Get the flavor object
-        if "FlavorID" in self.parameters and "FlavorName" not in self.parameters:
+        if "FlavorID" not in self.parameters and "FlavorName" in self.parameters:
             result = self.__getFlavorByName(self.parameters["FlavorName"])
             if not result["OK"]:
                 return result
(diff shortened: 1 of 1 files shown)


src/DIRAC/Resources/Cloud/CloudEndpoint.py
```diff
@@ -237,7 +237,7 @@ def createInstance(self, instanceID="", createPublicIP=True):
         createNodeDict["image"] = image
 
         # Get the flavor object
-        if "FlavorID" in self.parameters and "FlavorName" not in self.parameters:
+        if "FlavorID" not in self.parameters and "FlavorName" in self.parameters:
```
chrisburr, line 240: Would it be more readable to re-order the if conditions to be like this:
```python
        if "FlavorID" in self.parameters:
            flavor = self.__driver.ex_get_size(self.parameters["FlavorID"])
        elif "FlavorName" in self.parameters:
            result = self.__getFlavorByName(self.parameters["FlavorName"])
            if not result["OK"]:
                return result
            flavor = result["Value"]
        else:
            return S_ERROR("No flavor specified")
```

zhangxiaomei: Yes, it is better this way. Need another pull request?

sfayer (approved):
Thanks, the condition for this if statement got switched around by mistake during a refactor about 3 years ago... Fix looks fine to me.

fstagni (approved):

PR #6128, opened 2022-05-31
[8.0] Fix discovery of HTTPS methods from parent classes

chrisburr:
Using `cls.__dict__` means that methods from parent classes are ignored. Fixing it check parent classes then means that the OAuth stuff is broken as it starts exposing tornado methods like `get_secure_cookie` as `/secure_cookie/`.

BEGINRELEASENOTES

*Core
FIX: Discovery of HTTPS methods from parent classes

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/TornadoREST.py
+++ b/src/DIRAC/Core/Tornado/Server/TornadoREST.py
@@ -171,25 +171,12 @@ class TornadoREST(BaseRequestHandler):  # pylint: disable=abstract-method
         """
         urls = []
         # Look for methods that are exported
-        for mName in cls.__dict__:
-            mObj = cls.__dict__[mName]
-
-            if cls.METHOD_PREFIX and mName.startswith(cls.METHOD_PREFIX):
-                # Target methods begin with a prefix defined for all supported http methods,
-                # e.g.: def export_myMethod(self):
-                prefix = len(cls.METHOD_PREFIX)
-            elif _prefix := [
-                p for p in cls.SUPPORTED_METHODS if mName.startswith(f"{p.lower()}_")  # pylint: disable=no-member
-            ]:
-                # Target methods begin with the name of the http method,
-                # e.g.: def post_myMethod(self):
-                prefix = len(_prefix[-1]) + 1
-            else:
-                # The name of the target method must contain a special prefix
-                continue
-
-            # if the method exists we will continue
-            if callable(mObj) and (methodName := mName[prefix:]):
+        for prefix in [cls.METHOD_PREFIX] if cls.METHOD_PREFIX else cls.SUPPORTED_METHODS:
+            prefix = prefix.lower()
+            if prefix == "get":
+                raise NotImplementedError("This is fundamentally broken as tornado has methods named get_")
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/Tornado/Server/TornadoREST.py
```diff
@@ -171,25 +171,12 @@ def _pre_initialize(cls) -> list:
...
-                continue
-
-            # if the method exists we will continue
-            if callable(mObj) and (methodName := mName[prefix:]):
+        for prefix in [cls.METHOD_PREFIX] if cls.METHOD_PREFIX else cls.SUPPORTED_METHODS:
+            prefix = prefix.lower()
+            if prefix == "get_":
+                raise NotImplementedError("This is fundamentally broken as tornado has methods named get_")
+            for mName, mObj in inspect.getmembers(cls, lambda x: callable(x) and x.__name__.startswith(prefix)):
+                methodName = mName[len(prefix) :]
```
TaykYoku, line 179: IIUC `cls.SUPPORTED_METHODS` its a list of names without "_" (for ex.: ('GET', 'HEAD', 'POST', 'DELETE', 'PATCH', 'PUT', 'OPTIONS')), it means that `'GET'.lower() == "get_"` isn`t true. So it seems that exception will not work. 

I propose to leave the logic as it was, adding the exception of tornado methods. As I understand it, this will solve the purpose of this PR and will not break the existing logic.


```suggestion
        # Look for methods that are exported, except for original tornado methods, such as `get_secure_cookie`
        tornadoMethods = [name for name, cls_obj in inspect.getmembers(BaseRequestHandler, lambda x: callable(x))]

        for mName, mObj in [(name, cls_obj) for name, cls_obj in inspect.getmembers(cls, lambda x: callable(x)) if name not in tornadoMethods]:

            if cls.METHOD_PREFIX and mName.startswith(cls.METHOD_PREFIX):
                # Target methods begin with a prefix defined for all supported http methods,
                # e.g.: def export_myMethod(self):
                prefix = len(cls.METHOD_PREFIX)
            elif _prefix := [
                p for p in cls.SUPPORTED_METHODS if mName.startswith(f"{p.lower()}_")  # pylint: disable=no-member
            ]:
                # Target methods begin with the name of the http method,
                # e.g.: def post_myMethod(self):
                prefix = len(_prefix[-1]) + 1
            else:
                # The name of the target method must contain a special prefix
                continue

            # if the method exists we will continue
            if methodName := mName[prefix:]:
```

TaykYoku: @chrisburr any updates here?

chrisburr: I don't think this is an acceptable solution and I think we should instead come up with a long-term solution for 8.1.

chrisburr:
@atsareg This PR is still needed for extensions to HTTPS services to work in 8.0

atsareg:
I guess there is no conclusion on the above discussion. Are you sure this fox works actually ?

TaykYoku:
How about my comment? Are we deliberately setting exceptions and destroying the existing logic?

PR #6137, opened 2022-06-06
[8.0] [dirac-configure] add https to issuer address automatically

TaykYoku:
This PR aims to simplify the choice of authorization server for users, for example:

```
dirac-configure --login=dirac.egi.eu
```

BEGINRELEASENOTES

*Core
NEW: add https to issuer address automatically

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/scripts/dirac_configure.py
+++ b/src/DIRAC/Core/scripts/dirac_configure.py
@@ -164,7 +164,7 @@ class Params(object):
         # If the user selects the issuer, it means that there will be authorization through tokens
         os.environ["DIRAC_USE_ACCESS_TOKEN"] = "True"
         # Allow the user to enter only the server domain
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/scripts/dirac_configure.py
```diff
@@ -161,7 +161,13 @@ def setExtensions(self, optionValue):
         return DIRAC.S_OK()
 
     def setIssuer(self, optionValue):
+        # If the user selects the issuer, it means that there will be authorization through tokens
         os.environ["DIRAC_USE_ACCESS_TOKEN"] = "True"
+        # Allow the user to enter only the server domain
+        if not optionValue.startswith("http://") or not optionValue.startswith("https://"):
```
andresailer, line 167:
```suggestion
        if not optionValue.startswith(("http://", "https://")):
```
https://docs.python.org/3/library/stdtypes.html#str.startswith also takes a tuple

PR #6151, opened 2022-06-12
[integration] Replace deprecated threading.activeCount with active_count

hugovk:
`threading.activeCount` was deprecated in Python 3.10 (October 2021) and will be removed in Python 3.12 (October 2023).

Replace with `threading.active_count`, added in Python 2.6 (October 2008).

* https://docs.python.org/3.12/whatsnew/3.10.html#deprecated
* https://github.com/python/cpython/pull/25174

BEGINRELEASENOTES

*Core
FIX: Replace deprecated threading.activeCount with active_count

ENDRELEASENOTES


chrisburr (approved):
Thanks for this @hugovk!

(I'll take care of backporting this to the other branches)

PR #6155, opened 2022-06-13
[v7r3] replaced old vomses/vomsdir code to be able to handle cvmfs vomses

marianne013:
@sfayer and me think the old code is basically cargo cult code.

BEGINRELEASENOTES

*Core
FIX: Replaced old vomses/vomsdir code to be able to handle cvmfs vomses. See #6150 

ENDRELEASENOTES


marianne013:
Only locally, i.e. not in v8.

fstagni (approved):

PR #6160, opened 2022-06-14
[Integration] fix: removing VOPolicy from job description

simon-mazenoux:
remove the VOPolicy field from the job description, which is not used since #5184.


fstagni (approved):
OK. Also updated https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-8.0

PR #6174, opened 2022-06-17
[8.0] Rename __executeMethod

chrisburr:
I propose to rename these methods to make clear they're part of the public interface of this class and can be overridden in the webapp.


chaen:
they are not really public, they are to be used by derived classes only. Single underscore makes sense in my opinion, but clearly not double underscore

fstagni (approved):

TaykYoku:
I think in the description you mean renaming only `__executeMethod`

TaykYoku (approved):

PR #6180, opened 2022-06-21
[v7r2] fix: address issue 6092 where built-in metakeys were modified for MultiVO

martynia:
BEGINRELEASENOTES

*DataManagementSystem
FIX: Built-in metakeys were incorrectly modified in a MultiVO environment. This fix keeps them intact while modifying user-supplied metadata names by append a VO suffix, as before.

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py
+++ b/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py
@@ -19,14 +19,14 @@ VO_SUFFIX_SEPARATOR = "___"
 def _getMetaName(meta, credDict):
     """
     Return a fully-qualified metadata name based on client-supplied metadata name and
-    client credentials. User VO is added to the metadata passed in.
+    client credentials. User VO is added to the metadata passed in. Built-in meta keys are not affected.
 
     :param meta: metadata name
     :param credDict: client credentials
-    :return: fully-qualified metadata name
+    :return: fully-qualified metadata name for user metadata and unmodified metadata name for built-in keys.
     """
 
-    return meta + _getMetaNameSuffix(credDict)
+    return meta + _getMetaNameSuffix(credDict) if meta not in MetaQuery.FILE_STANDARD_METAKEYS else meta
 
 
 def _getMetaNameDict(metaDict, credDict):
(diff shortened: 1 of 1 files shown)


src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py
```diff
@@ -40,7 +41,10 @@ def _getMetaNameDict(metaDict, credDict):
 
     fMetaDict = {}
     for meta, value in metaDict.items():
-        fMetaDict[_getMetaName(meta, credDict)] = value
+        if meta in MetaQuery.FILE_STANDARD_METAKEYS:
+            fMetaDict[meta] = value
+        else:
+            fMetaDict[_getMetaName(meta, credDict)] = value
```
andresailer, line 47: `_getMetaName` is used in 10 places (according to github code reference finder). Is this the only place where this makes a difference? Why not implement this check for standard keys in `_getMetaName`?

martynia: A good point. This is what we found when trying to select files based on creation date in a multi VO environment. The build-in meta keys are not VO aware, so there should be no VO suffix used when searching. The method in question is is here:
[https://github.com/DIRACGrid/DIRAC/blob/6764f043334b7d90f28df8fa7b62e41e1cb835c9/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py#L183
](https://github.com/DIRACGrid/DIRAC/blob/6764f043334b7d90f28df8fa7b62e41e1cb835c9/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py#L183)

There are 4 `_getMetaNameDict` calls, 2 for files and 2 for directories. 

As you said there are multiple calls to `_getMetaName`, adding a modification there would be most like not harmful, Let me run some checks to see if its OK. Many of those calls are used for user metadata and are kept in dedicated DB tables (with a VO suffix).

martynia: Done.

andresailer (approved):

PR #6188, opened 2022-06-22
[integration] Sweep #6043 "Propagate proxy strength" 

chrisburr:
As suggested in https://github.com/DIRACGrid/DIRAC/pull/6043#issuecomment-1153998451

BEGINRELEASENOTES

*Core
CHANGE: Propagate original proxy strength when generating new proxies

ENDRELEASENOTES


chaen:
Thanks for going through the burden !

chaen (approved):

PR #6204, opened 2022-06-23
[8.0] Hackaton fix: TimeUtilities module 

rupozzi:
Fixing errors in TimeUtilites module. Adding furhter testing for new methods.

BEGINRELEASENOTES

*Core
FIX: Fixing errors, adding tests for TimeUtilities module

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/TimeUtilities.py
+++ b/src/DIRAC/Core/Utilities/TimeUtilities.py
@@ -24,6 +24,8 @@ import time
 import datetime
 import sys
 
+from DIRAC import S_OK, S_ERROR, gLogger
+
 
 # Some useful constants for time operations
 microsecond = datetime.timedelta(microseconds=1)
@@ -73,7 +75,7 @@ def timeThis(method):
                     except IndexError:
                         argsLen = ""
 
-        print("%s Exec time ===> function %r %s -> %2.2f sec" % (pre, method.__name__, argsLen, te - ts))
+        gLogger.info("%s Exec time ===> function %r %s -> %2.2f sec" % (pre, method.__name__, argsLen, te - ts))
         return result
 
     return timed
@@ -83,8 +85,14 @@ def toEpoch(dateTimeObject=None):
     """
     Get seconds since epoch
     """
-    if not dateTimeObject:
-        dateTimeObject = datetime.datetime.utcnow()
+    try:
+        # If object of date type is given, return a datetime object set to midnight
+        if isinstance(dateTimeObject, datetime.date):
+            dateTimeObject = datetime.datetime.combine(dateTimeObject, datetime.time.min)
+        if not dateTimeObject:
+            dateTimeObject = datetime.datetime.utcnow()
+    except AttributeError as e:
+        return S_ERROR(e)
     return calendar.timegm(dateTimeObject.utctimetuple())
 
 
--- a/src/DIRAC/Core/Utilities/TimeUtilities.py
+++ b/src/DIRAC/Core/Utilities/TimeUtilities.py
@@ -24,7 +24,7 @@ import time
 import datetime
 import sys
 
-from DIRAC import S_OK, S_ERROR, gLogger
+from DIRAC import gLogger
 
 
 # Some useful constants for time operations
@@ -83,30 +83,37 @@ def timeThis(method):
 
 def toEpoch(dateTimeObject=None):
     """
-    Get seconds since epoch
+    Get seconds since epoch. Accepts datetime or date objects
     """
-    try:
-        # If object of date type is given, return a datetime object set to midnight
-        if isinstance(dateTimeObject, datetime.date):
-            dateTimeObject = datetime.datetime.combine(dateTimeObject, datetime.time.min)
-        if not dateTimeObject:
-            dateTimeObject = datetime.datetime.utcnow()
-    except AttributeError as e:
-        return S_ERROR(e)
-    return calendar.timegm(dateTimeObject.utctimetuple())
+    return toEpochMilliSeconds(dateTimeObject) // 1000
 
 
 def toEpochMilliSeconds(dateTimeObject=None):
     """
     Get milliseconds since epoch
     """
-    return toEpoch(dateTimeObject) * 1000
+    if dateTimeObject is None:
+        dateTimeObject = datetime.datetime.utcnow()
+    if dateTimeObject.resolution == datetime.timedelta(days=1):
+        # Add time information corresponding to midnight UTC if it's a datetime.date
+        dateTimeObject = datetime.datetime.combine(
+            dateTimeObject, datetime.time.min.replace(tzinfo=datetime.timezone.utc)
+        )
+    posixTime = dateTimeObject.replace(tzinfo=datetime.timezone.utc).timestamp()
+    return int(posixTime * 1000)
 
 
 def fromEpoch(epoch):
     """
     Get datetime object from epoch
     """
+    # Check if the timestamp is in milliseconds
+    if epoch > 10**17:  # nanoseconds
+        epoch /= 1000**3
+    elif epoch > 10**14:  # microseconds
+        epoch /= 1000**2
+    elif epoch > 10**11:  # milliseconds
+        epoch /= 1000
     return datetime.datetime.utcfromtimestamp(epoch)
 
 
@@ -124,10 +131,7 @@ def toString(myDate=None):
       [hour]:[min]:[sec]:[microsec]
       where min, sec, microsec are always positive integers and hour carries the sign.
     """
-    if isinstance(myDate, datetime.datetime):
-        return str(myDate)
-
-    elif isinstance(myDate, datetime.date):
+    if isinstance(myDate, datetime.date):
         return str(myDate)
 
     elif isinstance(myDate, datetime.time):
--- a/tests/Integration/Core/Test_TimeUtilities.py
+++ b/tests/Integration/Core/Test_TimeUtilities.py
@@ -1,0 +1,52 @@
+"""Test for the TimeUtilities module
+"""
+import pytest
+
+import datetime, pytz
+
+from DIRAC.Core.Utilities import TimeUtilities
+
+americanTimezone = pytz.timezone("America/Los_Angeles")
+japaneseTimezone = pytz.timezone("Asia/Tokyo")
+euroTimezone = pytz.timezone("Europe/Rome")
+
+
+@pytest.mark.parametrize(
+    "dt, expected, expectedMillis, timezone",
+    [
+        # Try with a datetime type object
+        (datetime.datetime(1997, 6, 12, 2, 30, 0, 0), 866082600, None, None),
+        # Now with a date type object, it should return the timestamp of the date with the time set to midnight
+        (datetime.date(1997, 6, 12), 866073600, None, None),
+        # Try the same with Milliseconds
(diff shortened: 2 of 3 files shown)


src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -83,8 +85,14 @@ def toEpoch(dateTimeObject=None):
     """
     Get seconds since epoch
     """
-    if not dateTimeObject:
-        dateTimeObject = datetime.datetime.utcnow()
+    try:
```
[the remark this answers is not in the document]
rupozzi: Will make one. Shall I do it in a new PR?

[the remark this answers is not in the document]
fstagni: no, you can do it here.

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -24,6 +24,8 @@
 import datetime
 import sys
 
+from DIRAC import S_OK, S_ERROR, gLogger
```
fstagni, line 27: Do you need ` S_OK, S_ERROR`?

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -73,19 +75,25 @@ def timed(*args, **kw):
...
     """
-    Get seconds since epoch
+    Get seconds since epoch. Accepts datetime or date objects
     """
-    if not dateTimeObject:
-        dateTimeObject = datetime.datetime.utcnow()
-    return calendar.timegm(dateTimeObject.utctimetuple())
+    try:
+        # If object of date type is given, return a datetime object set to midnight
+        if type(dateTimeObject) is datetime.date:
```
fstagni, line 90:
```suggestion
        if isinstance(dateTimeObject, datetime.date):
```

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -99,6 +107,9 @@ def fromEpoch(epoch):
     """
     Get datetime object from epoch
     """
+    # Check if the timestamp is in milliseconds
+    if len(str(epoch)) == 13:
```
fstagni, line 111: Mmm... I don't think I like it enormously

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -116,13 +127,13 @@ def toString(myDate=None):
       [hour]:[min]:[sec]:[microsec]
       where min, sec, microsec are always positive integers and hour carries the sign.
     """
-    if isinstance(myDate, datetime.datetime):
+    if type(myDate) is datetime.datetime:
```
fstagni, line 130: why not `isinstance`?

tests/Integration/Core/Test_TimeUtilities.py
```diff
@@ -0,0 +1,41 @@
...
+    # Let us try now with a date type object
+    dt = datetime.date(1997, 6, 12)
+    res = TimeUtilities.toEpoch(dt)
+    # it should return the timestamp of the date with the time set to midnight
+    assert res == 866073600
+
+    # Now with another type
+    dt = datetime.time()
+    res = TimeUtilities.toEpoch(dt)
+    assert not res["OK"]
```
fstagni, line 25: why this not?

src/DIRAC/Core/Utilities/TimeUtilities.py
rupozzi: This unfortunately does not work as a datetime.date type it returns `True` for both `isinstance(dateTimeObject, datetime.date)` and `isinstance(dateTimeObject, datetime.datetime)`

src/DIRAC/Core/Utilities/TimeUtilities.py
rupozzi: `
In [9]: dt = datetime.datetime.utcnow()

In [10]: isinstance(dt, datetime.date)
Out[10]: True

In [11]: isinstance(dt, datetime.datetime)
Out[11]: True
`

src/DIRAC/Core/Utilities/TimeUtilities.py
rupozzi: @chrisburr mainly we want to double check that the `dateTypeObject` is not a `datetime.date` type, so that would still return `True` for both date and datetime objects no?

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -73,19 +75,25 @@ def timed(*args, **kw):
...
-        dateTimeObject = datetime.datetime.utcnow()
-    return calendar.timegm(dateTimeObject.utctimetuple())
+    try:
+        # If object of date type is given, return a datetime object set to midnight
+        if type(dateTimeObject) is datetime.date:
+            dateTimeObject = datetime.datetime.combine(dateTimeObject, datetime.time.min)
+        if not dateTimeObject:
+            dateTimeObject = datetime.datetime.utcnow()
+        return calendar.timegm(dateTimeObject.utctimetuple())
+    except AttributeError as e:
```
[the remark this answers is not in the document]
rupozzi: Because in case a wrong datetime type is given, the exception would  be caught

src/DIRAC/Core/Utilities/TimeUtilities.py
chrisburr: I think we should at least do this which gives us from 1973 to 5138:
```suggestion
    if epoch > 10**17:  # nanoseconds
        epoch /= 1000**3
    elif epoch > 10**14:  # microseconds
        epoch /= 1000**2
    elif epoch > 10**11:  # milliseconds
        epoch /= 1000
```

Please add unit tests to make sure that `datetime(5100, 1, 1).timestamp()` and `datetime(1974, 1, 1).timestamp()` are always converted correctly even when multiplied by 1000, 1000000 and 1000000000.

src/DIRAC/Core/Utilities/TimeUtilities.py
chrisburr: I think this can be:

```
-if type(myDate) is datetime.datetime:
-    return str(myDate)
-elif type(myDate) is datetime.date:
+if isinstance(myDate, datetime.date):
```

src/DIRAC/Core/Utilities/TimeUtilities.py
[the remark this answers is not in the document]
chrisburr: What would be an example of a wrong "datetime type"?

Also this function doesn't return a `S_OK` structure so returning `S_ERROR` doesn't make any sense as it'll never be checked for `retVal["OK"]`.

tests/Integration/Core/Test_TimeUtilities.py
chrisburr: Calling this function with `datetime.time()` makes no sense so it's okay for it to crash.

src/DIRAC/Core/Utilities/TimeUtilities.py
rupozzi: Yes we want that specific condition for only datetime.date objects. That if statement doesnt filter them out.
```python
In [19]: dateTimeObject = datetime.datetime.utcnow()

In [20]: if not isinstance(dateTimeObject, datetime.time):
    ...:     print("Type: ", type(dateTimeObject))
    ...:     print("True")
    ...: 
Type:  <class 'datetime.datetime'>
True

In [21]: dateTimeObject = datetime.date.today()

In [22]: if not isinstance(dateTimeObject, datetime.time):
    ...:     print("Type: ", type(dateTimeObject))
    ...:     print("True")
    ...: 
Type:  <class 'datetime.date'>
True
```

tests/Integration/Core/Test_TimeUtilities.py
rupozzi: Do we want the `toEpoch()` to take in also `datetime.time` objects?

src/DIRAC/Core/Utilities/TimeUtilities.py
chrisburr: Ah I see the problem now, let me have a think...

tests/Integration/Core/Test_TimeUtilities.py
chrisburr: What would that even mean?

`toEpoch` returns the number of seconds since 1970-01-01 so a time without a date doesn't make sense for this function.

src/DIRAC/Core/Utilities/TimeUtilities.py
chrisburr: How about:
```python
def toEpoch(dateTimeObject=None):
    """
    Get seconds since epoch. Accepts datetime or date objects
    """
    return toEpochMilliSeconds(dateTimeObject) // 1000


def toEpochMilliSeconds(dateTimeObject=None):
    """
    Get milliseconds since epoch
    """
    # Is this supposed to handle dateTimeObject == None? Else why does it have a default value?
    # If it should then you need:
    # if dateTimeObject == None:
    #     dateTimeObject = datetime.datetime.now()
    if dateTimeObject.resolution == datetime.timedelta(days=1):
        # Add time information corresponding to midnight UTC if it's a datetime.date
        dateTimeObject = datetime.datetime.combine(dateTimeObject, datetime.time.min.replace(tzinfo=datetime.timezone.utc))
    posixTime = dateTimeObject.replace(tzinfo=datetime.timezone.utc).timestamp()
    return int(posixTime * 1000)
```

chrisburr: A few further comments:
* `toEpoch` should use `toEpochMilliSeconds` rather than the other way around to avoid loosing precision and rounding to the nearest 1000 milliseconds
* The `dateTimeObject == None` should either be handled or reoved from the signature
* Does this need to handle timezone naive datetimes? Or just timezone aware ones? Either way it should be tested.

src/DIRAC/Core/Utilities/TimeUtilities.py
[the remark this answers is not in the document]
rupozzi: Yepp rookie mistake there!

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -73,32 +75,45 @@ def timed(*args, **kw):
...
-    return calendar.timegm(dateTimeObject.utctimetuple())
+    return toEpochMilliSeconds(dateTimeObject) // 1000
 
 
 def toEpochMilliSeconds(dateTimeObject=None):
     """
     Get milliseconds since epoch
     """
-    return toEpoch(dateTimeObject) * 1000
+    if dateTimeObject == None:
```
fstagni, line 95:
```suggestion
    if dateTimeObject is None:
```

src/DIRAC/Core/Utilities/TimeUtilities.py
```diff
@@ -73,32 +75,45 @@ def timed(*args, **kw):
...
+        dateTimeObject = datetime.datetime.utcnow()
+    if dateTimeObject.resolution == datetime.timedelta(days=1):
```
fstagni, line 97: What is this for? In case it's a `date`?

rupozzi: Yes exactly

chrisburr (approved):

PR #6252, opened 2022-07-11
[v7r3] Support Site/CE specific container bind mounts

chrisburr:
Some sites use a local networked filesystem to provide access to data and therefore extra paths need to be bind bounted inside containers. This PR expands the ContainerBind option to be available at the Site and CE level rather than only the computing element type.

BEGINRELEASENOTES

*WorkloadManagement
NEW: ContainerBind option can be set on the per-site and per-CE level

ENDRELEASENOTES


chrisburr:
This is failing because of rucio, I started fixing it in https://github.com/DIRACGrid/DIRAC/pull/6254 however the code itself is broken so I'm not in a position to look further.

chaen (approved):

PR #6254, opened 2022-07-11
 [v7r2] Force the use of last rucio release with support for Python 2

chrisburr:
Sweep of https://github.com/DIRACGrid/DIRAC/pull/6253

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -79,4 +79,4 @@ dependencies:
     - git+https://github.com/DIRACGrid/tornado.git@iostreamConfigurable
     # This is an extension of Tornado to use M2Crypto
     - git+https://github.com/DIRACGrid/tornado_m2crypto
-    - rucio-clients >= 1.25.6
+    - rucio-clients==1.26.*


martynia:
The `default_extract `options is necessary for client config file being optional. This was added to Rucio in 1.27, the first multiVO Dirac-Rucio integration. Since 1.27 does not work with Python 2, I don't think there is an easy solution for this.

chrisburr:
> 1.27 does not work with Python 2

rucio-clients 1.27 does work with Python 2, it's just not the in DIRACOS so Python 2 based DIRAC installations don't have access to it.

martynia:
Hi,

default_extract options is necessary for client config file being optional. This was added to Rucio in 1.27, the first multiVO Dirac-Rucio integration.

I understand that you cannot enforce 1.27 for Python 2, can you ? (Possibly because Rucio have dropped Python 2 support)

Cheers, JM

chrisburr:
@martynia I don't understand what you're asking here.

I thought we had resolved this months ago by deciding that we let rucio support be broken when using DIRAC with Python 2? If there is something new can you open an issue.

martynia:
I'm not asking for anything ;-) . I'm just explaining that we need Rucio 1.27 and Python 3 requirement is a side effect of this. Everything is fine.

PR #6265, opened 2022-07-14
[v7r2] Fix FC chgrp on multiple LFNs

sfayer:
Hi,

I found that running the FC changePathGroup function only works if one LFN is given: If multiple are specified only the first is updated and the rest return an error as the loop variable "res" gets overwritten.

Regards,
Simon

BEGINRELEASENOTES
*DataManagement
FIX: Fix FC change{group,owner,mode} for multiple LFNs
ENDRELEASENOTES


andresailer:
Very nice! Can you change the same thing in changeFileOwner and changeFileMode

sfayer:
Sure, I've fixed the other functions too now & updated the release notes to match.

andresailer (approved):

chrisburr (approved):

PR #6307, opened 2022-08-11
[sweep:integration] Don't notify users with expired proxies

sfayer:
Sweep #6305 `Don'"'"'t notify users with expired proxies` to `integration'.

Not sure why this was a sweep failure, the patch applied cleanly, so I guess some kind of problem with the sweeper itself?


chrisburr:
> Not sure why this was a sweep failure, the patch applied cleanly, so I guess some kind of problem with the sweeper itself?

I think it has problems with single quotes being badly escaped. I thought I'd fixed it but maybe not.

chrisburr:
The failures are due to a change in one of the container images used for testing and is fixed by #6312.

chrisburr (approved):

PR #6327, opened 2022-08-16
[8.0] Replace deprecated Subquery coercion with explicit select

chrisburr:
Fixes this deprecation notice:

```python
src/DIRAC/DataManagementSystem/DB/test/Test_FTS3DB.py::test_raceCondition
  /Users/cburr/Development/DIRAC-dev/DIRAC/src/DIRAC/DataManagementSystem/DB/FTS3DB.py:550: SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly
    .filter(~FTS3Operation.operationID.in_(opIDsWithJobAssigned))
```

Please review carefully in case I've done something stupid 😄


chaen (approved):
I compared the SQL output, and it's the same (it's actually even better) :-)

PR #6352, opened 2022-08-31
[v7r3] Add a gfal2 context manager as well as a removal timeout for echo

chaen:
BEGINRELEASENOTES

*Resources
NEW: Add a setGfalSetting decorator 
CHANGE: add a 30s timeout for removal at RAL

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Storage/EchoStorage.py
+++ b/src/DIRAC/Resources/Storage/EchoStorage.py
@@ -10,12 +10,15 @@ import time
 from timeit import default_timer
 
 # from DIRAC
-from DIRAC.Resources.Storage.GFAL2_StorageBase import GFAL2_StorageBase
+from DIRAC.Resources.Storage.GFAL2_StorageBase import GFAL2_StorageBase, setGfalSetting
 from DIRAC import gLogger, S_ERROR, S_OK
 
 # Duration in sec of a removal from which we start throttling
 # The value is empirical
 REMOVAL_DURATION_THROTTLE_LIMIT = 3
+
+# Timeout for the removal operation
+REMOVAL_TIMEOUT = 20
 
 
 class EchoStorage(GFAL2_StorageBase):
@@ -160,7 +163,11 @@ class EchoStorage(GFAL2_StorageBase):
         """
 
         startTime = default_timer()
-        res = super(EchoStorage, self)._removeSingleFile(path)
+        with setGfalSetting(self.ctx, "CORE", "NAMESPACE_TIMEOUT", REMOVAL_TIMEOUT):
+            # Because HTTP Plugin does not read the CORE:NAMESPACE_TIMEOUT as it should
+            # I also specify it here
+            with setGfalSetting(self.ctx, "HTTP PLUGIN", "OPERATION_TIMEOUT", REMOVAL_TIMEOUT):
+                res = super(EchoStorage, self)._removeSingleFile(path)
         duration = default_timer() - startTime
 
         # If it took too long, we sleep for a bit
--- a/src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
+++ b/src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
@@ -29,6 +29,7 @@ import six
 import os
 import datetime
 import errno
+from contextlib import contextmanager
 from stat import S_ISREG, S_ISDIR, S_IXUSR, S_IRUSR, S_IWUSR, S_IRWXG, S_IRWXU, S_IRWXO
 
 import gfal2  # pylint: disable=import-error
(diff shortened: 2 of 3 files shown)


src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -50,6 +51,54 @@
     ECOMM = 70
 
 
+@contextmanager
+def setGfalSettings(ctx, pluginName, optionName, optionValue):
```
chrisburr, line 55: Would it be common to set multiple settings at the same time? If so would this signature make more sense?

```python
def setGfalSettings(ctx, pluginName, **options):
    for optionName, optionValue in options.items():
        ...
```

src/DIRAC/Resources/Storage/EchoStorage.py
```diff
@@ -160,7 +163,11 @@ def _removeSingleFile(self, path):
         """
 
         startTime = default_timer()
-        res = super(EchoStorage, self)._removeSingleFile(path)
+        with setGfalSettings(self.ctx, "CORE", "NAMESPACE_TIMEOUT", REMOVAL_TIMEOUT):
```
andresailer, line 166: This should also work:
```python
with (setGfalSettings(...),
      setGfalSettings(...)):
   pass
```

chrisburr: Only in Python 3.10+: https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers

(3.9 optionally has this feature but is targetting v7r3 and needs to work with Python 2.7)

andresailer: weird, this is claimed as a python3.10 feature, but seems to work in python 3.9.12 as well
```python
with (open('/dev/null') as foo,
      open('/dev/zero') as bar):
      print("baz!")
```
But not python 2.7, so 
```python
with open('/dev/null') as foo, \
     open('/dev/zero') as bar:
      print("baz!")
```

chrisburr: In Python 3.9 it's a build flag/environment variable to control it. I don't know how widely enabled it is outside of conda-forge.

```python
$ ipython
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:33)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import tempfile
   ...: with (tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile()):
   ...:     pass
   ...:
   
$ PYTHONOLDPARSER=x ipython
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:33)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.
In [1]: import tempfile
   ...: with (tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile()):
   ...:     pass
   ...:
Input In [1], in <cell line: 2>()
      1 import tempfile
----> 2 with (tempfile.NamedTemporaryFile(), tempfile.NamedTemporaryFile()):
      3     pass

AttributeError: __enter__
```

chrisburr: Apprently it can be expected to be enabled everywhere as black will reformat to modernise the `\` automatically if it's targetting Python 3.9: https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#using-backslashes-for-with-statements

fstagni (dismissed):
I approve as I don't see obvious issues, but I am not the most expert on this part of code.

src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
```diff
@@ -50,6 +51,54 @@
...
+        _getter = ctx.get_opt_boolean
+    elif isinstance(optionValue, int):
+        _setter = ctx.set_opt_integer
+        _getter = ctx.get_opt_integer
+    elif isinstance(optionValue, str):
+        _setter = ctx.set_opt_string
+        _getter = ctx.get_opt_string
+    elif isinstance(optionValue, list):
+        _setter = ctx.set_opt_string_list
+        _getter = ctx.get_opt_string_list
```
chrisburr, line 85:
```suggestion
        _getter = ctx.get_opt_string_list
    else:
        raise NotImplementedError("Unknown option type %s" % type(optionValue))
```

chaen: done

chrisburr (approved):

PR #6386, opened 2022-09-27
[8.0] Remove Monitoring DataOperationPlotter

chaen:
The plotting of data operation just does not work. 

The data in ES is correct, yet when comparing Accounting and Monitoring, you get non sense

image

Worse even: plotting the same thing but with a different grouping gives a different total !!!!

image

So clearly, the plotting is broken. And I'd rather put effort into having nice kibana/grafana dashboard. 

BEGINRELEASENOTES
*Monitoring
CHANGE: Remove DataOperationPlotter

ENDRELEASENOTES


chrisburr:
I guess this also needs an associated WebAppDIRAC PR?

fstagni:
The rationale behind having DataOperation in Monitoring plotted through the "old Accounting plotters" is that what was already existing (e.g. WMSHistory, DataOperation, PilotSubmission) should have kept the plotters done also via the "old Accounting plotters", so to allow a smooth transition between the MySQL backend and the ES backed -- basically replicating what was done for the WMSHistory.

chaen:
I understand. But it is completely broken. And getting something like that in 2h seems to me a much better time investment

image

fstagni:
Let's discuss this on the next BiLD meeting.

chaen:
A few example of working dashboards, just to feed the discussion, should it suffer any argument whatsoever...
Screenshot from 2022-09-29 17-06-02
Screenshot from 2022-09-29 17-06-30
Screenshot from 2022-09-29 17-06-41
Screenshot from 2022-09-29 17-07-24
Screenshot from 2022-09-29 17-07-43

chaen:
Matching WebApp request in https://github.com/DIRACGrid/WebAppDIRAC/pull/692
Once we use these dashboards in prod, I'll make them better and add some docs

fstagni (approved):

PR #6395, opened 2022-09-30
[v7r3] Use safer mode for grid-security directories

sfayer:
Hi,

It turns out the xrootd library checks the permissions on the CA dir and possibly others when connecting to some (but bizarrely not all) SEs... If DIRAC is installed on a cluster with umask 0002, you get rwxrwxr-x permissions on the certificates, vomses & vomdir directories which result in the cryptic error message "Network dropped connection on reset" from GFAL when trying to to read from those SEs.

I believe a umask of 0002 is the default on clusters with older RHEL versions and newer ones where private groups are used (e.g. uid = default gid): I'm a little surprised we haven't run into this before.

I'm slightly of the opinion that the libraries shouldn't do checks on the permissions, but perhaps it's just easier to patch it here?

Regards,
Simon

BEGINRELEASENOTES
*Core
FIX: Use safer mode for grid-security directories
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/File.py
+++ b/src/DIRAC/Core/Utilities/File.py
@@ -30,12 +30,19 @@ SIZE_UNIT_CONVERSION = {
 }
 
 
-def mkDir(path):
-    """Emulate 'mkdir -p path' (if path exists already, don't raise an exception)"""
+def mkDir(path, mode=None):
+    """Emulate 'mkdir -p path' (if path exists already, don't raise an exception)
+
+    :param str path: directory hierarchy to create
+    :param int mode: Use this mode as the mode for new directories, use python default if None.
+    """
     try:
         if os.path.isdir(path):
             return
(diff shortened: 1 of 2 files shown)


src/DIRAC/Core/Utilities/File.py
```diff
@@ -30,12 +30,17 @@
 }
 
 
-def mkDir(path):
-    """Emulate 'mkdir -p path' (if path exists already, don't raise an exception)"""
+def mkDir(path, mode=None):
+    """Emulate 'mkdir -p path' (if path exists already, don't raise an exception)
+       Use mode as the mode for new directories, use python default if None.
```
andresailer:
```suggestion

    :param str path: directory hierarchy to create
    :param int mode: Use this mode as the mode for new directories, use python default if None.
```

fstagni (approved):

PR #6402, opened 2022-10-06
[v7r3] Don't write SQL when drunk

chaen:
That join here is totally useless and slows down queries by an order of magnitude...

BEGINRELEASENOTES
*DMS
FIX: Faster query to list directory for DFC LHCb managers

ENDRELEASENOTES


chrisburr:
@chaen I presume this is ready?

chrisburr (approved):

chrisburr:
This is waiting on @chaen applying it in prod to make sure all is well.

chaen:
applied and happy

andresailer (approved):

PR #6580, opened 2022-11-24

atsareg:
This PR adds possibility to submit pilots with tokens. The tokens are obtained by SiteDirectors via a standard mechanism from the TokenManager. In order to use the oidc-agent instead of the OAuth2 flow, a new OidcAgentIdProvider is added. This can be useful in some environments, also for debugging. Using OAuth2 IdProviders with pilot tokens generated from the long tokens in the TokenDB are also supported
  The oidc-agent is supposed to have a client created with the name of the user submitting pilots and having necessary compute scopes.    

  The following ComputingElement should be enabled with the tokens: 

- [x] HTCondorCE
- [x] ARC/ARC6
- [ ] AREX

  The case of CloudComputingElement is not addressed specifically as the case can be solved by using Application Credentials authentication mechanism where Application Credentials can be created after authentication with OAuth2 tokens. 

BEGINRELEASENOTES

*Framework

*WorkloadManagement

*Resources
NEW: HTCondorComputingElement and ARC(6)ComputingElement enabled to for job operations with tokens

*Configuration
CHANGE: Registry - IdProvider can be defined either on VO or Group level

ENDRELEASENOTES


marianne013:
@atsareg If there is an issue with the CloudComputingElement (selecting projects in Openstack), could you please make a separate issue for Simon ? We'll not port VMDIRAC, we'd consider this obsolete.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -173,6 +173,17 @@ def __writeSub(self, executable, nJobs, location, processors):
 
         executable = os.path.join(self.workingDirectory, executable)
 
+        useCredentials = ""
```
marianne013: Should be `useCredentials = "use_x509userproxy = true"` as otherwise if there isn't a token the pilot will fail with "no proxy found".

atsareg: Yes, fixed

atsareg:
@marianne013: There is indeed an issue if I want to specify both domain and a project where there are several projects available to me in the domain. However, I have a workaround (or it was foreseen like that) in using the Application Credentials. In fact, we can standardize on using the Application Credentials as the main mechanism for the authentication to a cloud site (Openstack). In this case, with the tokens one will only need once to access the site to define AppCred and then we forget the tokens. So, there is no need to code in token based access into the CloudComputingElement.

sfayer:
> @marianne013: There is indeed an issue if I want to specify both domain and a project where there are several projects available to me in the domain. However, I have a workaround (or it was foreseen like that) in using the Application Credentials. In fact, we can standardize on using the Application Credentials as the main mechanism for the authentication to a cloud site (Openstack). In this case, with the tokens one will only need once to access the site to define AppCred and then we forget the tokens. So, there is no need to code in token based access into the CloudComputingElement.

You can use `Driver_ex_tenant_name` option on the CE (or queue) to select the project if your user has access to more than one (and you're using an authentication method that doesn't bind to a specific one). I think the domain is probably mapped to region in openstack? That can be selected with `Driver_ex_force_service_region`.

Regards,
Simon

atsareg:
@sfayer: in the libcloud Openstack driver for the oidc_access_token identification method the tenant_name is used to actually contain the protocol name: 
[https://github.com/apache/libcloud/blob/650d0e1293db8d3576f45e06f3538280c04cb0b8/libcloud/common/openstack_identity.py#L1786](url) to get first an unscoped token.
And then the first available project is chosen if the project scope is requested. As a result I can not choose the necessary project in all the cases. This is not probably an overlook by the Openstack developers but rather the way how EGI Fedcloud sites are instructed to implement access by tokens.

src/DIRAC/ConfigurationSystem/Client/Helpers/Registry.py
```diff
@@ -430,7 +430,7 @@ def getIdPForGroup(group):
 
     :return: str
     """
-    return getVOOption(getVOForGroup(group), "IdP")
+    return getGroupOption(group, "IdProvider") or getVOOption(getVOForGroup(group), "IdProvider")
```
aldbr: I guess the change from "IdP" to "IdProvider" should also be updated in the documentation:
- https://dirac.readthedocs.io/en/rel-v8r0/AdministratorGuide/Configuration/ExampleConfig.html

aldbr: Why one would declare the `IdProvider` from a group section?
I understand that it brings flexibility, but isn't it too much?

Let's say we have `VO1_user` and `VO1_pilot` bound to `VO1`. 
IIUC in this context, we could have `IdProvider = WLCG_IAM` in `VO1_user` and `IdProvider = EGI_CheckIn` in `VO1_pilot`, couldn't we?

src/DIRAC/FrameworkSystem/Client/TokenManagerClient.py
```diff
@@ -86,5 +86,8 @@ def getToken(
                 idpObj.revokeToken(token["refresh_token"])
 
         return self.executeRPC(
-            username, userGroup, scope, audience, identityProvider, requiredTimeLeft, call="getToken"
+            userName, userGroup, scope, audience, identityProvider, requiredTimeLeft, call="getToken"
         )
+
+
+gTokenManager = TokenManagerClient()
```
aldbr, line 90: Just out of curiosity, why don't we just declare `getToken()` as a class method instead of creating a `gTokenManager` instance?

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
```diff
@@ -241,9 +241,31 @@ def export_getToken(
...
 
+        # No token in the cache, try getting
+        kwargs = {
+            "userName": userName,
+            "timeLeft": requiredTimeLeft,
+            "group": userGroup,
+            "scope": scope,
+            "audience": audience,
+        }
+        result = idpObj.getToken(**kwargs)
```
aldbr, line 252: I understand that this is needed here but I don't get why it was not here before.
Do you have any idea?

src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
```diff
@@ -1201,6 +1223,21 @@ def _updatePilotStatusPerQueue(self, queue, proxy):
...
+                return result
+            userName = result["Value"]
+            result = gTokenManager.getToken(
+                userName=userName,
+                userGroup=self.pilotGroup,
+                requiredTimeLeft=3600,
+            )
+            if not result["OK"]:
+                return result
+            ce.setToken(result["Value"], 3500)
```
aldbr: We could probably make a method for that? Because it is repeated twice the same way

src/DIRAC/Resources/IdProvider/IdProvider.py
```diff
@@ -26,3 +33,27 @@ def setParameters(self, parameters):
         """
         self.parameters = parameters
         self.name = parameters.get("ProviderName")
+
+    def getGroupScopes(self, group: str) -> list:
```
aldbr, line 37: Should we move `getGroupScopes()` and `getScopeGroups()` here? Seems like `scope_to_list()` is a method bound to oauth2.

More generally, do we need to have a `IdProvider` class?
Can we currently have IdProviders not depending on OAuth2?

src/DIRAC/Resources/Computing/ARC6ComputingElement.py
```diff
@@ -141,6 +141,9 @@ def submitJob(self, executableFile, proxy, numberOfJobs=1):
                     break
 
                 # Submit the job
+                # Set up the token in the environment
+                if self.token:
+                    os.environ["BEARER_TOKEN"] = self.token["access_token"]
```
aldbr, line 157: Is there a case where we need more than the `access_token` field here?
Why not `self.token` == `access_token`?

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -237,6 +250,32 @@ def _reset(self):
         self.log.debug("Remote scheduler option:", self.remoteScheddOptions)
         return S_OK()
 
+    def _executeCondorCommand(self, cmd, tokenFile=None, keepTokenFile=False):
```
aldbr: I am not sure to understand this method.
For instance, if we want to get the status of 50 pilots: `getJobStatus()`:
- we have `tokenFile = None`
- because it is `None`, it will create 50 token files which are not returned by the method
- None of them will be deleted because `tokenFile` is still `None` at the end.

Am I missing something here?

I would have created a `_prepareTokenFile(): str (tokenFile)` method that I would have used only `self.token` exists and I would have deleted it manually.

aldbr (commented):
FYI, for the `AREXCE`, I added the following line 3 months ago: https://github.com/DIRACGrid/DIRAC/blob/integration/src/DIRAC/Resources/Computing/AREXComputingElement.py#L68

This was manually tested during the "ARC/HTCondor & tokens hackathon" in September 2022.

src/DIRAC/ConfigurationSystem/Client/Helpers/Registry.py
atsareg: After the discussion this was rolled back

src/DIRAC/FrameworkSystem/Client/TokenManagerClient.py
atsareg: It should keep a global token cache. This is the pattern we use for the gProxyManager, for example and it worked so far.

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
atsareg: No, this was introduced for an IdProviders that allow specifically for getting tokens without a user authentication flow. The only one we have is the OidcAgentIdProvider.

src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
atsareg: Yes, can be improved

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: Well, in theory we can have other types of providers, based on SAML for example

aldbr: Alright, I understand the need to get a flexible and generic architecture, but at the same time, unless I am not aware, we only have OAuth2 use cases for now.
Are we sure that SAML assertions work similarly to OAuth2 tokens? I read that SAML2 "does not help with API authorization" for instance.
Are we sure that this architecture will be enough to support new types of providers?


At least, I think that `getGroupScopes()` and `getScopeGroups()` should be defined in `IdProvider`, empty, and should be overridden in `OAuth2IdProvider` with the current content of the methods (since `scope_to_list()` is bound to OAuth2).
What do you think?

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
atsareg: Yes, it looks to be a flaw in the logic. I have updated this part of the code.

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: I would say that the scope_to_list authlib function is quite a generic helper. If ventually there will be other non-OAuth2 type of providers, this function will be still valid.

src/DIRAC/Resources/Computing/ARC6ComputingElement.py
atsareg: Most likely there will not be such cases. The TokenManager(Client) returns an OAuth2Token object. So, it is natural to pass it to the CE object as it is. May be later there will be cases where some extra information can be used from there.

atsareg:
The PR was screwed while rebasing and will be redone

PR #6643, opened 2022-12-19
[v7r3] Backport OpenSSL 3 support

chrisburr:
Leaving as draft until this has been tested in certificiation with v8.0.x

BEGINRELEASENOTES

*Core
NEW: Support OpenSSL 3.0.x

ENDRELEASENOTES

### DIFF
--- a/environment-py3.yml
+++ b/environment-py3.yml
@@ -22,7 +22,7 @@ dependencies:
   - fts3
   - future
   - gitpython >=2.1.0
-  - m2crypto >=0.36,!=0.38.0
+  - m2crypto >=0.38.0
   - matplotlib
   - numpy
   - pexpect >=4.0.1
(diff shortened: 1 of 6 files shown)


chaen:
Tested OK with python 3. Would be worth checking with python 2 if somebody still has an instance running python 2 ? @andresailer @marianne013  ?

andresailer:
I will not have 7.3 run with python 2

marianne013:
We don't have python 2 here either.

chaen:
OK, then last potential people would be @hmiyake @iueda and @zhangxiaomei 
Otherwise, good to go

zhangxiaomei:
We are running with python3 only

hmiyake:
Could you give us a couple of days? My initial test on client showed positive result but we would test the code on our certification server.

chrisburr:
> Could you give us a couple of days? My initial test on client showed positive result but we would test the code on our certification server.

That's fine, I doubt we'll be releasing anything until the new year.

hmiyake:
This patch is working well on our v7r3 python2 environment. Please go ahead.

fstagni (approved):

PR #6661, opened 2023-01-10
[v7r3] Use full jobID in AREX jobsToCancel

sfayer:
Hi,

I've noticed that the AREX jobsToCancel list was being built out of ARC IDs rather than full DIRAC job IDs... I think the killJob function needs the latter. I've not seen this happen, but it appears it'll just result in a warning being logged for each ID, but the fewer spurious warnings we generate the better!

Regards,
Simon

BEGINRELEASENOTES
*Resources
FIX: Use full jobID in AREX jobsToCancel
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -321,6 +321,42 @@ class AREXComputingElement(ARCComputingElement):
             self.log.verbose("Input correctly uploaded", fileToSubmit)
         return S_OK()
 
+    def _killJob(self, arcJobList):
+        """Kill the specified jobs
+
+        :param list arcJobList: list of ARC Job IDs
+        """
+        if not self.session:
+            return S_ERROR("REST interface not initialised. Cannot kill jobs.")
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("Failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # List of jobs in json format for the REST query
+        jobsJson = {"job": [{"id": job} for job in arcJobList]}
+
+        # Prepare the command
+        params = {"action": "kill"}
+        query = self._urlJoin("jobs")
+
+        # Killing jobs should be fast
+        response = self.session.post(
+            query,
+            data=json.dumps(jobsJson),
+            headers=self.headers,
+            params=params,
+            timeout=self.arcRESTTimeout,
+        )
+        if not response.ok:
+            return S_ERROR("Failed to kill all these jobs: %s %s" % (response.status_code, response.reason))
+
+        self.log.debug("Successfully deleted jobs")
+        return S_OK()
+
     def submitJob(self, executableFile, proxy, numberOfJobs=1, inputs=None, outputs=None):
         """Method to submit job
         Assume that the ARC queues are always of the format nordugrid-<batchSystem>-<queue>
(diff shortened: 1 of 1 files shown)


fstagni (dismissed):

aldbr:
Indeed, `killJob()` is expecting a DIRAC job ID, but since we already have the ARC job ID, we can avoid one operation.
`DIRACToArcID()` will just return the ARC job ID passed in parameter.
But ok anyway this is minor.

Maybe we could also change the level of the logs in `DiracToArcID()` and `arcToDiracID()`: from `warn` to `verbose` or `debug`.
What do you think?

sfayer:
I agree that doing the conversion multiple times is a bit excessive... Perhaps we could split killJob into an internal _killJob function that uses the ArcIDs, call that from getJobStatus and in the public killJob function just do the DiracToArcID and then call _killJob. That way we'd only do the conversion once where needed and avoid spurious logging, what do you think?

aldbr:
Alright, yes I agree.
Do you want to take care of this?

sfayer:
Sure, done.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -393,38 +429,11 @@ def killJob(self, jobIDList):
 
         :param list: list of DIRAC Job IDs
```
aldbr:
```suggestion
        :param list jobIDList: list of DIRAC Job IDs
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -321,6 +321,42 @@ def _uploadJobDependencies(self, arcJobID, executableFile, inputs, executables):
             self.log.verbose("Input correctly uploaded", fileToSubmit)
         return S_OK()
 
+    def _killJob(self, jobIDList):
+        """Kill the specified jobs
+
+        :param list: list of ARC Job IDs
```
aldbr:
```suggestion
        :param list arcJobList: list of ARC Job IDs
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -321,6 +321,42 @@ def _uploadJobDependencies(self, arcJobID, executableFile, inputs, executables):
             self.log.verbose("Input correctly uploaded", fileToSubmit)
         return S_OK()
 
+    def _killJob(self, jobIDList):
```
aldbr: Just to be in line with `renewJobs(self, arcJobList)`:
```suggestion
    def _killJob(self, arcJobList):
```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -321,6 +321,42 @@ def _uploadJobDependencies(self, arcJobID, executableFile, inputs, executables):
...
+
+        # Get a proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("Failed to set up proxy", result["Message"])
+            return result
+        self.session.cert = Locations.getProxyLocation()
+
+        # List of jobs in json format for the REST query
+        jobsJson = {"job": [{"id": job} for job in jobIDList]}
```
aldbr: Oops, it should be changed here too. This should fix the problems with the CI: 
```suggestion
        jobsJson = {"job": [{"id": job} for job in arcJobList]}
```

aldbr (approved):

fstagni (approved):

PR #6662, opened 2023-01-10
[v8r0] Fixes from the LHCbDIRAC hackaton

chaen:
After the refactoring of gfal2, uploading a file from `./myFile.txt` was broken

BEGINRELEASENOTES

*Resources
FIX: fix upload from relative path

ENDRELEASENOTES


src/DIRAC/Resources/Storage/StorageElement.py
```diff
@@ -1439,6 +1440,7 @@ def addAccountingOperation(self, urlDict, startDate, elapsedTime, storageParamet
                             "Protocol": accountingDict["Protocol"],
                             "Error": str(errorMsg),
                             "Component": "StorageElement",
+                            "Hostname": socket.getfqdn(),
```
chrisburr: We should probably stick to using `Network.getFQDN()`

fstagni (approved):

PR #6713, opened 2023-01-24
[7.3] Matcher fix tag is empty string (backport)

andresailer:
Backport from #6701 
BEGINRELEASENOTES

*WMS
FIX: Matcher handles cases of empty Tag in resource description

ENDRELEASENOTES


chrisburr:
The `sweep:ignore` label takes priority so you don't need to worry about `alsoTargeting:xxx` (it's a limitation of the action we use to add labels but it's not worth writing our own).

chrisburr (approved):

PR #6728, opened 2023-01-26
[v7r3] Add a note about XRSL times in ARC(6) vs AREX

sfayer:
As discussed in the BiLD meeting.

BEGINRELEASENOTES
*Resources
FIX: Add a note about XRSL times in ARC(6) vs AREX
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/ARCComputingElement.py
+++ b/src/DIRAC/Resources/Computing/ARCComputingElement.py
@@ -32,10 +32,16 @@ XRSLExtraString:
 
      (key = "value")
 
+   Please note that for ARC & ARC6, any times (such as wall or CPU time) in the XRSL should be specified in minutes.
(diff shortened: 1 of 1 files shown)


src/DIRAC/Resources/Computing/ARCComputingElement.py
```diff
@@ -32,11 +32,16 @@
 
      (key = "value")
 
+   Please note that for ARC & ARC6, any times (such as wall or CPU time) in the XRSL should be specified in minutes.
+   For AREX, these times should be given in seconds.
```
aldbr: You might want to add a link pointing to the official documentation even if it is not crystal clear:


```suggestion
   For AREX, these times should be given in seconds (see https://www.nordugrid.org/arc/arc6/users/xrsl.html?#cputime).

```

fstagni (approved):

PR #6738, opened 2023-01-30
[v7r3] Ensure all entrypoints are importable

chrisburr:
Adds a test which we've found was useful in LHCb to make sure that all of the scripts are importable.

### DIFF
--- a/src/DIRAC/Core/Utilities/test/Test_entrypoints.py
+++ b/src/DIRAC/Core/Utilities/test/Test_entrypoints.py
@@ -1,0 +1,16 @@
+import importlib_metadata as metadata
+import pytest
+import six
+
+
+@pytest.mark.skipif(six.PY2, reason="Only makes sense for Python 3 installs")
+def test_entrypoints():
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/Utilities/test/Test_entrypoints.py
```diff
@@ -0,0 +1,13 @@
+import importlib_metadata as metadata
+
+
+def test_entrypoints():
+    """Make sure all console_scripts defined by LHCbDIRAC are importable."""
+    errors = []
+    for ep in metadata.entry_points(group="console_scripts"):
+        if ep.module.startswith("LHCbDIRAC"):
```
aldbr: I guess it should be `DIRAC` instead of `LHCbDIRAC` here, shouldn't it?
```suggestion
        if ep.module.startswith("DIRAC"):
```

chrisburr: 🤦

fstagni (approved):

PR #6745, opened 2023-01-31
[v7r3] Test for StorageElementHandler / FileHelper 

andresailer:
Draft until fully implemented. Also concerns #6724 

BEGINRELEASENOTES

*Tests
NEW: Integration test for StorageElementHandler and FileHelper

ENDRELEASENOTES

### DIFF
--- a/tests/Integration/DataManagementSystem/Test_StorageElement.py
+++ b/tests/Integration/DataManagementSystem/Test_StorageElement.py
@@ -1,0 +1,110 @@
+"""This module contains tests for the (DIP) StorageElement
+
+The tests upload gets a directory
+
+Requirements:
+
+* Running StorageElement instance, containing an LFN folder '/Jenkins', writable by the jenkins_user proxy
+* A jenkins_user proxy
+
+"""
+
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+import os
+import sys
+import tempfile
+import time
+
+import pytest
+
+from pprint import pprint, pformat
+
+from DIRAC.Core.Base.Script import parseCommandLine
+
+parseCommandLine()
+
+from DIRAC.Core.Security.ProxyInfo import getProxyInfo  # noqa: module-import-not-at-top-of-file
+from DIRAC.ConfigurationSystem.Client.Helpers.Registry import getVOForGroup  # noqa: module-import-not-at-top-of-file
+
+
+@pytest.fixture(scope="module")
+def se():
+    from DIRAC.Resources.Storage.StorageElement import StorageElementItem as StorageElement
+
+    return StorageElement("SE-1")
+
+
+try:
+    res = getProxyInfo()
+    if not res["OK"]:
+        raise Exception(res["Message"])
+
+    proxyInfo = res["Value"]
+    username = proxyInfo["username"]
+    vo = getVOForGroup(proxyInfo["group"])
+
+    if not vo:
+        raise ValueError("Proxy has no VO")
+
+    LFN_PATH = "/%s/test/unit-test/StorageElement/" % vo
+
+except Exception as e:  # pylint: disable=broad-except
+    print(repr(e))
+    sys.exit(2)
+
+
+@pytest.fixture
+def tempFile():
+    """Create unique temprary file, will be cleaned up at the end."""
+    with tempfile.NamedTemporaryFile("w") as theTemp:
+        theTemp.write(str(time.time()))
+        theTemp.flush()
+        yield theTemp.name
+
+
+def assertResult(res, lfn, sub="Successful"):
+    """Check if lfn is in Successful or Failed, given by sub."""
+    assert res["OK"], res.get("Message", "All OK")
+    assert sub in res["Value"]
+    assert res["Value"][sub], "Did not find %r in result" % sub
+    assert lfn in res["Value"][sub]
+    assert "Files" in res["Value"][sub][lfn]
+    assert "Size" in res["Value"][sub][lfn]
+
+
+def test_getDirectory(se, tempFile, tmp_path):
+    print("\n\n#########################################################################\n\n\t\t\tSE.GetDirectory\n")
+    targetDir = os.path.join(LFN_PATH, "myGetDir")
+    targetLFN = os.path.join(targetDir, os.path.basename(tempFile))
+    putFileRes = se.putFile({targetLFN: tempFile})
+    pprint(putFileRes)
+    localDir = os.path.join(tmp_path, "myGetDir")
+    print("is not a dir?", localDir, os.path.isdir(localDir))
+    assert not os.path.isdir(localDir)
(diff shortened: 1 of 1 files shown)


andresailer:
Ready for review.

tests/Integration/DataManagementSystem/Test_StorageElement.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@pytest.fixture
+def tempFile():
+    """Create unique temprary file, will be cleaned up at the end."""
+    theTemp = tempfile.NamedTemporaryFile("w")
+    theTemp.write(str(time.time()))
+    theTemp.seek(0)  # write the characters to file
+    yield theTemp.name
+    theTemp.close()
```
chrisburr: Is there a reason for not using the context manager and flushing instead of `seek(0)`?
```suggestion
    with tempfile.NamedTemporaryFile("w") as theTemp:
        theTemp.write(str(time.time()))
        theTemp.flush()
        yield theTemp.name
```

tests/Integration/DataManagementSystem/Test_StorageElement.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@pytest.fixture
+def localDir():
```
chrisburr: I think this can be replaced with pytest's [`tmp_path`](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path) fixture

tests/Integration/DataManagementSystem/Test_StorageElement.py
andresailer: I just copied this from Test_DataManagementClient (I think).

tests/Integration/DataManagementSystem/Test_StorageElement.py
andresailer: Nice!

tests/Integration/DataManagementSystem/Test_StorageElement.py
andresailer: Note: [flush()](https://docs.python.org/2/library/stdtypes.html#file.flush) does not necessarily write the file’s data to disk. Use [flush()](https://docs.python.org/2/library/stdtypes.html#file.flush) followed by [os.fsync()](https://docs.python.org/2/library/os.html#os.fsync) to ensure this behavior.

chrisburr: We don't care about `fsync`-ing the file to disk. We just need to make sure the write buffer is flushed to the filesystem.

andresailer: I sure don't appreciate the difference there :).

andresailer: Seems to work with just the `flush`.

andresailer:
All comments implemented, thanks @chrisburr!

tests/Integration/DataManagementSystem/Test_StorageElement.py
chrisburr: > I sure don't appreciate the difference there :).

By default Python uses buffered IO which means when you write a small amount of data to the file it doesn't get written:

```python
In [1]: import os
   ...: fh = open("/tmp/example.txt", "wt")
   ...: fh.write("hello")
   ...: !echo File contains: $(cat /tmp/example.txt)
File contains:
```

Calling `fsync()` tells the kernel to make sure the data has been written to the physical media, however nothing has been written yet so nothing actually happens:

```python
In [2]: os.fsync(fh.fileno())
   ...: !echo File contains: $(cat /tmp/example.txt)
File contains:
```

Calling `flush()` has the effect of manually clearing the file's buffer at writing it so it shows up when you read the file:

```python
In [3]: fh.flush()
   ...: !echo File contains: $(cat /tmp/example.txt)
File contains: hello
```

However at this point the file might only be written to the kernel's filesystem cache in memory. This means that a sudden power loss or kernel panic might result in the data being irrecoverably lost. To avoid this you can manually call [`fsync()`](https://man7.org/linux/man-pages/man2/fsync.2.html) which tells the kernel to make sure the data is cleared from the filesystem cache and is actually written to disk. After `fsync()` returns the data should be "safe":

```python
In [4]: os.fsync(fh.fileno())
   ...: !echo File contains: $(cat /tmp/example.txt)
File contains: hello
```

This tends to be used by databases to ensure data is really written and you never end up rolling back to a previous point in time as a result of a crash.

For completeness, you can use unbuffered io (aka Raw IO) which means `flush()` no longer does anything (though `fsync()` is still needed to ensure data is safe).

```python
In [5]: import os
   ...: fh = open("/tmp/example.txt", "wb", buffering=0)
   ...: fh.write(b"hello")
   ...: !echo File contains: $(cat /tmp/example.txt)
File contains: hello
```

(There is a ton of nuance and non-conformity, but I think this is roughly how it's _supposed_ to work.)

andresailer: Thanks for the explanation!

chrisburr (dismissed):

tests/Integration/DataManagementSystem/Test_StorageElement.py
```diff
@@ -0,0 +1,126 @@
...
+
+import pytest
+
+from pprint import pprint, pformat
+
+from DIRAC.Core.Base.Script import parseCommandLine
+
+parseCommandLine()
+
+from DIRAC.Core.Security.ProxyInfo import getProxyInfo  # noqa: E402
```
chaen: Could you please use the error names instead of the error code (`E4021)

tests/Integration/DataManagementSystem/Test_StorageElement.py
```diff
@@ -0,0 +1,126 @@
...
+    assert se.listDirectory(targetDir)
+
+
+def test_getOccupancy(se):
+    print("\n\n#########################################################################\n\n\t\t\tSE.PutDirectory\n")
+    res = se.getOccupancy()
+    assert res["OK"], pformat(res)
+
+
+def test_seDump(se):
```
chaen: there's already an integration test for this method, just in case...
https://github.com/DIRACGrid/DIRAC/blob/ae63eee7c66f255d50e4d8856f6e5c5ecb03ba28/tests/Integration/DataManagementSystem/Test_Client_DFC.py#L345

tests/Integration/DataManagementSystem/Test_StorageElement.py
andresailer: Done

tests/Integration/DataManagementSystem/Test_StorageElement.py
andresailer: Removed the test for getSEDump here

andresailer:
All comments implemented, thanks @chaen

fstagni (approved):

PR #6750, opened 2023-02-01

atsareg:
This PR adds possibility to submit pilots with tokens. It replaces #6580 PR which was corrupted and restarted anew here.
See review comments and answers there.

The tokens are obtained by SiteDirectors via a standard mechanism from the TokenManager. Using OAuth2 IdProviders with pilot tokens generated from the long tokens in the TokenDB are supported.

 In order to use the oidc-agent instead of the OAuth2 flow, a new OidcAgentIdProvider is added. This can be useful in some environments, also for debugging.  The oidc-agent is supposed to have a client created with the name of the user submitting pilots and having necessary compute scopes.

The following ComputingElements should be enabled with the tokens:

- [x] HTCondorCE
- [x] ARC/ARC6
- [x] AREX

    The case of CloudComputingElement is not addressed specifically as it can be solved by using Application Credentials authentication mechanism where Application Credentials can be created after authentication with OAuth2 tokens.

BEGINRELEASENOTES

*Framework
FIX: TokenManager - use refreshToken flow to generate access tokens from the stored refresh tokens

*WorkloadManagement

*Resources
NEW: HTCondorComputingElement, ARC(6)ComputingElement and AREXComputingElement enabled to for job operations with tokens

ENDRELEASENOTES


src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -93,6 +95,18 @@ def _reset(self):
 
     #############################################################################
 
+    def setToken(self, token, valid=0):
```
aldbr: `setToken()` and `self.valid` does not seem to be used, is this expected?

Also, the content of `AREX.setToken()` is similar to the content of `ComputingElement.setToken()`, here is a suggestion:

```python
def setToken(self, token, valid=0):
    """..."""
    super(AREXComputingElement, self).setToken()
    self.headers["Authorization"] = f"Bearer {self.token['access_token']}"
```

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -237,6 +251,40 @@ def _reset(self):
...
+            htcEnv = {"_CONDOR_SEC_CLIENT_AUTHENTICATION_METHODS": "GSI"}
+
+        result = executeGridCommand(
+            self.proxy,
+            cmd,
+            gridEnvScript=self.gridEnv,
+            gridEnvDict=htcEnv,
+        )
+        # Remove token file if we do not want to keep it
+        if tFile:
```
aldbr, line 279: If `self.token` is not defined here `tFile` is not initialized and does not exist.
`tFile=None` should be moved before `self.token`.

src/DIRAC/FrameworkSystem/private/authorization/AuthServer.py
```diff
@@ -441,12 +441,12 @@ def validateIdentityProvider(self, request, provider):
...
         groupProvider = getIdPForGroup(request.group) if request.groups else None
 
-        # If requested access token for group that is not registred in any identity provider
+        # If requested access token for group that is not registered in any identity provider
         # or the requested provider does not match the group return error
         if request.group and not groupProvider and "proxy" not in request.scope:
             raise Exception("The %s group belongs to the VO that is not tied to any Identity Provider." % request.group)
 
-        sLog.debug("Check if %s identity provider registred in DIRAC.." % request.provider)
+        sLog.debug("Check if %s identity provider registered in DIRAC.." % request.provider)
```
aldbr:
```suggestion
        sLog.debug(f"Check if {request.provider} identity provider registered in DIRAC...")
```

src/DIRAC/FrameworkSystem/private/authorization/AuthServer.py
```diff
@@ -441,12 +441,12 @@ def validateIdentityProvider(self, request, provider):
         # Find identity provider for group
         groupProvider = getIdPForGroup(request.group) if request.groups else None
 
-        # If requested access token for group that is not registred in any identity provider
+        # If requested access token for group that is not registered in any identity provider
         # or the requested provider does not match the group return error
         if request.group and not groupProvider and "proxy" not in request.scope:
             raise Exception("The %s group belongs to the VO that is not tied to any Identity Provider." % request.group)
```
aldbr, line 447:
```suggestion
            raise Exception(f"The {request.group} group belongs to the VO that is not tied to any Identity Provider")
```

src/DIRAC/Resources/IdProvider/IdProvider.py
```diff
@@ -26,3 +33,27 @@ def setParameters(self, parameters):
...
+        """Get DIRAC groups related to scope"""
+        groups = []
+        for group in getAllGroups():
+            if (g_scope := self.getGroupScopes(group)) and set(g_scope).issubset(scope_to_list(scope)):
+                groups.append(group)
+        return groups
+
+    def refreshToken(self, refreshToken=None, **kwargs):
+
+        return S_ERROR("Not implemented")
```
chrisburr: How about making this an abstract base class so it can only be used if you have an actual implementation?

```
In [1]: from abc import ABCMeta, abstractmethod
   ...:
   ...: class AbstractIdProvider(metaclass=ABCMeta):
   ...:     @abstractmethod
   ...:     def refreshToken(self, refreshToken=None, **kwargs):
   ...:         ...
   ...:
   ...: class CheckInIdProvider(AbstractIdProvider):
   ...:     def refreshToken(self, refreshToken=None, **kwargs):
   ...:         print("Hello world")
   ...:

In [2]: AbstractIdProvider()
Cell In[2], line 1
----> 1 AbstractIdProvider()

TypeError: Can't instantiate abstract class AbstractIdProvider with abstract method refreshToken

In [3]: CheckInIdProvider()
Out[3]: <__main__.CheckInIdProvider at 0x7f1293768a90>
```

src/DIRAC/FrameworkSystem/private/authorization/AuthServer.py
```diff
@@ -282,7 +282,7 @@ def parseIdPAuthorizationResponse(self, response, session):
         :return: S_OK(dict)/S_ERROR()
         """
         providerName = session.pop("Provider")
-        sLog.debug("Try to parse authentification response from %s:\n" % providerName, pprint.pformat(response))
+        sLog.debug("Try to parse authentication response from %s:\n" % providerName, pprint.pformat(response))
```
aldbr:
```suggestion
        sLog.debug(f"Try to parse authentication response from {providerName}:\n{pprint.pformat(response)}")
```

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: What is a real advantage of creating an abstract class ?

chrisburr: * When we configure it incorrectly we don't get `{"OK": False, "Message": "Not implemented"}` in the logs with no idea where it came from
* It becomes impossible to instantiate the base class so you get an error as early as possible
* Static analysers are more likely to be able to understand what you're doing
* The abstract methods don't need to return failures so you can still use `super()` to call them and get correct MRO (though I hope we don't have any class hierarchies that are complex enough for that to matter)

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -29,12 +29,14 @@
...
 import json
 import requests
+import datetime
 
 from DIRAC import S_OK, S_ERROR
 from DIRAC.Core.Security import Locations
 from DIRAC.Core.Security.ProxyInfo import getVOfromProxyGroup
 from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
 from DIRAC.Resources.Computing.ARCComputingElement import ARCComputingElement
+from DIRAC.Core.Utilities.TimeUtilities import second
```
aldbr: Not needed anymore
```suggestion

```

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -29,12 +29,14 @@
 import os
 import json
 import requests
+import datetime
```
aldbr: Not needed anymore
```suggestion

```

src/DIRAC/Resources/IdProvider/OidcAgentIdProvider.py
```diff
@@ -0,0 +1,36 @@
...
+
+import subprocess
+
+from DIRAC import S_OK, S_ERROR
+from DIRAC.Core.Utilities.Subprocess import shellCall
+from DIRAC.Resources.IdProvider.IdProvider import IdProvider
+from DIRAC.FrameworkSystem.private.authorization.utils.Tokens import OAuth2Token
+
+
+class OidcAgentIdProvider(IdProvider):
```
aldbr: I am probably missing some details here, but it seems like there is an "architectural issue": In the `IdProvider` directory, I see a mix between identity providers (`IAM`, `CheckIn`), protocols (`OAuth2`) and methods to get tokens (`OIDCAgent`).

In the current model, you cannot get IAM/CheckIn scopes from a group if you use `OIDCAgentIdProvider`, can you?
Also, there is no `getToken()` in `OAuth2IdProvider`, is it expected? (`OAuth2IdProvider.fetchToken()` looks very similar though).

Here is a suggestion:


IdProvidersUML

src/DIRAC/Resources/Computing/ARCComputingElement.py
```diff
@@ -390,6 +394,10 @@ def killJob(self, jobIDList):
         for jobID in jobList:
             jobs.append(self._getARCJob(jobID))
 
+        # Set the token in the environment if needed
+        if self.token:
+            os.environ["BEARER_TOKEN"] = self.token["access_token"]
```
fstagni: I see a repeated pattern in most of these methods (common to at least ARC, ARC6, AREX), that is: 
1. call `_prepareProxy`, set it in `os.environ["X509_USER_PROXY"]`
2. add token to `os.environ["BEARER_TOKEN"]` if it exists

I would suggest to code a simple decorator that would do the above operations. It would reduce code repetition and have a clear entry pointy for future modifications.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -367,42 +430,82 @@ def getJobStatus(self, jobIDList):
...
             resultDict[job] = pilotStatus
 
+        if self.tokenFile:
+            os.remove(self.tokenFile)
+            self.tokenFile = None
+
         self.log.verbose("Pilot Statuses: %s " % resultDict)
         return S_OK(resultDict)
 
+    def _treatCondorHistory(self, condorHistCall, qList):
```
fstagni, line 471: Did you move this from `DIRAC.Resources.Computing.BatchSystems.Condor.treatCondorHistory`? Why?
Anyway, you can remove the import now but mostly remove the other code to avoid duplications?

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -367,42 +430,82 @@ def getJobStatus(self, jobIDList):
...
+        """concatenate clusterID and processID to get the same output as condor_q
+        until we can expect condor version 8.5.3 everywhere
+
+        :param str condorHistCall: condor_history command to run
+        :param qList: list of jobID and status from condor_q output, will be modified in this function
+        :type qList: python:list
```
fstagni:
```suggestion
        :param list qList: list of jobID and status from condor_q output, will be modified in this function
```

src/DIRAC/Resources/IdProvider/OidcAgentIdProvider.py
fstagni: I second @aldbr here. I would just integrate also the suggestion of @chrisburr above to make `IdProvider`, and also this new (?) `AccessStrategy` modules Abstract Classes (there would no be more "extensions" rather than "implementations" then...)

Q to @aldbr : in the drawing above are "OAuth2" and "@OAuth2lib" the "same" (apart from they inherit from)?

And in general, I think we agreed that we might not really want to support oidc-agent.

aldbr: By the way, I wonder whether the term of `IdProvider` is adapted in this context.
In the papers I have read so far, `IAM` and `CheckIn` are defined as `Token Issuer` or `OpenID Providers`, while `IdProvider` refers to services that actually store information about the users such as `Google`, `Facebook`, `CERN?`.

What do you think?

aldbr: > Q to @aldbr : in the drawing above are "OAuth2" and "@OAuth2lib" the "same" (apart from they inherit from)?

Yes I didn't specify it sorry, but `OAuth2` in the current model = `OAuth2lib` in the suggested model.

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: MRO ?

src/DIRAC/Resources/IdProvider/OidcAgentIdProvider.py
atsareg: The OidcAgentIdProvider removed in favor of a method in the OAuth2IdProvider class

src/DIRAC/Resources/Computing/ARCComputingElement.py
atsareg: Agreed

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: Updated

src/DIRAC/FrameworkSystem/private/authorization/AuthServer.py
atsareg: Updated

atsareg:
I hope to have addressed all the comments. Please, rereview

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
```diff
@@ -177,7 +177,7 @@ def __checkProperties(self, requestedUserDN: str, requestedUserGroup: str):
     @gTokensSync
     def export_getToken(
         self,
-        username: str,
+        userName: str,
```
chrisburr, line 179: Does this need to be changed? It becomes inconsistent with other usages like `getDNForUsername`.

src/DIRAC/Resources/Computing/ARCComputingElement.py
```diff
@@ -88,6 +87,29 @@
...
+        # Prepare first the proxy
+        result = self._prepareProxy()
+        if not result["OK"]:
+            self.log.error("ARCComputingElement: failed to set up proxy", result["Message"])
+            return result
+        self.usercfg.ProxyPath(os.environ["X509_USER_PROXY"])
+
+        # Set the token in the environment if needed
+        if self.token:
+            os.environ["BEARER_TOKEN"] = self.token["access_token"]
```
chrisburr, line 106: Should this be cleaned from the environment after the function returns?

src/DIRAC/Resources/IdProvider/IdProvider.py
```diff
@@ -1,9 +1,12 @@
...
 """
-from DIRAC import gLogger
 
+from abc import ABCMeta, abstractmethod
 
-class IdProvider:
+from DIRAC import gLogger, S_ERROR
+
+
+class IdProvider(metaclass=ABCMeta):
```
aldbr: Do we keep `IdProvider`?
I thought we would just have `OAuth2IDProvider` as the base class, with `IAM` and `CheckIn` inheriting from it.

src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -93,6 +93,17 @@ def _reset(self):
 
     #############################################################################
 
+    def setToken(self, token, valid=0):
```
chrisburr: Does it make sense to have `valid=0` as an optional argument? If the default is 0 I guess it always needs to be set:
```suggestion
    def setToken(self, token, valid):
```

src/DIRAC/Resources/IdProvider/IdProvider.py
```diff
@@ -13,16 +16,20 @@ def __init__(self, **kwargs):
...
+        self._initialize()
 
-    def _initialization(self, **kwargs):
+    def _initialize(self):
         """Initialization"""
-        pass
+        self.name = self.parameters.get("ProviderName")
 
-    def setParameters(self, parameters):
+    def setParameters(self, parameters: dict):
```
chrisburr: `dict` isn't a valid type hint. The types of the keys should be included in square brackets.

src/DIRAC/Resources/IdProvider/OAuth2IdProvider.py
```diff
@@ -560,3 +560,27 @@ def getUserProfile(self):
         except Exception as e:
             self.log.exception(e)
             return S_ERROR("Cannot get user profile: %s" % repr(e))
+
+    def getTokenFromAgent(self, userName: str, requiredScope=None, requiredTimeLeft=3500):
```
chrisburr: I thought we decided not to have any code for `oidc-agent`?

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -237,6 +251,40 @@ def _reset(self):
...
 
+    def _executeCondorCommand(self, cmd, keepTokenFile=False):
+
+        tFile = None
+        if self.token:
+            # Create a new token file if we do not keep it across several calls
+            if keepTokenFile:
+                tFile = self.tokenFile
+            if not tFile:
+                fd, tFile = tempfile.mkstemp(suffix=".token", prefix="HTCondorCE_", dir=self.workingDirectory)
```
chrisburr: Do these files ever get cleaned up?

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
```diff
@@ -211,39 +211,40 @@ def export_getToken(
...
-                    return result
-                self.log.verbose(f"Failed to get token with cached tokens: {result['Message']}")
-                # Let's try to revoke broken token
-                idpObj.revokeToken(token["refresh_token"])
 
         err = []
-        # The cache did not help, so let's make an exchange token
-        result = Registry.getDNForUsername(username)
+        # Do we have an oidc-agent configured ? If yes, try to use it
+        if userName in idpObj.parameters.get("OIDCAgentUsers", []):
```
aldbr: The `OIDC-agent` will not be supported within DIRAC, therefore I would not include it as an option of the `TokenManagerHandler` (IMHO, it adds more complexity than before where the implementation was encapsulated in a separate class).

Last time you said that it was a recommended solution, where did you hear/see this?
I still don't get why we need it somewhere in the code.

src/DIRAC/FrameworkSystem/API/AuthHandler.py
```diff
@@ -435,9 +435,9 @@ def __researchDIRACGroup(self, extSession, chooseScope, state):
         username = extSession["authed"]["username"]
         # Requested arguments in first request
         provider = firstRequest.provider
-        self.log.debug("Next groups has been found for %s:" % username, ", ".join(firstRequest.groups))
+        self.log.debug("The following groups found for %s:" % username, ", ".join(firstRequest.groups))
```
chrisburr:
```suggestion
        self.log.debug("The following groups found", f"for {username}: {', '.join(firstRequest.groups)}")
```

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
```diff
@@ -211,39 +211,40 @@ def export_getToken(
...
+                self.__tokensCache.add(
+                    cacheKey,
+                    result["Value"].get_claim("exp", "refresh_token") or self.DEFAULT_RT_EXPIRATION_TIME,
+                    result["Value"],
+                )
+                return result
+            else:
+                err.append(result["Message"])
+
+        # No luck so far, let's refresh the token stored in the database
```
chrisburr, line 230: I'm struggling to understand where this is using the REST interface of IAM/CheckIn to get the tokens. Am I missing somthing?

src/DIRAC/WorkloadManagementSystem/ConfigTemplate.cfg
```diff
@@ -279,6 +279,8 @@ Agents
     CETypes = any
     # List of CEs that will be treated by this SiteDirector ("any" can refer to any type of CE defined in the CS)
     CEs = any
+    # List of Tags that are required to be present in the CE/Queue definition
+    Tags =
```
fstagni, line 283: Should it be more like "RequiredTags"?

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
aldbr: IIUC, here: https://github.com/DIRACGrid/DIRAC/pull/6750/files/d94157958af74dffd0dca4abea0ef82cf6106342#diff-d582aa662e6992b4d896c3050cfb033627550a2c5ea1f0b67f6f54298c0647beR262

If `idpObj` is an instance of `OAuth2IdProvider`, then it calls `authlib.OAuthClient.refreshToken()` which makes the REST request.
(`OAuth2IdProvider` inherits from `OAuth2Session`, which inherits from `authlib.OAuth2Session`, which inherits from `authlib.OAuth2Client`)

src/DIRAC/WorkloadManagementSystem/ConfigTemplate.cfg
atsareg: No, these are not the Tags required by the site to be present in the Jobs. This is the semantics of the RequiredTags option. This is to say that the SiteDirector only serves sites with certain properties.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
atsareg: Yes. Either in this same function if keeping tokens across several calls is not requested. Or in the calling function, e.g. killJob,  where the same token file is used for several calls.

src/DIRAC/Resources/IdProvider/OAuth2IdProvider.py
atsareg: I think this does not harm and can be useful without any extra charge

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: Never used it before. Should we probably discuss how to use type hints in the DIRAC code ? Use it systematically, or just when updating codes, or... For example, the syntax without square brackets is apparently also valid as the tests pass.

src/DIRAC/Resources/Computing/ARCComputingElement.py
atsareg: I do not think it is necessary. This is only the environment of a single SiteDirector for which there is a single possible value.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
chrisburr: I'd suggest using `NamedTemporaryFile` rather than `mkstemp` in which case the file will be cleaned automatically when the object is garbage collected.

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
atsareg: We do have conventions on variables naming that we should follow. I will correct for that also the Registry helper

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: Well, if there is such a strong consensus against the oidc-agent, I will drop it. Farewell, you were very useful for the tests ...

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -168,6 +170,17 @@ def __writeSub(self, executable, nJobs, location, processors):
 
         executable = os.path.join(self.workingDirectory, executable)
 
+        useCredentials = "use_x509userproxy = true"
+        if tokenFile:
+            useCredentials = (
```
andresailer: please use `textwrap.dedent` https://docs.python.org/3/library/textwrap.html#textwrap.dedent

src/DIRAC/Resources/IdProvider/IdProvider.py
fstagni: There is an issue for this: https://github.com/DIRACGrid/DIRAC/issues/6315

I think that as for other cases of syntax updates we should rely on existing tools rather than try to do it ourselves.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -168,6 +170,17 @@ def __writeSub(self, executable, nJobs, location, processors):
...
+                """
+use_x509userproxy = true
```
andresailer: This is now always present, is that intentional? if yes, why not leave it in the default?

src/DIRAC/Resources/IdProvider/IdProvider.py
atsareg: Well, I tried the abstract class to see it works. It works. So, I can drop it now :)

src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
atsareg: OK, I will drop it. After the initial testing, this is not actually important.

src/DIRAC/Resources/IdProvider/IdProvider.py
chrisburr: > For example, the syntax without square brackets is apparently also valid as the tests pass.

cpython doesn't use the type hints at runtime and we don't have any CI jobs to check them so you can put almost anything without breaking the tests.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
atsareg: Yes, we always ask to pass X509 proxy to the job environment, even if we submit with tokens.

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -188,7 +201,7 @@ def __writeSub(self, executable, nJobs, location, processors):
         sub = """
 executable = %(executable)s
 universe = %(targetUniverse)s
-use_x509userproxy = true
+%(useCredentials)s
```
andresailer, line 205:
```suggestion
use_x509userproxy = true
%(useCredentials)s
```
And remove from useCredentials

atsareg: Agreed

src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
atsareg: OK, done

atsareg:
I do not see the problematic commit message in the interactive rebase, so I can not fix it. Any idea how this can be overcome ?

chaen:
Squash all your commits in one or two (it's much nicer in general)

atsareg:
Something went wrong with squashing commits. Redoing the PR

PR #6764, opened 2023-02-06
[8.0] dirac-configure warning about existing file

chaen:
BEGINRELEASENOTES

*Core
New: dirac-configure warns you about existing config file

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/scripts/dirac_configure.py
+++ b/src/DIRAC/Core/scripts/dirac_configure.py
@@ -593,6 +593,8 @@ def runDiracConfigure(params):
         mkDir(configDir)
         params.update = True
         DIRAC.gConfig.dumpLocalCFGToFile(params.outputFile)
+    elif not params.forceUpdate:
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/scripts/dirac_configure.py
```diff
@@ -593,6 +593,8 @@ def runDiracConfigure(params):
         mkDir(configDir)
         params.update = True
         DIRAC.gConfig.dumpLocalCFGToFile(params.outputFile)
+    else:
+        DIRAC.gLogger.info(f"{params.outputFile} exists, not overwriting it.")
```
andresailer:
```suggestion
        DIRAC.gLogger.notice(f"{params.outputFile} exists, not overwriting it. Or use the '--ForceUpdate' Flag")
```
* Very good addition!
* Info still doesn't print in normal script use, does it? (neither does warn??) So have to use `notice`...
* ForceUpdate still works, right?

chaen: indeed ! Thanks

fstagni (approved):

PR #6773, opened 2023-02-08
[8.0] Intermediate replication uses all possible SE

chaen:
When replicating a file interactively with the `DataManager`, we first try TPC. If that does not work, we replicate locally before uploading. The sources used for that intermediate replication were only the SE for which no potential TPC protocol were found. Now we also take as sources the SE for which there were possible protocols, but the replication still failed.

BEGINRELEASENOTES

*DMS
CHANGE: take into account all possible sources for intermediate copy when replicating files

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Client/DataManager.py
+++ b/src/DIRAC/DataManagementSystem/Client/DataManager.py
@@ -853,6 +853,10 @@ class DataManager:
             else:
                 log.debug("Catalog size and physical size match")
 
+            # The file at the SE seems healthy, so we could potentially use this SE
+            # for non TPC transfer in case everything else fails.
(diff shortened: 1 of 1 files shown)


src/DIRAC/DataManagementSystem/Client/DataManager.py
```diff
@@ -853,6 +853,10 @@ def __replicate(self, lfn, destSEName, sourceSEName="", destPath="", localCache=
             else:
                 log.debug("Catalog size and physical size match")
 
+            # The file at the SE seems sain, so we could potentially use this SE
```
andresailer:
```suggestion
            # The file at the SE seems healthy, so we could potentially use this SE
```

fstagni (approved):

PR #6837, opened 2023-02-21
[v7r3] Complete AREX renewal

sfayer:
Hi,

After some more AREX testing, I noticed that it doesn't complete proxy renewals: It gets the new CSR but never actually re-does the delegation.

Regards,
Simon

BEGINRELEASENOTES
*Resources
FIX: Complete AREX renewal
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -553,6 +553,7 @@ class AREXComputingElement(ARCComputingElement):
                     "%s whose proxy expires at %s" % (arcJob, timeLeft),
                 )
                 # Proxy needs to be renewed - try to renew it
+                # First, get a new CSR from the delegation
                 params = {"action": "renew"}
                 query = self._urlJoin(os.path.join("delegations", delegationID))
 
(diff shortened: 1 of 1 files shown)


src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -562,11 +562,19 @@ def _renewJobs(self, arcJobList):
                     params=params,
                     timeout=self.arcRESTTimeout,
                 )
+
                 if response.ok:
-                    self.log.debug("Proxy successfully renewed", "for job %s" % arcJob)
+                    result = self.__uploadCertificate(delegationID, response.text)
```
aldbr, line 569: Indeed, you are right, it is actually explained in the [ARC REST documentation](http://www.nordugrid.org/arc/arc6/tech/rest/rest.html#delegations-management): 

```
Renew delegation
    POST <base URL>/delegations/<delegation id>?action=renew initiates renewal of delegation. Response is 200 with certificate request of application/x-pem-file type
```

You could probably just add small comments above to briefly explain that:

Line 556:
```python
# Proxy needs to be renewed - try to renew it
# First, get a new CSR from the delegation
params = {"action": "renew"}
...
```

Line 567:
```python
# Then, sign and upload the certificate
result = self.__uploadCertificate(delegationID, response.text)
...
```



Thanks a lot for fixing the proxy renewal process, we have not started to use it yet and, therefore, have not focused on it.

sfayer: Sure, I've added the comments you've suggested.

aldbr (approved):

PR #6850, opened 2023-02-23
[v7r3] Add setPriority to Job API

sfayer:
As discussed in #6841 .

Regards
Simon

BEGINRELEASENOTES
*Interfaces
NEW: Add setPriority to Job API
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Interfaces/API/Job.py
+++ b/src/DIRAC/Interfaces/API/Job.py
@@ -805,6 +805,30 @@ class Job(API):
         return S_OK()
 
     #############################################################################
+    def setPriority(self, priority):
+        """Helper function.
+
+        Sets the job priority in the range 1-10.
+        Priorities 1-9 are matched probabilistically with higher numbers being more likely.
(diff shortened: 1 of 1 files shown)


src/DIRAC/Interfaces/API/Job.py
```diff
@@ -804,6 +804,28 @@ def setJobGroup(self, jobGroup):
         self._addParameter(self.workflow, "JobGroup", "JDL", jobGroup, description)
         return S_OK()
 
+    #############################################################################
+    def setPriority(self, priority):
+        """Helper function.
+
+        Sets the job priority in the range 1-10. Higher numbers have more priority.
```
chrisburr:
```suggestion
        Sets the job priority in the range 1-10.
        Priorities 1-9 are matched probabilistically with higher numbers being more likely.
        Priority 10 jobs are always started before lower priority jobs from the same user.
```

chrisburr (approved):

PR #6855, opened 2023-02-23
[8.0] Use default VO File Catalog in the CLI 

atsareg:
The dirac-dms-filecatalog-cli command will work with the default File Catalog defined for the VO to which user belongs. If several catalogs are defined for a VO, the FileCatalog container object will be used instead.  This should meet the #6836 request.

BEGINRELEASENOTES

*DataManagement
NEW: dirac-dms-filecatalog-cli - use the default VO FileCatalog if not explicitly specified

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
+++ b/src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
@@ -20,50 +20,66 @@ Example:
 
   FC:/>
 """
-import sys
 
 from DIRAC.Core.Base.Script import Script
 
 
 @Script()
 def main():
-    fcType = "FileCatalog"
+    fcType = None
+    catalog = None
     Script.registerSwitch("f:", "file-catalog=", f"   Catalog client type to use (default {fcType})")
     Script.parseCommandLine(ignoreErrors=False)
 
-    from DIRAC import gConfig, exit as dexit
-    from DIRAC.Resources.Catalog.FileCatalogFactory import FileCatalogFactory
-
-    fcType = gConfig.getValue("/LocalSite/FileCatalog", "")
-
-    res = gConfig.getSections("/Resources/FileCatalogs", listOrdered=True)
-    if not res["OK"]:
-        dexit(1)
-    fcList = res["Value"]
-    if not fcType:
-        if res["OK"]:
-            fcType = res["Value"][0]
+    from DIRAC import exit as dexit
 
     for switch in Script.getUnprocessedSwitches():
         if switch[0].lower() == "f" or switch[0].lower() == "file-catalog":
             fcType = switch[1]
 
     if not fcType:
-        print("No file catalog given and defaults could not be obtained")
-        sys.exit(1)
+        # A particular catalog is not specified, try to instantiate the catalog container
+        from DIRAC.Resources.Catalog.FileCatalog import FileCatalog
+
+        catalog = FileCatalog()
+        if not catalog.valid:
+            print("Failed to create the FileCatalog container. Try to use a specific catalog with -f option")
+            dexit(-1)
+        result = catalog.getMasterCatalogNames()
+        if not result["OK"]:
+            print("Failed to get the Master catalog name for the FileCatalog container")
+            dexit(-1)
+        masterCatalog = result["Value"][0]
+        readCatalogs = [c[0] for c in catalog.getReadCatalogs()]
+        writeCatalogs = [c[0] for c in catalog.getWriteCatalogs()]
+        allCatalogs = list(set([masterCatalog] + readCatalogs + writeCatalogs))
+
+        if len(allCatalogs) == 1:
+            # If we have a single catalog in the container, let's use this catalog directly
+            fcType = allCatalogs[0]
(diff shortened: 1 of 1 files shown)


src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
```diff
@@ -20,50 +20,66 @@
...
 
 from DIRAC.Core.Base.Script import Script
 
 
 @Script()
 def main():
-    fcType = "FileCatalog"
+    fcType = None
+    catalog = None
     Script.registerSwitch("f:", "file-catalog=", f"   Catalog client type to use (default {fcType})")
```
andresailer: What is the default now that `fcType = None` ?

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
```diff
@@ -20,50 +20,66 @@
...
+        readCatalogs = [c[0] for c in catalog.getReadCatalogs()]
+        writeCatalogs = [c[0] for c in catalog.getWriteCatalogs()]
+        allCatalogs = list(set([masterCatalog] + readCatalogs + writeCatalogs))
+
+        if len(allCatalogs) == 1:
+            # If we have a single catalog in the container, let's use this catalog directly
+            fcType = allCatalogs[0]
+            catalog = None
+        else:
+            print("Starting FileCatalog container client:")
```
andresailer: Why does this say "starting", but is it actually starting anything?
And what is a "FileCatalog Container client"?
Can you add a comment here what is happening and why?

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
```diff
@@ -20,50 +20,66 @@
...
-    if not result["OK"]:
-        print(result["Message"])
-        if fcList:
-            print("Possible choices are:")
-            for fc in fcList:
-                print(" " * 5, fc)
-        sys.exit(1)
-    print(f"Starting {fcType} client")
-    catalog = result["Value"]
     cli = FileCatalogClientCLI(catalog)
```
andresailer: so if `catalog` is still None, it does what?

andresailer (commented):
Lots of questions...

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
atsareg: If the fcType is None, the usual logic of discovering the appropriate catalog defined for the VO to which the user belongs applies. The user's VO is determined based on the user's proxy information.

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
atsareg: OK, I'll try to make better comments and printouts

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
atsareg: ``catalog`` can not be None by this point. It is either a specific FileCatalog, e.g. AugerFileCatalog, or a FileCatalog() object which is containing several catalogs as defined for a VO, e.g. DIRAC File Catalog and Bookkeeping in LHCb. In the ILC I think you have also Transformation File Catalog. If the File Catalog resolution somehow failed, the command should have exited already before this point.

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
andresailer: Sorry for not being clear in this case: the docstring for the switch should be adapted `default {fcType}`

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
andresailer: Well, you added an `if catalog:` now! :grinning:

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
atsareg: Ah yes, you are right

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
atsareg: I added it just in case that I am not aware of :)

andresailer (approved):

src/DIRAC/DataManagementSystem/scripts/dirac_dms_filecatalog_cli.py
andresailer: Thanks! Things are clearer now!

PR #6859, opened 2023-02-27
[8.0] Rewrite DFC delete stored procedure

chaen:
The queries were very performant in MySQL 5.6, but ended up disastrous in 5.7.
In 8.0, they seem back to normal. 

BEGINRELEASENOTES

*DMS
CHANGE: refactor (DFC): rewrite the file/replica deletion stored procedure as they were in 5.6

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/DB/FileCatalogWithFkAndPsDB.sql
+++ b/src/DIRAC/DataManagementSystem/DB/FileCatalogWithFkAndPsDB.sql
@@ -903,47 +903,38 @@ BEGIN
     RESIGNAL;
   END;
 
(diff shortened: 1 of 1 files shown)


chrisburr:
I guess this is basically the same as dropping MySQL 5.7 support from v8.0? If so, are we ready to do that? I thought we were planning to drop it from v8.1 onwards.

chaen:
no, not quite, because it was likely a bug in earlier release of 5.7 that was fixed later down the road... :-) 
Also, I don't think anyone else uses these manager, so we are pretty free here :-)

fstagni (approved):

PR #6872, opened 2023-02-28
[8.0] Resolve methods of the REST interface

atsareg:
BEGINRELEASENOTES

*Core
FIX: TornadoREST - resolve interface method names for standard HTTP access methods 

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/TornadoREST.py
+++ b/src/DIRAC/Core/Tornado/Server/TornadoREST.py
@@ -171,10 +171,24 @@ class TornadoREST(BaseRequestHandler):  # pylint: disable=abstract-method
         """
         urls = []
         # Look for methods that are exported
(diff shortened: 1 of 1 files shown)


chaen:
Can you explain what you are trying to do and why ? Just as a reminder, this part of the code has posed big problems in the past already (https://github.com/DIRACGrid/DIRAC/pull/6258) 
And it was only visible on the extension. So it should first be tried on a v8.1 release such that the LHCbDIRAC hackaton can pick it.

atsareg:
This fix avoids exposing methods of the parent classes through the REST interface, e.g. methods of Tornado RequestHandler. Only methods in the handler implementation classes with names of the form <http_method>_<method_name> are routed by Tornado. This fix was done long ago and was running in the EGI production as hot fix that I was not aware of before.

atsareg:
I will make this fix for the integration branch only now. This will be then backported to 8.0 later after testing in certification including LHCb

PR #6873, opened 2023-03-01
[8.1] Resolve methods of the REST interface

atsareg:
This PR avoids exposing methods from parent classes of the REST handler implementation class. 

BEGINRELEASENOTES

*Core
FIX: TornadoREST - resolve interface method names for standard HTTP access methods

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/TornadoREST.py
+++ b/src/DIRAC/Core/Tornado/Server/TornadoREST.py
@@ -171,10 +171,24 @@ class TornadoREST(BaseRequestHandler):  # pylint: disable=abstract-method
         """
         urls = []
         # Look for methods that are exported
-        for prefix in [cls.METHOD_PREFIX] if cls.METHOD_PREFIX else cls.SUPPORTED_METHODS:
-            prefix = prefix.lower()
-            for mName, mObj in inspect.getmembers(cls, lambda x: callable(x) and x.__name__.startswith(prefix)):
-                methodName = mName[len(prefix) :]
+        for mName in cls.__dict__:
+            mObj = cls.__dict__[mName]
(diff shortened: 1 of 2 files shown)


src/DIRAC/Core/Tornado/Server/TornadoREST.py
```diff
@@ -171,10 +167,24 @@ def _pre_initialize(cls) -> list:
         """
         urls = []
         # Look for methods that are exported
-        for prefix in [cls.METHOD_PREFIX] if cls.METHOD_PREFIX else cls.SUPPORTED_METHODS:
-            prefix = prefix.lower()
-            for mName, mObj in inspect.getmembers(cls, lambda x: callable(x) and x.__name__.startswith(prefix)):
-                methodName = mName[len(prefix) :]
+        for mName in cls.__dict__:
```
chaen, line 170: If I understand properly, the only real difference between these two implementations is that you are looping over `__dict__` instead of `inspect.getmembers`, is that correct ? 
The logic otherwise seems the same ?

atsareg: Yes, the logic does not change, just extra unnecessary stuff does not get exposed

chrisburr: We already rejected the use of `__dict__`: https://github.com/DIRACGrid/DIRAC/pull/6128

chaen: yes but then the parent class is not exposed, so the extensions are broken

edit: oops, hadn't refreshed, so did not see @chrisburr comment

PR #6874, opened 2023-03-01
[8.1] TornadoREST - fix errors in the exposed methods resolution

atsareg:
This PR does not change the current implementation of TornadoREST but just fixes an error in the logic. Parent class methods are exposed (as before) in the REST interface. Replaces the attempt #6873   

BEGINRELEASENOTES

*Core
FIX: TornadoREST - fix the error in resolution of exposed methods

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/TornadoREST.py
+++ b/src/DIRAC/Core/Tornado/Server/TornadoREST.py
@@ -4,11 +4,15 @@ It directly inherits from :py:class:`tornado.web.RequestHandler`
 """
 
 import os
+import inspect
+from functools import partial
+from urllib.parse import unquote
+
 from tornado.escape import json_decode
 from tornado.web import url as TornadoURL
 
 from DIRAC.ConfigurationSystem.Client import PathFinder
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/Tornado/Server/TornadoREST.py
```diff
@@ -4,13 +4,9 @@
 """
 
 import os
-import inspect
```
chrisburr, line 7: Why? The `inspect` module is heavily used by this file.

chaen:
And sorry but I still don't understand what is the aim ? Are  you just exposing `get_method` and not `getMethod` ?

atsareg:
@chaen yes, I am exposing get_method and not getMethod. This is just a convention.

src/DIRAC/Core/Tornado/Server/TornadoREST.py
atsareg: It gets imported by

    from DIRAC.Core.Tornado.Server.private.BaseRequestHandler import *

It is better to import everything explicitly, I will change it

fstagni (approved):

PR #6886, opened 2023-03-06
[8.0 ] Implement actual service monitoring

chaen:
This PR contains bug fix and features

The main feature is that the activity monitoring now reports information that are actually useful (method name, status, duration), for both `dips` and `https`

It ACTUALLY does what was requested in https://github.com/DIRACGrid/DIRAC/issues/2138

```json
{
    "timestamp": 1678119467614,
    "Host": "lbvobox300.cern.ch",
    "ServiceName": "Bookkeeping_BookkeepingManager",
    "Location": "dips://lbvobox300.cern.ch:9200/Bookkeeping/BookkeepingManager",
    "ResponseTime": 0.3284897804260254,
    "MethodName": "getProductionProducedEvents",
    "Protocol": "dips",
    "Status": "OK"
  }
```

BEGINRELEASENOTES

*Core
FIX: Tornado PeriodicCallbacks are actually called
FEAT: Tornado Handlers report to ActivityMonitor
FEAT: Activity Monitoring reports useful information

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
+++ b/src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
@@ -20,9 +20,11 @@ from tornado.ioloop import IOLoop
 import DIRAC
 
 from DIRAC import gConfig, gLogger, S_OK, S_ERROR
+from DIRAC.ConfigurationSystem.Client.Helpers.Operations import Operations
 from DIRAC.Core.Utilities import DErrno
 from DIRAC.Core.DISET.AuthManager import AuthManager
 from DIRAC.Core.Utilities.JEncode import decode, encode
+from DIRAC.Core.Utilities import Network, TimeUtilities
 from DIRAC.Core.Utilities.ReturnValues import isReturnStructure
 from DIRAC.Core.Security.X509Chain import X509Chain  # pylint: disable=import-error
 from DIRAC.Resources.IdProvider.Utilities import getProvidersForInstance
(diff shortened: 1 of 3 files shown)


src/DIRAC/Core/Tornado/Server/private/BaseRequestHandler.py
```diff
@@ -662,26 +670,46 @@ def on_finish(self):
...
 
+        if self.activityMonitoringReporter:
+            record = {
+                "timestamp": int(TimeUtilities.toEpochMilliSeconds()),
+                "Host": Network.getFQDN(),
+                "ServiceName": "_".join(self._fullComponentName.split("/")),
+                "Location": self.request.uri,
+                "ResponseTime": elapsedTime,
+                # Take the method name from the POST call
+                "MethodName": self.request.arguments.get("method", ["Unknown"])[0],
```
chrisburr, line 706: Should we include the HTTP verb (POST/GET/...) as well? Either in the method name or as a separate field?

chaen: That was my original idea, but it will only ever be different from a `POST` when it's using `TornadoREST`. And there, it gets reaaaaaally messy..... so for the moment, that works with our standard `TornadoServices`, not for the `TornadoRest`

chrisburr: Ah yeah, the more I think about it the messier it gets...

chrisburr (approved):

fstagni (approved):

PR #6890, opened 2023-03-07
[8.0] Perform pilot operations with tokens

atsareg:
This PR resolves the issue of access to the pilot outputs, status, etc with tokens. 

BEGINRELEASENOTES

*Resources
CHANGE: (AREX)ComputingElement - use proxy only if it is set up

*WorkloadManagement
NEW: Enable PilotManager to manage pilots with tokens

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/ComputingElement.py
+++ b/src/DIRAC/Resources/Computing/ComputingElement.py
@@ -91,20 +91,13 @@ class ComputingElement:
 
     def _prepareProxy(self):
         """Set the environment variable X509_USER_PROXY"""
-        if not self.proxy:
-            result = getProxyInfo()
-            if not result["OK"]:
-                return S_ERROR("No proxy available")
-            if "path" in result["Value"]:
-                os.environ["X509_USER_PROXY"] = result["Value"]["path"]
-                return S_OK()
-        else:
+        if self.proxy:
             result = gProxyManager.dumpProxyToFile(self.proxy, requiredTimeLeft=self.minProxyTime)
             if not result["OK"]:
                 return result
             os.environ["X509_USER_PROXY"] = result["Value"]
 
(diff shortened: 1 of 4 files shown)


fstagni (commented):
Looks OK to me, maybe @aldbr wants to also add his review.

src/DIRAC/Resources/Computing/ComputingElement.py
```diff
@@ -91,20 +91,13 @@ def setToken(self, token, valid=0):
 
     def _prepareProxy(self):
         """Set the environment variable X509_USER_PROXY"""
-        if not self.proxy:
```
aldbr, line 94: Why do we completely remove this block?
I suspect that `self.proxy` is, or at least should be, always defined, is it correct?

Then may be we could return an error if it is not defined, such as:

```python
if not self.proxy:
    return S_ERROR("No proxy available")

# Continue with line95-line101
```

What do you think?

atsareg: If proxy is defined, self.proxy will not be None. This is simply a protection agains non-existing proxy which is a valid case

aldbr: > If proxy is defined, self.proxy will not be None

Ok fine.

But with your PR, if I execute `_prepareProxy()` without having a proxy, I will still get `S_OK()`.
It seems misleading.

atsareg: Why would you execute _prepareProxy without proxy ? This does not make sense in this case, that's why the protection. The outcome of _prepareProxy is irrelevant, it would mean a bug in the code that should be spotted while testing.

aldbr (approved):

fstagni (approved):

PR #6899, opened 2023-03-10
[8.0] fix: make -F (file) option work for drm command

### LINKED ISSUE #6896: COMDIRAC drm -F option broken
The -F option to drm complains that it didn't receive any arguments, even if they are provided. We'll fix this for v8r0 (although it'd be a pretty simple backport to the unmerged COMDIRAC for v7r3 if someone wanted to do that).

@IgorPelevanyuk We think this is the issue you mentioned in the BiLD meeting today.

marianne013:
BEGINRELEASENOTES

*Interfaces
FIX:  make -F (file) option work for drm command. This fixes #6896

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Interfaces/scripts/drm.py
+++ b/src/DIRAC/Interfaces/scripts/drm.py
@@ -3,12 +3,13 @@
 Remove files from the FileCatalog (and all replicas from Storage Elements)
 
 Examples:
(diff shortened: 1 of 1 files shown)


fstagni (approved):
When I see setters in python I always have the feeling that they can be avoided. Nevertheless...

Issue #6893, opened 2023-03-09
CA certificates for submitting htcondor with tokens


chrisburr:
In order to submit to many htcondor CEs the path the CAs has to be overriden. With X509 auth this was done with the `X509_CERT_DIR` environment variables however this seems to be ignored when submitting with tokens.

From checking it appears the only way to change the CAs is to set `AUTH_SSL_CLIENT_CADIR` in the htcondor config. This is not practical to do in DIRACOS so we need another solution.

@atsareg should contact the htcondor authors to see if there is a way to set this with an environment variable of if one can be added.

If this isn't possible the only solution will be to write a temporary file and set `CONDOR_CONFIG`.

atsareg:
The HTCondorCE developers confirmed that setting the _CONDOR_AUTH_SSL_CLIENT_CADIR env variable should be taken into account as expected. Although some past tests failed to do that, now it works indeed. I have set it up in the /opt/dirac/bashrc on lbcertifdirac70 and dropped from the condor_config and it works. So, the issue is closed.

chrisburr:
Shouldn't this by handled by the agent rather than requiring a specific variable to be set by admins?

atsareg:
Can be done in the agent indeed

chaen:
Please, please, please. Stop editing files on lbcertifdirac70. They are managed by puppet. Then things don't work anymore, and we wonder why, and start blaming in the wrong direction.

PR #6920, opened 2023-03-18
[8.0] Set CA directory location in HTCondorCE

atsareg:
Address #6893 issue

BEGINRELEASENOTES

*Resources
FIX: HTCondorCEComputingElement - set _CONDOR_AUTH_SSL_CLIENT_CADIR before executing condor commands

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
+++ b/src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
@@ -272,6 +272,7 @@ Queue stamp in %(pilotStampList)s
             htcEnv = {
                 "_CONDOR_SEC_CLIENT_AUTHENTICATION_METHODS": "SCITOKENS",
                 "_CONDOR_SCITOKENS_FILE": self.tokenFile.name,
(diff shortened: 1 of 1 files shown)


src/DIRAC/Resources/Computing/HTCondorCEComputingElement.py
```diff
@@ -272,6 +272,7 @@ def _executeCondorCommand(self, cmd, keepTokenFile=False):
             htcEnv = {
                 "_CONDOR_SEC_CLIENT_AUTHENTICATION_METHODS": "SCITOKENS",
                 "_CONDOR_SCITOKENS_FILE": self.tokenFile.name,
+                "_CONDOR_AUTH_SSL_CLIENT_CADIR": os.environ.get("X509_CERT_DIR", "/etc/grid-security/certificates"),
```
chrisburr: I guess this should use `getCAsLocation()` rather than relying on the variable.

atsareg: Indeed

chrisburr (approved):

PR #6921, opened 2023-03-20
[TBD][v7r3] test (CI): use github registry instead of docker

chaen:
Use github registry instead of docker
To be discussed, in the view of the new docker paid policy

BEGINRELEASENOTES

Please follow the template:
*Subsystem
NEW/CHANGE/FIX: explanation

ENDRELEASENOTES


chrisburr:
I agree with changing the default but I think we should be eligible for the opensource program: https://www.docker.com/community/open-source/application/

fstagni:
Can you check if there's something to be done also for _Pilot_ and _management_ repos?

fstagni:
https://www.docker.com/developers/free-team-faq/

fstagni:
This won't be needed, it seems.

PR #6940, opened 2023-03-28
[8.0] fix IAM integration tests

aldbr:
1.

The following error can happen if the docker container has an issue and fails to get a token:

```bash
Getting an IAM admin token
  File "/builds/lhcb-dirac/DIRAC/./integration_tests.py", line 1090, in <module>
    app()
  File "/builds/lhcb-dirac/DIRAC/./integration_tests.py", line 155, in create
    prepare_environment(flags, editable, extra_module, release_var)
  File "/builds/lhcb-dirac/DIRAC/./integration_tests.py", line 285, in prepare_environment
    _prepare_iam_instance()
  File "/builds/lhcb-dirac/DIRAC/./integration_tests.py", line 664, in _prepare_iam_instance
    admin_access_token = tokens.get("access_token")
UnboundLocalError: local variable 'tokens' referenced before assignment
```

In this PR, we raise an exception if there are too many failures instead.

2.

This problem can occur when a CI job running in a container is using a different network namespace. Therefore, commands interacting with IAM should be executed from one of the containers managed by `docker-compose`.

In this PR, we replace `requests` by `subprocess` to run the commands from a docker container.

Thanks @chrisburr for identifying this and fixing it.

BEGINRELEASENOTES
*Integration tests
FIX: Make error nicer when the IAM docker instance is not working properly
ENDRELEASENOTES

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -660,6 +660,8 @@ def _prepare_iam_instance():
             break
         except requests.ConnectionError:
             time.sleep(10)
+    else:
+        raise RuntimeError("All attempts to _get_iam_token failed ")
 
     admin_access_token = tokens.get("access_token")
 
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -4,7 +4,6 @@ import json
 import os
 from pathlib import Path
 import re
-import requests
 import shlex
 import shutil
 import subprocess
@@ -15,7 +14,6 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from typing import Optional
 
-import click
 import git
 import typer
 import yaml
@@ -647,7 +645,7 @@ def _prepare_iam_instance():
       * dirac/prod
       * dirac/user
     """
-    issuer = f"http://localhost:{IAM_PORT}"
+    issuer = f"http://iam-login-service:{IAM_PORT}"
 
     typer.secho("Getting an IAM admin token", fg=c.GREEN)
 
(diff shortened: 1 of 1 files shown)


aldbr:
Tests are failing in some contexts (CERN GitLab runners), I need to do things a bit differently.

integration_tests.py
```diff
@@ -705,18 +706,38 @@ def _get_iam_token(issuer, user, password, client_id, client_secret):
...
+        "-s",
+        "-u",
+        f"{client_id}:{client_secret}",
+        "-d",
+        "grant_type=password",
+        "-d",
+        f"username={user}",
+        "-d",
+        f"password={password}",
         query,
```
chrisburr: As this is repeated a lot, what do you think about abstracting this with something like:

```python
def iam_curl(url, *, data: list[str] = [], verb: Optional[str] = None, user = Optional[str], headers: list[str] = []):
    cmd = ["docker", "exec", "server", "curl", "-L", "-s"]
    if verb:
        cmd += ["-X", verb]
    if user:
        cmd += ["-u", user]
    for arg in data:
        cmd += ["-d", data]
    for header in headers:
        cmd += ["-H", header]
    cmd += [url]

    return subprocess.run(cmd, capture_output=True, check=False)
```

and then use it with something like:

```python
    ret = iam_curl(
        url,
        data=["grant_type=password", f"username={user}", f"password={password}"],
        user=f"{client_id}:{client_secret}",
    )
```

chrisburr (approved):

integration_tests.py
aldbr: Indeed, this is much cleaner, thanks!

PR #6950, opened 2023-03-29
[8.0] feat: use client token instead of user token for submission

aldbr:
With this PR, DIRAC can support the OAuth2 `client_credentials` flow.

- if the `username` is provided to the `TokenManagerClient`, the `TokenManagerHandler` tries to retrieve a user access token.
- else, the `TokenManagerHandler` contacts the IdP to get a client access token.

What's new in this PR:
- The `username` parameter becomes optional in `TokenManager(Client/Handler).getToken()`
- The `SiteDirector` only passes the `group` to the `TokenManagerClient`
- The `TokenManagerHandler` fetches a client access token based on the `group` parameter (using the underlying IdP).
- Code that was common to the `TokenManager(Client/Handler)` has been moved to a new module named `TokenManagementUtilities`. 

BEGINRELEASENOTES
*WorkloadManagementSystem
CHANGE: TokenManagerHandler provides client access token if username is not provided.
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -470,16 +470,7 @@ class SiteDirector(AgentModule):
 
         :return: S_OK/S_ERROR, Token object as Value
         """
-
-        result = Registry.getUsernameForDN(self.pilotDN)
-        if not result["OK"]:
-            return result
-        username = result["Value"]
-        result = gTokenManager.getToken(
-            username=username,
-            userGroup=self.pilotGroup,
-            requiredTimeLeft=3600,
-        )
+        result = gTokenManager.getToken(userGroup=self.pilotGroup, requiredTimeLeft=600)
         return result
 
     def _ifAndWhereToSubmit(self):
--- a/src/DIRAC/FrameworkSystem/Client/TokenManagerClient.py
+++ b/src/DIRAC/FrameworkSystem/Client/TokenManagerClient.py
@@ -8,6 +8,13 @@ from DIRAC.Core.Base.Client import Client, createClient
 from DIRAC.ConfigurationSystem.Client.Helpers import Registry
 from DIRAC.Resources.IdProvider.IdProviderFactory import IdProviderFactory
 from DIRAC.FrameworkSystem.private.authorization.utils.Tokens import OAuth2Token
+from DIRAC.FrameworkSystem.Utilities.TokenManagementUtilities import (
+    getIdProviderClient,
+    getCachedKey,
+    getCachedToken,
+    DEFAULT_AT_EXPIRATION_TIME,
+    DEFAULT_RT_EXPIRATION_TIME,
+)
 
 gTokensSync = ThreadSafe.Synchronizer()
 
@@ -15,8 +22,6 @@ gTokensSync = ThreadSafe.Synchronizer()
 @createClient("Framework/TokenManager")
 class TokenManagerClient(Client):
     """Client exposing the TokenManager Service."""
-
-    DEFAULT_RT_EXPIRATION_TIME = 24 * 3600
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -27,7 +32,7 @@ class TokenManagerClient(Client):
     @gTokensSync
     def getToken(
         self,
-        username: str,
+        username: str = None,
         userGroup: str = None,
         scope: str = None,
         audience: str = None,
@@ -45,42 +50,35 @@ class TokenManagerClient(Client):
 
         :return: S_OK(dict)/S_ERROR()
         """
-        if not identityProvider and userGroup:
-            identityProvider = Registry.getIdPForGroup(userGroup)
-        if not identityProvider:
-            return S_ERROR(f"The {userGroup} group belongs to a VO that is not tied to any Identity Provider.")
-
-        # prepare the client instance of the appropriate IdP
-        result = self.idps.getIdProvider(identityProvider)
+        # Get an IdProvider Client instance
+        result = getIdProviderClient(userGroup, identityProvider)
         if not result["OK"]:
             return result
         idpObj = result["Value"]
 
-        if userGroup and (result := idpObj.getGroupScopes(userGroup)):
-            # What scope correspond to the requested group?
-            scope = list(set((scope or []) + result))
+        # Search for an existing token in tokensCache
+        cachedKey = getCachedKey(idpObj.name, username, userGroup, scope, audience)
+        result = getCachedToken(self.__tokensCache, cachedKey, requiredTimeLeft)
+        if result["OK"]:
+            # A valid token has been found and is returned
+            return result
 
-        # Set the scope
-        idpObj.scope = " ".join(scope)
+        # No token in cache: get a token from the server
+        result = self.executeRPC(username, userGroup, scope, audience, idpObj.name, requiredTimeLeft, call="getToken")
 
-        # Let's check if there are corresponding tokens in the cache
-        cacheKey = (username, idpObj.scope, audience, identityProvider)
-        if self.__tokensCache.exists(cacheKey, requiredTimeLeft):
-            # Well we have a fresh record containing a Token object
-            token = self.__tokensCache.get(cacheKey)
-            # Let's check if the access token is fresh
-            if not token.is_expired(requiredTimeLeft):
-                return S_OK(token)
-
-        result = self.executeRPC(
-            username, userGroup, scope, audience, identityProvider, requiredTimeLeft, call="getToken"
-        )
-
+        # Save token in cache
         if result["OK"]:
             token = OAuth2Token(dict(result["Value"]))
+
+            # Get the date at which the token will expire
+            # If the refresh token is present, we use it as we can easily generate an access token from it
+            duration = token.get_claim("exp", "access_token") or DEFAULT_AT_EXPIRATION_TIME
+            if token.get("refresh_token"):
+                duration = token.get_claim("exp", "refresh_token") or DEFAULT_RT_EXPIRATION_TIME
+
             self.__tokensCache.add(
-                cacheKey,
-                token.get_claim("exp", "refresh_token") or self.DEFAULT_RT_EXPIRATION_TIME,
+                cachedKey,
+                duration,
                 token,
             )
 
--- a/src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
+++ b/src/DIRAC/FrameworkSystem/Service/TokenManagerHandler.py
@@ -9,7 +9,7 @@ requesting new tokens for DIRAC components that have the appropriate permissions
     :dedent: 2
     :caption: TokenManager options
 
-The most common use of this service is to obtain user tokens with certain scope to return to the user for its purposes,
+The most common use of this service is to obtain tokens with certain scope to return to the user for its purposes,
 or to provide to the DIRAC service to perform asynchronous tasks on behalf of the user.
 This is mainly about the :py:meth:`export_getToken` method.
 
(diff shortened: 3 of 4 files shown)


fstagni (commented):
I would be OK to merge this one in order to try in tomorrow's hackathon

src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
```diff
@@ -470,16 +470,7 @@ def __getPilotToken(self):
...
-        result = Registry.getUsernameForDN(self.pilotDN)
-        if not result["OK"]:
-            return result
-        username = result["Value"]
-        result = gTokenManager.getToken(
-            username=username,
-            userGroup=self.pilotGroup,
-            requiredTimeLeft=3600,
-        )
+        result = gTokenManager.getToken(userGroup=self.pilotGroup, requiredTimeLeft=600)
```
atsareg, line 473: I think it is better to make client or user token to use to be configurable. E.g. GenericPilotUser = client should trigger getToken with userName=None (or "client") which is interpreted as using Client.

chaen: Rather the other way around no ? If there is a username defined in the `SiteDirector` config, then use it, otherwise don't set it. But we should not have magic word like *client*

aldbr: I agree that a magic word sounds like a hack.
But I am not convinced about choosing what to use based on whether a username is defined in the `SiteDirector`.
We would have something like this in `SiteDirector.beginExecution()`:


```python

self.pilotUser = self.am_getOption("PilotUser", None)
# would need to change the logic in findGenericPilotCredentials() to potentially return a PilotUser
result = findGenericPilotCredentials(vo=self.vo, pilotDN=self.pilotDN, pilotUser=self.pilotUser, pilotGroup=self.pilotGroup) 
if not result["OK"]:
    return result
self.pilotUser, self.pilotDN, self.pilotGroup = result["Value"]
```

This seems tricky.
We would need both the `PilotDN` (for the proxy) and the `PilotUser` (for the token).
And they could be different.

Just relying on `PilotDN` and getting the `PilotUser` from it would not be a valid solution at the moment: we still need `PilotDN` to get a proxy for the pilots, so it has to be set in any case (which means we don't use the `client_credentials` flow).
We could also add a flag in the `SiteDirector` to know if we should use the username or not, but that seems inconvenient.

What would be the advantage of these solutions? Why would we need to submit pilots with a user access token? (at least for now we know that the client access tokens will be accepted by the sites).

aldbr: A summary of a discussion with @atsareg:

HTCondor sites want to know which token belongs to which VO.
The WLCG AuthZ group emits the idea that this could be extracted from the `issuer` claim by the sites.
VOs depending on Check In would have the same `issuer` claims so this would not be a good idea.

The WLCG AuthZ group suggested that they could have one instance but multiple issuer URLs (one per VO).
The Check In group did not accept this proposal as it would apparently go against the OIDC guidelines.
Another option for the sites, would be to use the `introspection` endpoint of the issuer to get more details about a given token.
This cannot be done for every job submission and would need to be done once and stored in a mapping file by the sites (this `issuer+subject` = this VO).
This is extra work to do for the sites, so there is no guarantee that they will accept.

They are working on a plugin to provide a common way to validate tokens.
This could be pushed at this level, but there are apparently no plans for that at the moment.
In any case, this needs some time to be developed.

This is why, at the moment, the SiteDirector should support both user and client access tokens.

chaen: I am not sure I have understood. 
As of today, what are the existing use cases that are REALLY used ? And I am not talking about what is in discussion, or could possibly come, etc. What is existing, and used ? `client_credential` is. 

The SiteDirector should implement only that, and not implement things that may potentially one day exist (and btw, that remark is true for everything, not just the SiteDirector)

aldbr: Personally, I would go with the current solution: the `SiteDirector` requests a client access token by not passing a username to the `TokenManager`.

Currently, Check In tokens are not supported by (HTCondor) sites yet and we are still not sure how these tokens will be handled (am I wrong saying this?).
But we do know that WLCG sites currently support WLCG client access tokens.
As @chaen said, I agree that we should only implement the code necessary to respond to existing use cases.

aldbr: A summary of a discussion with @atsareg reporting the progress occurring in Check In (please correct me if I am wrong):

- Check In is going to include the capability-based scopes (`compute`) in the access tokens.
- Check In is accepting the idea of employing client access tokens to submit jobs/pilots to sites. They are going to implement a way to associate a client to a VO. The information will probably be included in the `eduperson_entitlement` claim.
- Sites are currently accepting EGI Check In user access tokens (based on the `eduperson_entitlement` claim I guess) but not EGI Check In client access tokens (can you confirm @atsareg?) and this is why the `SiteDirector` should support both user and access tokens for the moment.

In this case, we would have the following logic defined by @chaen: if the `PilotUser` option is defined, then we use it to get a user access token, else we pass `username=None` to get a client access token.
By the way, should we remove the `PilotDN` parameter from the `SiteDirector` section and the `GenericPilotDN` parameter (in v8.1) from the `Operations` section to simplify the possibilities of making mistakes? The pilot DN could be obtained from `PilotUser`.
It could be implemented in this PR: https://github.com/DIRACGrid/DIRAC/pull/6566

chaen: Where can we actually see the status of token deployment and discussions in EGI CheckIn ?  
WLCG has public info on that, with precise milestones and documented minutes, so it is something we can work with. But we are completely blind with respect to CheckIn, and we can't work properly on hearsay.

atsareg: Update on the EGI Check-In. The Check-In related [doc](https://docs.google.com/document/d/1q67BWeFBmFmQ0crG5bVWwqecVMdN8PROYnciHO5JhRo/edit#heading=h.ropcuc1skhp5). The main points:

Check-In implemented compute scopes in the dev service that we are testing now.

Check-In agrees to allow client credentials tokens to be used to access CEs. These tokens should be obtained with dedicated clients that can get compute scopes (pilot clients). Those clients must not be used for user tokens manipulations. The pilot clients will eventually carry the VO/group claims in them. BUT the issuer claim will be the same for all the VO's. The VO/group claims might be analysed on the sites in special plugins in order to identify the user's VO/group and perform an appropriate mapping to local accounts. This means that eventually it will be enough to have a single DIRAC client to run pilots for all the VOs. This is not yet the case, for the moment it is necessary to have a separate client per VO. 

So, the bottom line is that in the EGI/Check-In provider we will have to have at least 2 configured clients. By the way, as far as I can see in the ATLAS and CMS cases, they have separate user IDs for pilot submission, for example, for SAM/ETF tests: [topology](https://topology.opensciencegrid.org/collaborations/osg-scitokens-mapfile.conf). So, if even they will start using clients, they will have different clients for different purposes.
That's why I think we have to foresee the possibility to define multiple clients per identity provider. This should be done in a way that if a single client is necessary, no extra configuration is needed. I have coded this in the EGI dev service and these are just several extra lines of code. I insist that this possibility is included. 

I suggest to implement it in the following way. The TokenManagerClient/TokenManagerService getToken() interface is using IdP object client_credentials if the username is "client". The IdPFactory.getIdentityProvider() interface is checking kwargs["client"] value and, if present, uses it as the clientName: if it is given and not None, the IdProvider configuration dictionary is updated by the contents of the /Resources/IdProviders/<IdProvider>/Clients/<clientName> section. In the particular client section client_id/client_secret are overriding the default values in the  /Resources/IdProviders/<IdProvider>/ section. Finally, the TokenManagerClient/Service instantiate the IdP object with the client key argument; its value is simply equal to the pilot user group, e.g. biomed_pilot. If there is now  /Resources/IdProviders/<IdProvider>/Clients/<clientName> section, the defaults from /Resources/IdProviders/<IdProvider> are taken.  So, this is a minimal implementation and for cases where it will be possible to use a single client for all purposes, no extra configuration will be needed. I have a working code of this working, nothing really special but it differs much from this PR which includes other code refactorings. So, we have to see how this can be merged if we agree.

atsareg: @aldbr: We can remove the PilotDN option indeed but in many installations it is used in the configurations. So, I would suggest to keep it just in order not to create a mess for various services. Or at least it should be well announced in the release notes and in the meetings

aldbr:
> These tokens should be obtained with dedicated clients that can get compute scopes (pilot clients). Those clients must not be used for user tokens manipulations. 

If I remember correctly, you said that EGI-CheckIn decided to prevent us to get a token with the offline scope from a client supporting the `client_credentials` flow because it is "not compliant with the OIDC specifications".
Therefore, we would need at least two clients: i) for user tokens manipulations; ii) for the pilots.

I have not found any detail about that in the OIDC specifications (I don't see any rule preventing a user to get a token with the offline scope from a client supporting the `client_credentials` flow). Do you have further details about that? Any public source of information supporting this statement?

> By the way, as far as I can see in the ATLAS and CMS cases, they have separate user IDs for pilot submission, for example, for SAM/ETF tests: topology. So, if even they will start using clients, they will have different clients for different purposes.

IIUC, ETF/SAM tests will need a specific LHCb IAM client, but it would not be integrated to the LHCbDIRAC instance, so I am not sure that this counts as a use case. 

> That's why I think we have to foresee the possibility to define multiple clients per identity provider. This should be done in a way that if a single client is necessary, no extra configuration is needed. I have coded this in the EGI dev service and these are just several extra lines of code. I insist that this possibility is included.
> The IdPFactory.getIdentityProvider() interface is checking kwargs["client"] value and, if present, uses it as the clientName: if it is given and not None, the IdProvider configuration dictionary is updated by the contents of the /Resources/IdProviders//Clients/ section. In the particular client section client_id/client_secret are overriding the default values in the /Resources/IdProviders// section. Finally, the TokenManagerClient/Service instantiate the IdP object with the client key argument; its value is simply equal to the pilot user group, e.g. biomed_pilot. If there is now /Resources/IdProviders//Clients/ section, the defaults from /Resources/IdProviders/ are taken. So, this is a minimal implementation and for cases where it will be possible to use a single client for all purposes, no extra configuration will be needed.

You proposed the following implementation:

```yaml
Resources
{
  IdProviders
  {
    IAM_LHCb
    {
      # Default configuration (e.g. to submit pilots)
      issuer = "https://lhcb-example.com"
      client_id = "lhcb_pilot_client_id"
      client_secret = "secret"

      # Possibility to have other clients
      # They would inherit from the IAM_LHCb parameters
      Clients
      {
        LHCb_User
        {
          client_id = "lhcb_user_client_id"
          client_secret = "secret2"
        }
      }
    }
  }
}
```

This solution might be a little bit confusing as we are mixing OAuth2 client parameters (client_id, client_secret) with OAuth2 IdProvider parameters (issuer).
I would suggest the following approach instead:

```yaml
Resources
{
  IdProviders
  {
    IAM_LHCb
    {
      issuer = "https://lhcb-example.com"
      DefaultClient = LHCb_pilot
      Clients
      {
        LHCb_User
        {
          client_id = "lhcb_user_client_id"
          client_secret = "secret2"
        }
        LHCb_Pilot
        {
          client_id = "lhcb_pilot_client_id"
          client_secret = "secret"
        }
      }
    }
  }
}
```

But here again, it seems that we are trying to solve a problem that is still evolving.
Therefore, I would suggest that we keep this PR as it is now, we merge it, and then we open a new one to modify the configuration structure.

> I suggest to implement it in the following way. The TokenManagerClient/TokenManagerService getToken() interface is using IdP object client_credentials if the username is "client".

As we agreed last time, I will not implement the possibility of using a user access token to submit pilots. It is pretty easy to add later if we really need that as you know, but so far it is not the case as we can all use client access tokens to submit pilots (btw, ATLAS and CMS are also using client access tokens for this use case).

chaen:
I agree, let's proceed with the currently existing use cases, and adapt if needed. The infrastructure is such that it can easily be modified if the need arises

atsareg:
OK, let's merge it as it is. I will need to make another PR to cope with the current EGI cases, but I would like to do it on top of this PR.

atsareg:
@aldbr : The configuration structure that I suggested above would allow to use "unnamed" client by default which will get its credentials from the IdProvider level. In the case of services with a single client needed, like presumed for LHCb for example, no Client section would be needed in the configuration at all avoiding unnecessary complication.

fstagni (approved):

PR #7040, opened 2023-06-05
[v8r0] Add /Resources/Computing/DefaultLocalCEType option

chrisburr:
Whne trying to enable SingularityCE by default for all queues I found that there doesn't seem to be an option for it? If this is the case I propose adding `/Resources/Computing/DefaultLocalCEType`. What do you think?

BEGINRELEASENOTES

*WorkloadManagement
NEW: Add /Resources/Computing/DefaultLocalCEType option

ENDRELEASENOTES


fstagni (approved):

andresailer:
I think it should also be documented.

PR #7044, opened 2023-06-09
[v8.0] Implement double slash in the Echo Xroot URL

chaen:
Echo now supports the standard double slash in thexroot  URL.

BEGINRELEASENOTES

*Resources
CHANGE: use double slash in Echo xroot URLs

ENDRELEASENOTES


chaen:
Successfully tested in prod at RAL

fstagni (approved):

PR #7061, opened 2023-06-19
[v8r0] Close DISET selector when we no longer need it

sfayer:
Hi,

We've noticed a case where the number of file descriptors used by DIRAC services & optimizer can slowly creep up: It seems this is caused by not explicitly closing the selector used by DISET (so we're relying on the garbage collector to delete the selector objects and close the eventpoll socket). We've seen nodes with >1000 FDs open for this, whereas with this patch the open FDs stays under 10.

Regards,
Simon

BEGINRELEASENOTES
FIX: Close DISET selector when we no longer need it
ENDRELEASENOTES


marianne013:
Tagging @martynia who is probably very  very interested in this fix ;-)

chrisburr (approved):

fstagni (approved):

PR #7077, opened 2023-06-24
[v8.0] fix (ReplicateAndRegister): better error reporting for files with no replicas

chaen:
BEGINRELEASENOTES

*DMS
FIX: better error reporting when an existing LFN has no replicas (at least it doesn't crash)

ENDRELEASENOTES


chaen:
@marianne013 @sfayer happy with that ?

sfayer:
Yes, this seems fine to me.

fstagni (approved):

PR #7135, opened 2023-07-27
[v8.0]Propagate more env variable in the SingularityCE

chaen:
Without this, CERN jobs started failing upload (could not find the CA, still investigating), and RAL job with oldish `ROOT` versions fail (redirector setup)

BEGINRELEASENOTES

*Resources
FIX: Propagate environment variables from host to container.

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/SingularityComputingElement.py
+++ b/src/DIRAC/Resources/Computing/SingularityComputingElement.py
@@ -14,6 +14,7 @@
 import io
 import json
 import os
+import re
 import shutil
 import sys
 import tempfile
@@ -83,6 +84,18 @@ chmod 644 retcode
 echo "Finishing inner container wrapper scripts at `date`."
 
 """
+
+
+ENV_VAR_WHITELIST = [
+    r"TERM",
+    r"VOMS_.*",
+    r"X509_.*",
+    r"XRD_.*",
+    r"Xrd.*",
+    r"DIRAC_.*",
+    r"BEARER_TOKEN.*",
+]
+ENV_VAR_WHITELIST = re.compile(r"^(" + r"|".join(ENV_VAR_WHITELIST) + r")$")
 
 
 class SingularityComputingElement(ComputingElement):
@@ -311,12 +324,17 @@ class SingularityComputingElement(ComputingElement):
         """Gets the environment for use within the container.
         We blank almost everything to prevent contamination from the host system.
         """
-        payloadEnv = {}
-        if "TERM" in os.environ:
-            payloadEnv["TERM"] = os.environ["TERM"]
+
+        if not self.__installDIRACInContainer:
+            payloadEnv = {k: v for k, v in os.environ.items() if ENV_VAR_WHITELIST.match(k)}
+        else:
+            payloadEnv = {}
+
         payloadEnv["TMP"] = "/tmp"
         payloadEnv["TMPDIR"] = "/tmp"
         payloadEnv["X509_USER_PROXY"] = os.path.join(self.__innerdir, "proxy")
+        payloadEnv["DIRACSYSCONFIG"] = os.path.join(self.__innerdir, "pilot.cfg")
+
         return payloadEnv
 
     @staticmethod


src/DIRAC/Resources/Computing/SingularityComputingElement.py
```diff
@@ -85,6 +86,18 @@
 """
 
 
+ENV_VAR_WHITELIST = [
+    r"TERM",
+    r"VOMS_.*",
+    r"X509_.*",
+    r"XRD_.*",
+    r"Xrd.*",
+    r"DIRAC.*",  # will take also DIRAC and DIRACOS
```
chrisburr: This is probably an issue

sfayer:
Hi,

We're having related problems: In no-install mode, the DIRACSYSCONFIG variable containing the pilot.cfg location doesn't get pulled through to the innner container so none of the command line tools work... It sounds like this would fix that (assuming pilot.cfg is actually somewhere in the bit that's mounted in the container, I haven't checked that yet).

However, I suspect some of the variables will be wrong if __installDIRACInContainer is True as the paths outside & inside the container may not match up, so perhaps it should only happen if not __installDIRACInContainer (or maybe different sets of vars depending on the mode)?

Regards,
Simon

src/DIRAC/Resources/Computing/SingularityComputingElement.py
chaen: then we can take `DIRAC_`

chaen:
> However, I suspect some of the variables will be wrong if __installDIRACInContainer is True as the paths outside & inside the container may not match up, so perhaps it should only happen if not __installDIRACInContainer (or maybe different sets of vars depending on the mode)?

I think that makes sense

chrisburr:
> perhaps it should only happen if not __installDIRACInContainer (or maybe different sets of vars depending on the mode)?

Perhaps, though I wonder if we can drop the `__installDIRACInContainer` option?

sfayer:
> > perhaps it should only happen if not __installDIRACInContainer (or maybe different sets of vars depending on the mode)?
> 
> Perhaps, though I wonder if we can drop the `__installDIRACInContainer` option?

We were planning on switching over, but as I say, we ran into pilot.cfg issues and are back running in install mode for now: If we can fix those then I don't see it being a problem.

chaen:
How about treating this `pilot.cfg` like `X509_USER_PROXY` and copy it in the work area ?

sfayer:
> How about treating this `pilot.cfg` like `X509_USER_PROXY` and copy it in the work area ?

Yes, I think that would work.

chaen:
> > How about treating this `pilot.cfg` like `X509_USER_PROXY` and copy it in the work area ?
> 
> Yes, I think that would work.

In fact, isn't it what is already done here  ? 
https://github.com/DIRACGrid/DIRAC/blob/3ab57e91cc4de7841551e79619aef1bb7c55b51e/src/DIRAC/Resources/Computing/SingularityComputingElement.py#L261-L285

sfayer:
> > > How about treating this `pilot.cfg` like `X509_USER_PROXY` and copy it in the work area ?
> > 
> > 
> > Yes, I think that would work.
> 
> In fact, isn't it what is already done here ?
> 
> https://github.com/DIRACGrid/DIRAC/blob/3ab57e91cc4de7841551e79619aef1bb7c55b51e/src/DIRAC/Resources/Computing/SingularityComputingElement.py#L261-L285

It looks that way... In which case it just needs the environment variable setting correctly inside the container to point at the copy...

chaen:
You guys had so many trouble with this `pilot.cfg` not being found so I am not even sure which one is the correct one :laughing: 
Setting `DIRACSYSCONFIG` to the `pilot.cfg` would solve the issue ?

sfayer:
> You guys had so many trouble with this `pilot.cfg` not being found so I am not even sure which one is the correct one laughing Setting `DIRACSYSCONFIG` to the `pilot.cfg` would solve the issue ?


Yes please... DIRACSYSCONFIG should be set to the full "inside the container" path to pilot.cfg.

chaen:
Does that seem reasonable to everybody ?

sfayer:
Looks good to me.

chaen:
we will try it in the LHCb prod environment

chrisburr (approved):

chaen:
It works just fine in LHCb. @sfayer @marianne013 did you have a chance at trying it ?

marianne013:
We ran it on the preprod server using the tag directly from your repo. It worked, otherwise you would have heard about it :-D.  Whether we covered all use cases remains to be seen.

chaen:
:laughing: OK, merging then

PR #7140, opened 2023-07-28
[v8r0] doc: replace sphinx panels with sphinx design (its successor)

andresailer:
This unblocks the sphinx < 5 version constraint from sphinx panels

https://sphinx-design.readthedocs.io/en/furo-theme/get_started.html#migrating-from-sphinx-panels

BEGINRELEASENOTES

*docs:
change: Replace the use of sphinx-panels with sphinx-design, this unblocks the use of latest sphinx versions (panels was requiring sphinx < 5)

ENDRELEASENOTES

- [x] Still need to replace panels somewhere!?!

### DIFF
--- a/docs/setup.py
+++ b/docs/setup.py
@@ -29,5 +29,5 @@ setup(
     package_dir=PACKAGE_DIR,
     packages=ALL_PACKAGES,
     scripts=SCRIPTS,
-    install_requires=["sphinx_rtd_theme", "sphinx_panels"],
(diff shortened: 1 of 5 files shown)


andresailer:
https://dirac--7140.org.readthedocs.build/en/7140/

Pictures are a bit bigger, and hover behaviour is different, but I don't want to delve deeper into this to be honest...

sphinx_rtd_theme is requiring sphinx<7, but at least sphinx is now 6.2.1.

fstagni (approved):

PR #7166, opened 2023-08-10
[v8r0] Stop using PID namespaces with SingularityCE

chrisburr:
When sites use docker to isolate the pilot it's not possible to use both PID namespaces and mount /proc inside the container created by SingularityCE without the site starting the docker container with --privileged. See https://github.com/apptainer/singularity/issues/1768 for details.

I'm not convinced we gain much by passing --pid so I propose removing it.

BEGINRELEASENOTES

*WorkloadManagement
CHANGE: Stop using PID namespaces with SingularityCE

ENDRELEASENOTES


sfayer (approved):
This seems fine to me: We only added the PID namespace as a "nice to have", so if it's causing problems there is no need to keep it.

Regards,
Simon

fstagni (approved):

Issue #7183, opened 2023-08-30
Core dump on long running queries


chaen:
When running long lasting queries against a DB on a loaded server, we get a core dump of python. The cut is exactly at 600 seconds. 
When running just that query on an isolated server, it works fine. 
I did not have time to dig deep into yet, but LHCb is impacted in production

The client just receives a blunt 
```
"Error in _read: (104, 'Connection reset by peer') SSLError(104, 'Connection reset by peer')"
```

The full stacktrace is attached, but here are the relevant info

```
*** Error in `/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/bin/python3.11': double free or corruption (out): 0x00007fcbfc0acba0 ***
*** Error in `/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/bin/python3.11': corrupted double-linked list: 0x00007fcbfc093ce0 ***
======= Backtrace: =========
======= Backtrace: =========
/lib64/libc.so.6(+0x81329)[0x7fcc3e95f329]
/lib64/libc.so.6(+0x8097f)[0x7fcc3e95e97f]
/lib64/libc.so.6(+0x8120e)[0x7fcc3e95f20e]
/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/lib/python3.11/site-packages/MySQLdb/../../../libmysqlclient.so.21(+0x8d904)[0x7fcc2fa13904]
/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/lib/python3.11/site-packages/MySQLdb/../../../libmysqlclient.so.21(+0x304bb)[0x7fcc2f9b64bb]
/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/lib/python3.11/site-packages/MySQLdb/../../../libmysqlclient.so.21(+0x3551c)[0x7fcc2f9bb51c]
/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/lib/python3.11/site-packages/M2Crypto/../../../libssl.so.3(SSL_free+0x14e)[0x7fcc36d9780e]
/opt/dirac/versions/v11.0.18-1691646408/Linux-x86_64/lib/python3.11/site-packages/MySQLdb/../../../libmysqlclient.so.21(+0x87ca9)[0x7fcc2fa0dca9]
[...]

PR #7184, opened 2023-08-30
[8.0] feat (TransformationCleaningAgent): chunk the file removals

chaen:
For https://github.com/DIRACGrid/DIRAC/issues/7183

BEGINRELEASENOTES

*TS
CHANGE: TransformationCleaningAgent removes files by chunk

ENDRELEASENOTES


andresailer:
What about forwarding the file removal to the RMS?

chaen:
I agree it would be better but probably a bit harder to implement as we need to make sure not to re-send requests from one cycle to the next.
Are you ok with having it merged like this, and we create a task to turn this into requests ?

andresailer:
Yes, fine to do the changes here.

andresailer (approved):

fstagni (approved):

PR #7199, opened 2023-09-14
[v8r0] Drop TS permission check for all read-only functions

### LINKED ISSUE #7198: New permissions in the transformation system
In https://github.com/DIRACGrid/DIRAC/pull/7113/ access to `TransformationClient.getTransformation` was restricted however access to `TransformationClient.getTransformations` was not.

Though in the actual method, it seems at a glance that `getTransformation` is just an wrapper around `getTransformations` so it's not clear to me why they wouldn't be treated the same way:

https://github.com/DIRACGrid/DIRAC/blob/df6a89f02aa6c9f04dc6891cd0a267e6699e3ce9/src/DIRAC/TransformationSystem/DB/TransformationDB.py#L318-L327

It's the same for `getTransformationParameters`. 

The WebApp also now silently fails due to accessing these methods as well as `getTransformationStats`, `getTransformationFilesCount` and potentially more.

I guess more importantly the original PR stated:

> we decided that we'll be fine with people seeing other transformations - just not modifying them.

but in practice the change restricts large parts of the "seeing other transfromations". This is somewhat pressing as LHCbDIRAC extensively relies on users having read access to all transformations.

sfayer:
Hi,

We got a bit carried away with the permissions checks and added them to some of the read-only functions, which breaks some things. This patch removes the checks from all of the get* functions.

I think this one should sweep OK as the changes needed in integration are identical.

Regards,
Simon

Fixes #7198 .

BEGINRELEASENOTES
*TransformationSystem
FIX: Drop TS permission check for all read-only functions
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/TransformationSystem/Service/TransformationManagerHandler.py
+++ b/src/DIRAC/TransformationSystem/Service/TransformationManagerHandler.py
@@ -207,15 +207,11 @@ class TransformationManagerHandlerMixin:
         tfDetails = self.transformationDB.getTransformation(transName, extraParams=extraParams)
         if not tfDetails["OK"]:
             return tfDetails
-        if not (result := self.checkPermissions(transName))["OK"]:
-            return result
         return tfDetails
 
     types_getTransformationParameters = [[int, str], [str, list]]
 
     def export_getTransformationParameters(self, transName, parameters):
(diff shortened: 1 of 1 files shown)


chrisburr (approved):
Thanks for taking a look at this so promptly.

fstagni (approved):

Issue #7207, opened 2023-09-19
Pilot submission with tokens with multi-VO setup


aldbr:
Following the discussion happening in https://github.com/DIRACGrid/DIRAC/pull/7176, I decided to create a proper issue.

The current logic does not support the multi-VO case: a Site Director will submit pilots with tokens to every CE embedding the `Token` tag. Problem: if a given VO is not yet ready to use tokens, then the VO will not be able to submit pilots at all.

Solutions:
1. https://github.com/DIRACGrid/DIRAC/pull/7176#issuecomment-1723411324
2. https://github.com/DIRACGrid/DIRAC/pull/7176#issuecomment-1723520992

PR #7208, opened 2023-09-19
[8.0] feat: pilot submission with tokens in a  multi-VO setup

aldbr:
Aims at solving https://github.com/DIRACGrid/DIRAC/issues/7207
This is the implementation of the second solution. 

Pros: very easy to implement as you can see
Cons: it does not work if a given CE accepts tokens from VO "A" but not from VO "B", while they are both ready to use tokens. Can it happen?

BEGINRELEASENOTES
*WorkloadManagement
CHANGE: pilot submission with tokens in a multi-VO context
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -471,10 +471,10 @@ class SiteDirector(AgentModule):
     def __supportToken(self, ce: ComputingElement) -> bool:
         """Check whether the SiteDirector is able to submit pilots with tokens.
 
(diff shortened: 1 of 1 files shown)


aldbr:
I changed the approach.
Now you can specify `Tag = Token:<VO>` to make sure that only a given set of VO will use tokens in a CE.

Pros: very easy to implement as you can see, DIRAC sysadmins keep control of what is done.
Cons: needs a lot of actions from the DIRAC sysadmins.

There was another approach requiring no change of the code, which would have consisted in copying each queue (one with `Tag = Token`, the other without it). The approach was consistent but hard to manage, @marianne013 and I decided not to go through this path.

marianne013 (approved):
This looks like what we agreed on.

fstagni (approved):

PR #7210, opened 2023-09-20
[8.0] fix: CI tests

aldbr:
Following https://github.com/DIRACGrid/DIRAC/pull/7209

BEGINRELEASENOTES
*Core
FIX: replace ColorBar.draw_all() with figure.draw_without_rendering()
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/Graphs/QualityMapGraph.py
+++ b/src/DIRAC/Core/Utilities/Graphs/QualityMapGraph.py
@@ -168,10 +168,10 @@ class QualityMapGraph(PlotBase):
         setp(self.ax.get_yticklines(), markersize=0.0)  # pylint: disable=not-callable
 
         cax, kw = make_axes(self.ax, orientation="vertical", fraction=0.07)
-        cb = ColorbarBase(
+        ColorbarBase(
             cax, cmap=self.cmap, norm=self.norms, boundaries=self.cbBoundaries, values=self.cbValues, ticks=self.cbTicks
         )
-        cb.draw_all()
+        self.figure.draw_without_rendering()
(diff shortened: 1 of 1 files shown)


src/DIRAC/Core/Utilities/Graphs/QualityMapGraph.py
```diff
@@ -168,10 +168,10 @@ def draw(self):
         setp(self.ax.get_yticklines(), markersize=0.0)  # pylint: disable=not-callable
 
         cax, kw = make_axes(self.ax, orientation="vertical", fraction=0.07)
-        cb = ColorbarBase(
+        ColorbarBase(
```
andresailer, line 171: Does this statement still do anything, or just create an object that is not used?

aldbr: This statement is actually still useful, I just made a quick test in the certification environment: 

- Without it:
image

- With it:
image


But the line that I replaced seems to have no effect, I have not spotted any change when I removed it: this is true for both `cb.draw_all()` and `self.figure.draw_without_rendering()`.
I would let it here for now as it might affect something I can't see.

fstagni (approved):

src/DIRAC/Core/Utilities/Graphs/QualityMapGraph.py
andresailer: Thanks for checking and explaining! I wanted to say, maybe add a comment.

aldbr:
@fstagni do you know why @DIRACGridBot added the `sweep:ignore` label to this PR?

PR #7215, opened 2023-09-22
[8.1] Add VO in Jobs table

simon-mazenoux:
See discussion #7197 for further information.

This will require DB changes.
- [x] Add to wiki - in https://github.com/DIRACGrid/DIRAC/wiki/DIRAC-8.1#wms

BEGINRELEASENOTES

*WMS
CHANGE: add VO info in the table Jobs of JobDB

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
+++ b/src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
@@ -11,33 +11,26 @@ The following options can be set in ``Systems/WorkloadManagement/<Setup>/Databas
 * *CompressJDLs*:        Enable compression of JDLs when they are stored in the database, default *False*.
 
 """
-import base64
-import zlib
 import datetime
-
 import operator
 
 from DIRAC.ConfigurationSystem.Client.Helpers.Registry import getVOForGroup
-from DIRAC.ConfigurationSystem.Client.Helpers.Operations import Operations
 from DIRAC.ConfigurationSystem.Client.Helpers.Resources import getSiteTier
 from DIRAC.Core.Base.DB import DB
 from DIRAC.Core.Utilities import DErrno
 from DIRAC.Core.Utilities.ClassAd.ClassAdLight import ClassAd
-from DIRAC.Core.Utilities.ReturnValues import S_OK, S_ERROR, convertToReturnValue
-from DIRAC.Core.Utilities.DErrno import EWMSSUBM, EWMSJMAN, cmpError
-from DIRAC.Core.Utilities.ObjectLoader import ObjectLoader
+from DIRAC.Core.Utilities.DErrno import EWMSJMAN, EWMSSUBM, cmpError
+from DIRAC.Core.Utilities.ReturnValues import S_ERROR, S_OK
 from DIRAC.ResourceStatusSystem.Client.SiteStatus import SiteStatus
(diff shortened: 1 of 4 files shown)


src/DIRAC/WorkloadManagementSystem/DB/JobDB.py
```diff
@@ -908,6 +908,7 @@ def insertNewJobIntoDB(
             "SubmissionTime": str(datetime.datetime.utcnow()),
             "Owner": owner,
             "OwnerGroup": ownerGroup,
+            "VO": getVOForGroup(ownerGroup),
```
aldbr: Is that enough?
Don't you have to add the `VO` field in `JobDB.sql` as well?

simon-mazenoux: Yes, you're correct.

src/DIRAC/WorkloadManagementSystem/DB/JobDB.sql
```diff
@@ -42,6 +42,7 @@ CREATE TABLE `Jobs` (
   `JobName` VARCHAR(128) NOT NULL DEFAULT 'Unknown',
   `Owner` VARCHAR(64) NOT NULL DEFAULT 'Unknown',
   `OwnerGroup` VARCHAR(128) NOT NULL DEFAULT 'Unknown',
+  `VO` VARCHAR(32),
```
aldbr: Why not: 
```suggestion
  `VO` VARCHAR(32) NOT NULL DEFAULT 'Unknown',
```

simon-mazenoux: By not having a default value, a query trying to insert a row without the VO field will raise an `IntegrityError` in sqlalchemy, which will quickly indicate to the developer that something is wrong. Besides, now that the VO is used to get information from the config and that this field can no longer be determined by the `OwnerGroup`, having values as "Unknown" could be problematic.

aldbr: > a query trying to insert a row without the VO field will raise an IntegrityError in sqlalchemy

Would it raise an `IntegrityError` if the column has no `NOT NULL` constraint?

simon-mazenoux: Yes, you're right. I'll correct it.

simon-mazenoux: In the end I ended up adding the DEFAULT 'Unknown' because the _update method in DIRAC actually translates to an UPSERT, which causes an error because of the insert statement:

```
2023-09-27T08:34:51,221683Z Framework/WorkloadManagement/JobDB DEBUG: _update: INSERT INTO Jobs (JobID, Status) VALUES (3, "Received") ON DUPLICATE KEY UPDATE Status=VALUES(Status)
2023-09-27T08:34:51,221956Z Framework/WorkloadManagement/JobDB ERROR: _update (INSERT INTO Jobs (JobID, Status) VALUES (3, "Received") ON DUPLICATE KEY UPDATE Status=VALUES(Status)): Execution failed. 1364: Field 'VO' doesn't have a default value
```

aldbr (approved):

simon-mazenoux:
@aldbr Would you mind also checking the changes I made to the wiki page ? (I don't know if you did it already)

fstagni (approved):

PR #7245, opened 2023-10-12
fix (dirac-login): write diracx token correctly

chaen:
BEGINRELEASENOTES

*Core
FIX: correctly dump diracx token when doing dirac-login

ENDRELEASENOTES


chrisburr:
This is already in #7244

PR #7284, opened 2023-11-09
fix: disabledVOs not available when doing dirac-configure

aldbr:
"chicken and egg" situation here.
When doing `dirac-configure`, the CS is not yet available to the user, so we cannot get the VO and the `DisabledVOs` value.
Therefore, `DisabledVOs` is equal to `[]` and the condition is always `True`, which triggers calls to `diracx`, which is not available.

I proposed the following "hack" to perform the DIRAC hackathon.
The problem with that approach is that you would need to perform `dirac-proxy-init` again, after `dirac-configure`, to get a token along with a proxy.

BEGINRELEASENOTES
*Framework
FIX: disabledVOs not available when doing dirac-configure
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/FrameworkSystem/scripts/dirac_proxy_init.py
+++ b/src/DIRAC/FrameworkSystem/scripts/dirac_proxy_init.py
@@ -240,7 +240,7 @@ class ProxyInit:
 
         vo = Registry.getVOMSVOForGroup(self.__piParams.diracGroup)
         disabledVOs = gConfig.getValue("/DiracX/DisabledVOs", [])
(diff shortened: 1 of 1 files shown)


chrisburr:
We already do it a second time so I think this makes sense as there is the same problem with adding VOMS extensions and uploadijng the proxy:

https://github.com/DIRACGrid/DIRAC/blob/50ff1ababd9000628c36a60336f56f8ed598a7e4/src/DIRAC/Core/scripts/dirac_configure.py#L286

chrisburr (approved):

PR #7304, opened 2023-11-23
[8.0] Use copy+remove in PilotSync agent to avoid SELinux problems

sfayer:
Hi,

We've been switching to using an external webserver to serve the pilot files as things were getting overloaded, but ran into SELinux problems. This small tweak changes the deploying of the pilot files from move to copy and delete; this causes the SELinux context of the files to get reset to the correct values for the destination.

Regards,
Simon

BEGINRELEASENOTES
*WorkloadManagement
FIX: Use copy+remove in PilotSync agent to avoid SELinux problems
ENDRELEASENOTES


fstagni (approved):

chaen:
maybe in the `__init__.py` of `DIRAC` we should have `subprocess.run("setenforce permissive")` ? :thinking:

Issue #7346, opened 2023-12-06
[diracx] Jobs stuck in the `Matched` status


aldbr:
You may have noticed that all the jobs we submitted in the certification environment have been stuck in a `Matched` status.
By taking a look at a few pilot outputs, I noticed that the `JobWrapper` was unable to report the status of the jobs:

```
2023-12-06T12:10:20,425314Z None/[7879]JobWrapper WARN: Failed setting job status ValueError('No dirax token in the proxy file /var/lib/condor/execute/dir_20392/tmpjubbjrey')
```

As you can see, the `JobWrapper` tries to interact with diracx and fails because the diracx token is not embedded in the user proxy. But should it be? 

- The user proxy used (/var/lib/condor/execute/dir_20392/tmpjubbjrey) is not supposed to interact with diracx: 

```
subject      : ...
issuer       : ...
identity     : ...
timeleft     : 23:59:58
DIRAC group  : dteam_user
DiracX       : False
path         : /var/lib/condor/execute/dir_20392/tmpjubbjrey
username     : ...
properties   : NormalUser
VOMS         : True
VOMS fqan    : ['/dteam']
```

- The configuration confirms that the VO `dteam` should not interact with diracx:

```
DiracX
{
  DisabledVOs = dteam, ...
}
```

So why does DIRAC tries to interact with diracx anyway? 
Because the `ClientSelector` in charge of choosing the service to use relies on `useLegacyAdapter`, which only checks if the service to interact with is allowed in the configuration (which is the case):

https://github.com/DIRACGrid/DIRAC/blob/01491bfad9a12740baaf44c83241d1f35e8d75f6/src/DIRAC/ConfigurationSystem/Client/PathFinder.py#L251-L261

```
DiracX
{
  LegacyClientEnabled
  {
    WorkloadManagement
    {
      JobStateUpdate = True
    }
  }
}
```

An easy fix would consist in extracting the VO from the proxy and checking whether it is part of the disabled VOs.
How to reproduce the issue:

```python
from DIRAC.WorkloadManagementSystem.Client.JobStateUpdateClient import JobStateUpdateClient

from DIRAC.Core.Base.Script import Script
Script.parseCommandLine()

result = JobStateUpdateClient().setJobSite(123, "LCG.LOL")
print(result)
```

chrisburr:
After discussing about this I don't think it's feasible:

* loading the proxy during the client selector will be slow
[...]

PR #7347, opened 2023-12-06
[9.0] fix: useLegacyAdapter should check the VO

aldbr:
Aims at solving https://github.com/DIRACGrid/DIRAC/issues/7346

BEGINRELEASENOTES
*Configuration
FIX: useLegacyAdapter now checks the VO
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/ConfigurationSystem/Client/PathFinder.py
+++ b/src/DIRAC/ConfigurationSystem/Client/PathFinder.py
@@ -248,7 +248,7 @@ def getServiceURLs(system, service=None, setup=False, failover=False):
     return resList
 
 
-def useLegacyAdapter(system, service=None) -> bool:
+def useLegacyAdapter(system, service=None, useCertificates=None) -> bool:
     """Should DiracX be used for this service via the legacy adapter mechanism
 
     :param str system: system name or full name e.g.: Framework/ProxyManager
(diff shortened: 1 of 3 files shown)


src/DIRAC/ConfigurationSystem/Client/PathFinder.py
```diff
@@ -256,9 +256,20 @@ def useLegacyAdapter(system, service=None) -> bool:
 
     :return: bool -- True if DiracX should be used
     """
+    from DIRAC.Core.Security.ProxyInfo import getProxyInfo
+
+    # Check if DiracX is enabled for this service
     system, service = divideFullName(system, service)
     value = gConfigurationData.extractOptionFromCFG(f"/DiracX/LegacyClientEnabled/{system}/{service}")
-    return (value or "no").lower() in ("y", "yes", "true", "1")
+    isServiceEnabled = (value or "no").lower() in ("y", "yes", "true", "1")
```
chrisburr: How about returning quickly and not calling `getProxyInfo` if it's pointless:

```suggestion
    if (value or "no").lower() not in ("y", "yes", "true", "1"):
        return False
```

src/DIRAC/ConfigurationSystem/Client/PathFinder.py
```diff
@@ -256,9 +256,20 @@ def useLegacyAdapter(system, service=None) -> bool:
...
+    # Check if DiracX is enabled for this VO
+    result = getProxyInfo()
```
chrisburr, line 274: What happens if we're using a host certificate rather than a proxy?

aldbr: How does `diracx` handle requests coming from a host?

chrisburr:
See https://github.com/DIRACGrid/DIRAC/issues/7346

PR #7394, opened 2024-01-15
[8.0] Correctly collect platform list from queues

sfayer:
Hi,

I think the platform matching logic in the 8.0 site director isn't quite right: The self.platforms list is initialised in the queue loop, so it only ever gets set to the platform of the last queue considered. This patch moves it out of the loop so that all of the platforms for all queues supporting the VO are collected.

(Please can you add a no sweep label to this: I think the 9.0 sitedirector has had all of the platform code removed?)

Regards,
Simon

BEGINRELEASENOTES
*WorkloadManagementSystem
FIX: Correctly collect platform list from queues
ENDRELEASENOTES


aldbr (approved):
Looks good to me!

PR #7425, opened 2024-01-26
[integration] add a wnVO argument to sendMessage and finaliseLogs

martynia:
BEGINRELEASENOTES

*WorkloadManagementSystem
NEW: TornadoPilotLoggingHandler modify the handler to accept VO name sent by a pilot. Required in a case where the VO cannot be guessed from a proxy. This change requires https://github.com/DIRACGrid/Pilot/pull/230

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/Service/TornadoPilotLoggingHandler.py
+++ b/src/DIRAC/WorkloadManagementSystem/Service/TornadoPilotLoggingHandler.py
@@ -50,7 +50,7 @@ class TornadoPilotLoggingHandler(TornadoService):
             os.makedirs(logPath)
         cls.log.verbose("Pilot logging directory:", logPath)
 
-    def export_sendMessage(self, message, pilotUUID):
+    def export_sendMessage(self, message, pilotUUID, wnVO):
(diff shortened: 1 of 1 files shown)


src/DIRAC/WorkloadManagementSystem/Service/TornadoPilotLoggingHandler.py
```diff
@@ -67,6 +68,9 @@ def export_sendMessage(self, message, pilotUUID):
 
         # determine client VO
         vo = self.__getClientVO()
+        if vo == "":
+            vo = wnVO
```
fstagni, line 72: In theory, we *could* have the case where vo is different from wnVO...

martynia:
Well, if it’s only a matter of a variable name, that’s not a big problem.
We have to be sure what value to pass from the pilot end..
  J.

fstagni (approved):

fstagni:
> Well, if it’s only a matter of a variable name, that’s not a big problem. We have to be sure what value to pass from the pilot end.. J.

It's probably not an issue... approved and merged.

PR #7451, opened 2024-02-08
[9.0] fix: dirac-proxy-init should not print the group anymore

aldbr:
Related to: https://github.com/DIRACGrid/DIRAC/commit/8d4217bc607c9d43c3ddce612fb2a28b20f8f48e

BEGINRELEASENOTES
*Framework
FIX: dirac-proxy-init printInfo without the dirac group 
ENDRELEASENOTES

PR #7452, opened 2024-02-08
[8.0] fix: dirac-proxy-init should not print the group anymore

aldbr:
Similar to https://github.com/DIRACGrid/DIRAC/pull/7451

BEGINRELEASENOTES
*Framework
FIX: dirac-proxy-init printInfo without the dirac group
ENDRELEASENOTES


aldbr:
I tested it both with a v8.0 (LHCbDIRAC) and a v9.0 server.

fstagni (approved):

PR #7470, opened 2024-02-14
[integration] fix: check running containers

arrabito:
BEGINRELEASENOTES

*tests
FIX: check running containers in integration tests script

ENDRELEASENOTES

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -593,7 +593,6 @@ def _check_containers_running(*, is_up=True):
             ["docker", "compose", "-f", docker_compose_fn, "ps", "-q", "-a"],
             stdout=subprocess.PIPE,
             env=_make_env({}),
-            check=True,
             text=True,
         ).stdout.split("\n")
     if is_up:
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -593,6 +593,8 @@ def _check_containers_running(*, is_up=True):
             ["docker", "compose", "-f", docker_compose_fn, "ps", "-q", "-a"],
             stdout=subprocess.PIPE,
             env=_make_env({}),
+            # docker compose ps has a non-zero exit code when no containers are running
+            check=False,
             text=True,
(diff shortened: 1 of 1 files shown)


arrabito:
At least on MacOS platform:

`./integration_tests.py`
fails if there are no running containers with the following error:

```
CalledProcessError: Command '['docker', 'compose', '-f', 
PosixPath('/var/folders/7j/qdb2nq852973tsy1_v99qzrm0000gq/T/tmp9hv6jxlc/ci/docker-compose.yml'), 'ps', '-q', '-a']' returned non-zero exit 
status 1.
```
Disabling the exit code check makes the script works on all platforms.

maxnoe:
I ran into the same issue on Ubuntu 22.04 with a fresh checkout.

integration_tests.py
```diff
@@ -593,7 +593,6 @@ def _check_containers_running(*, is_up=True):
             ["docker", "compose", "-f", docker_compose_fn, "ps", "-q", "-a"],
```
chrisburr, line 593: Can you try this instead?

```suggestion
            ["docker", "compose", "-f", str(docker_compose_fn), "ps", "-q", "-a"],
```

arrabito: I've tried but I get the same error.

maxnoe: The error is not a python error.

docker compose ps just has a non-zero exit code when no containers are running

integration_tests.py
```diff
@@ -593,7 +593,6 @@ def _check_containers_running(*, is_up=True):
...
             stdout=subprocess.PIPE,
             env=_make_env({}),
-            check=True,
             text=True,
```
chrisburr, line 598:
```suggestion
            # docker compose ps has a non-zero exit code when no containers are running
            check=False,
            text=True,
```

chrisburr (approved):

PR #7477, opened 2024-02-19
[8.0] VOMSRole noVOMS prevents adding VOMS extension to a proxy

atsareg:
In EGI there are multiple VOs that do not use grid storage. They need VOMS proxies only to access computing elements. So, only xxx_pilot groups need VOMS extension whereas xxx_user groups doesn't. This PR allows to configure this. As a result - considerably less load on multiple VOMS servers, faster responses from the ProxyManager.

BEGINRELEASENOTES

*Framework
CHANGE: ProxyDB - The VOMSRole in the group configuration set to noVOMS value prevents from adding VOMS extension

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/FrameworkSystem/DB/ProxyDB.py
+++ b/src/DIRAC/FrameworkSystem/DB/ProxyDB.py
@@ -929,17 +929,20 @@ class ProxyDB(DB):
                         "Stored proxy has already a different VOMS attribute %s than requested %s"
                         % (attrs[0], vomsAttr)
                     )
-            else:
+            elif vomsAttr != "noVOMS":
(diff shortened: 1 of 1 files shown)


chaen:
isn't `AutoAddVOMS=False` what you are looking for ? :thinking:

atsareg:
No, I am looking for not getting VOMS extension when executing requests. I got a situation with lots of ForwardDISET requests for a group which never accesses third party services. Actually, in the RequestTask always the VOMS proxy for the request owner is downloaded. It should be rather corrected there indeed to check first for the voms attribute as done for the shifter proxies.

PR #7478, opened 2024-02-19
[8.0] Fix DISET calls with proxy to be used passed as an argument

atsareg:
In case the DISET Client is used with proxyChain argument (this is the case of the payload proxy renewal in the JobAgent), the proxy string was not properly passed to the ssl transport call. This fixes #7450   

BEGINRELEASENOTES

*Core
FIX: Fix DISET calls with proxy to be used passed as an argument

ENDRELEASENOTES


fstagni (approved):
Seems correct.

PR #7489, opened 2024-02-28
[8.0] fix: do not print group info when doing dirac-proxy-init

chaen:
Same as https://github.com/DIRACGrid/DIRAC/pull/7452

BEGINRELEASENOTES
*Framework
FIX: dirac-proxy-info without the dirac group
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/FrameworkSystem/scripts/dirac_proxy_info.py
+++ b/src/DIRAC/FrameworkSystem/scripts/dirac_proxy_info.py
@@ -113,23 +113,21 @@ def main():
             if uploadedInfo:
                 gLogger.notice("== Proxies uploaded ==")
                 maxDNLen = 0
-                maxGroupLen = 0
+
                 for userDN in uploadedInfo:
                     maxDNLen = max(maxDNLen, len(userDN))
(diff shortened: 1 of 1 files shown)


src/DIRAC/FrameworkSystem/scripts/dirac_proxy_info.py
```diff
@@ -113,23 +113,21 @@ def invalidProxy(msg):
...
                 maxDNLen = 0
-                maxGroupLen = 0
+
                 for userDN in uploadedInfo:
                     maxDNLen = max(maxDNLen, len(userDN))
-                    for group in uploadedInfo[userDN]:
-                        maxGroupLen = max(maxGroupLen, len(group))
-                gLogger.notice(f" {'DN'.ljust(maxDNLen)} | {'Group'.ljust(maxGroupLen)} | Until (GMT)")
+                    # for group in uploadedInfo[userDN]:
+                    #     maxGroupLen = max(maxGroupLen, len(group))
```
aldbr, line 120:
```suggestion
```
?

PR #7496, opened 2024-03-01
[8.0] Monitoring type loader fixes for editable install

chaen:
I noticed when developing a custom monitoring type for LHCb that it was not picked up. 

When making an editable installation of `DIRAC`, the `loadObjects` function cannot work.
This is because it relies on DIRAC.__file__` which resolves to `<yourDIRACClone>/src/DIRAC/__init__.py` so this loop will not find the good directory for extensions

https://github.com/DIRACGrid/DIRAC/blob/16c620ed0446788e552067b439519c3de757c9a0/src/DIRAC/Core/Utilities/ObjectLoader.py#L183-L184

Instead of changing this for all the users (only a few, but who knows), I just changed the plotting loader to use the `ObjectLoader`, which does the correct thing

BEGINRELEASENOTES
*Core
FIX: plotting TypeLoader works with editable installation

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/ObjectLoader.py
+++ b/src/DIRAC/Core/Utilities/ObjectLoader.py
@@ -177,6 +177,11 @@ class ObjectLoader(metaclass=DIRACSingleton):
 
 def loadObjects(path, reFilter=None, parentClass=None):
     """
+
+    Note: this does not work for editable install because it hardcodes
+    DIRAC.__file__
+    It is better to use ObjectLoader().getObjects()
+
     :param str path: the path to the syetem for example: DIRAC/AccountingSystem
     :param object reFilter: regular expression used to found the class
     :param object parentClass: class instance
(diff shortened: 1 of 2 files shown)


src/DIRAC/Core/Utilities/ObjectLoader.py
```diff
@@ -177,6 +177,11 @@ def getObjects(self, modulePath, reFilter=None, parentClass=None, recurse=False,
 
 def loadObjects(path, reFilter=None, parentClass=None):
```
chrisburr, line 178: Add the deprecated decoarator? Or do we still need it?

chaen: It's still being used in a few place. I could probably find a proper fix, but that's for a major release

chrisburr (approved):

PR #7504, opened 2024-03-07
[8.0] fix update pilot status

arrabito:
BEGINRELEASENOTES
*Workload
FIX: update pilot status in case the pilot failed before being able to install DIRAC, the pilot status was not updated and got stuck to 'Submitted'.
ENDRELEASENOTES


arrabito:
This fix was provided by @aldbr and tested on the CTA instance.
The problem was that in case the pilot failed before being able to install DIRAC, the pilot status was not updated and got stuck to 'Submitted'.

aldbr (approved):

PR #7536, opened 2024-03-22
[8.0] feat: `RemoteRunner` can take a container image in parameter

aldbr:
In the LHCb context, applications have traditionally operated within a uniform container environment. This setup relied on specifying a singular container image via the `XRSLExtraString` parameter within the ARC/AREXCE configuration. This approach ensured simplicity and consistency across our computing tasks but also imposed limitations on flexibility for specific application requirements.

LHCb applications may now require tailored container environments to run properly. Applications are capable of autonomously selecting the most suitable container image from CVMFS before running. However, in environments lacking CVMFS support, applications would need to initiate a container within another container, a practice that may conflict with the policies enforced by system administrators I guess.

The goal of the PR is to integrate the possibility to dynamically select the appropriate container image best suited for each application's unique requirements.

BEGINRELEASENOTES
*WorkloadManagement
CHANGE: RemoteRunner can take a container image in parameters
*Resources
NEW: The ARC CE now has two new parameters, ContainerImage and ApptainerBinary, to configure which container image to run how
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/ARCComputingElement.py
+++ b/src/DIRAC/Resources/Computing/ARCComputingElement.py
@@ -27,6 +27,12 @@ WorkingDirectory:
 
      /opt/dirac/pro/runit/WorkloadManagement/SiteDirectorArc
 
+ContainerImage:
+    Container image to use for the job execution. Make sure the image is available on the worker node.
+
+ApptainerBinary:
+    Apptainer binary location. This is the binary that will be used to execute the container image.
+
 XRSLExtraString:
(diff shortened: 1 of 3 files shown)


chrisburr:
> However, in environments lacking CVMFS support, applications would need to initiate a container within another container, a practice that may conflict with the policies enforced by system administrators I guess.

I don't think this has anything to do with CVMFS not being available?

Regardless, I think this ia a good approach given the nature of these resources.

PR #7588, opened 2024-04-22
[v8.0] fix: AREX out and err need to exist before file integrity check

aldbr:
The Solution introduced in https://github.com/DIRACGrid/DIRAC/pull/7549 does not work: `RemoteRunner._checkOutputIntegrity()` checks the integrity of the output, but the `.out` and `.err` files are removed by the `AREXCE` before being checked. Therefore, the output is considered corrupted.

Here we remove the `.out` and `.err` in the `RemoteRunner` after the integrity check, following the `AREXCE` format. I don't think it is a problem because:
- the `RemoteRunner` works only with `AREX/ARC` CEs
- the `RemoteRunner` is going to disappear at some point.

BEGINRELEASENOTES
*WorkloadManagement
FIX: AREX "out" and "err" need to exist before file integrity check
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/AREXComputingElement.py
+++ b/src/DIRAC/Resources/Computing/AREXComputingElement.py
@@ -898,10 +898,8 @@ class AREXComputingElement(ARCComputingElement):
             if remoteOutput == f"{stamp}.out":
                 with open(localOutput) as f:
                     stdout = f.read()
-                os.unlink(localOutput)
             if remoteOutput == f"{stamp}.err":
                 with open(localOutput) as f:
                     stderr = f.read()
-                os.unlink(localOutput)
 
         return S_OK((stdout, stderr))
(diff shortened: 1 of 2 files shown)


src/DIRAC/Resources/Computing/AREXComputingElement.py
```diff
@@ -898,10 +898,8 @@ def getJobOutput(self, jobID, workingDirectory=None):
             if remoteOutput == f"{stamp}.out":
                 with open(localOutput) as f:
                     stdout = f.read()
-                os.unlink(localOutput)
```
andresailer, line 901: But it is possible that we are not using the RemoteRunner with AREX, right? So who cleans up these files then?

aldbr: Indeed, the `PilotManager` is also using `<CE>.getJobOutput()` and removes the `WorkingDirectory` (containing the outputs) at the end of the operation: https://github.com/DIRACGrid/DIRAC/blob/f877d14fb457e4da192fb5c90a82421f36678653/src/DIRAC/WorkloadManagementSystem/Service/PilotManagerHandler.py#L172

I agree that this is not ideal, but IIUC, at some point, pilots will report their outputs by themselves and I guess that the `getJobOutput` method would only be used for such a use case(?).

andresailer: OK, then there is no problem!

fstagni (approved):

Issue #7602, opened 2024-05-04
Prevent upload of files with names which are too long for the DFC


chrisburr:
We had a user upload files with extremely long LFNs that resulted in requests which couldn't be processed due to a MySQL issue:

```[01]
ID=217409110
LFN='/lhcb/user/x/xxxxxxx/2024_05/877947/877947554/13000GeV__Z__POWHEGPythia__8245ptmaxmatch2__NNPDF31_nlo_as_0118__0__as0.118000_IKT1.000000_s2tw0.233000_mz91.153500__evts20__seed2__lhcbacc.root'
Status='Waiting'
Checksum='91d143ba'
Error='failed to register LFN: {'FileCatalog': "MySQL Error ( 1131 : Execution failed.: ( 1406: Data too long for column 'FileName' at row 1 ))"}'
Attempts=70
```

This should probably be caught by the client before even trying to upload the file.

PR #7604, opened 2024-05-06
feat (DMS): DataManager.putAndRegister rejects too long filename

chaen:
closes https://github.com/DIRACGrid/DIRAC/issues/7602 (hopefully)

BEGINRELEASENOTES

*DMS
NEW: DataManager.putAndRegister rejects too long filename

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Client/DataManager.py
+++ b/src/DIRAC/DataManagementSystem/Client/DataManager.py
@@ -434,7 +434,7 @@ class DataManager:
         'overwrite' removes file from the file catalogue and SE before attempting upload
         """
 
-        if len(os.path.basename(lfn)) >= MAX_FILENAME_LENGTH:
+        if len(os.path.basename(lfn)) > MAX_FILENAME_LENGTH:
             return S_ERROR(errno.ENAMETOOLONG, f"maximum {MAX_FILENAME_LENGTH} characters allowed")
 
         res = self.__hasAccess("addFile", lfn)
--- a/src/DIRAC/DataManagementSystem/Client/DataManager.py
+++ b/src/DIRAC/DataManagementSystem/Client/DataManager.py
@@ -8,6 +8,7 @@
 This module consists of DataManager and related classes.
 
 """
+
 # # imports
 from datetime import datetime, timedelta
 import fnmatch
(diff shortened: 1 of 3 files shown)


src/DIRAC/DataManagementSystem/Client/FailoverTransfer.py
```diff
@@ -107,6 +109,9 @@ def transferAndRegisterFile(
                     break
                 elif cmpError(result, EFCERR):
                     self.log.debug("transferAndRegisterFile: FC unavailable, retry")
+                elif cmpError(result, errno.ENAMETOOLONG):
+                    self.log.debug("transferAndRegisterFile: this file won't be uploaded: {result}")
```
andresailer:
```suggestion
                    self.log.debug(f"transferAndRegisterFile: this file won't be uploaded: {result}")
```

chaen: thanks !!

src/DIRAC/DataManagementSystem/Client/__init__.py
```diff
@@ -1,3 +1,6 @@
 """
    DIRAC.DataManagementSystem.Client package
 """
+
+# This should be the same as the FileName column of the DFC
```
andresailer:
```suggestion
#: Maximum number of characters for a filename, this should be the same as the FileName column of the DFC
```

The `#:` to make the comment appear in the docs

src/DIRAC/DataManagementSystem/Client/DataManager.py
```diff
@@ -432,6 +434,9 @@ def putAndRegister(self, lfn, fileName, diracSE, guid=None, path=None, checksum=
         'overwrite' removes file from the file catalogue and SE before attempting upload
         """
 
+        if len(os.path.basename(lfn)) >= MAX_FILENAME_LENGTH:
```
andresailer:
```suggestion
        if len(os.path.basename(lfn)) > MAX_FILENAME_LENGTH:
```
We need `>` so we can use all the 128 characters, and not only 127, right?

(Sorry for dripping comments in like this.)

chaen: I actually did it on purpose because I seem to remember being bitten by something like it and the field being truncated, but I really can't recall what it was :laughing:

andresailer: Adding a comment, and changing the message to read 127 max characters is also fine.

chaen: I went with the approach of `>` and watch the world burn :laughing: 
Since my memory is not so good, maybe I am mistaken, or it will be a reminder of why I had that n my mind :-p

chaen:
@andresailer happy like that ? :-)

andresailer (approved):
Yes, thanks!

PR #7614, opened 2024-05-14
[8.0] fix for pylint 3.2.0 

chaen:
BEGINRELEASENOTES
*All
FIX: fix pylint 3.2.0 warnings

ENDRELEASENOTES


src/DIRAC/Core/Base/AgentModule.py
```diff
@@ -329,6 +329,7 @@ def _setShifterProxy(self):
 
     def am_go(self):
         # Set the shifter proxy if required
+        initialWallTime = initialCPUTime = cpuPercentage = -1
```
chrisburr: I'm not sure I like adding this for the sake of a linter...

I think I'd prefer refactoring it like this:

```diff
         if self.activityMonitoring:
             initialWallTime, initialCPUTime, mem = self._startReportToMonitoring()
-        cycleResult = self.__executeModuleCycle()
-        if self.activityMonitoring and initialWallTime and initialCPUTime:
+            cycleResult = self.__executeModuleCycle()
             cpuPercentage = self._endReportToMonitoring(initialWallTime, initialCPUTime)
+.       else:
+            cycleResult = self.__executeModuleCycle()
```

src/DIRAC/ConfigurationSystem/scripts/dirac_admin_check_config_options.py
```diff
@@ -174,6 +174,8 @@ def _printDiff(self, entry, level=""):
             diffType, entryName, _value, changes, _comment = entry
         elif len(entry) == 4:
             diffType, entryName, _value, changes = entry
+        else:
+            raise ValueError(f"Invalid entry {entry}")
```
chrisburr, line 178:
```suggestion
            raise NotImplementedError(f"Invalid entry {entry}")
```

src/DIRAC/Core/Security/VOMS.py
```diff
@@ -74,6 +75,7 @@ def getVOMSAttributes(self, proxy, switch="all"):
                     if value and value not in attributes and value in validVOMSAttrs:
                         attributes.append(value)
 
+        returnValue = None
```
chrisburr: Why not add `raise NotImplementedError(switch)`?

src/DIRAC/Core/Base/AgentModule.py
chaen: Works with me. I am surprise that pylint does not have a false positive on that one though :-)

src/DIRAC/ConfigurationSystem/scripts/dirac_admin_check_config_options.py
chaen: The definition of `ValueError` seems more fitting:
> Raised when an operation or function receives an argument that has the right type but an inappropriate value

src/DIRAC/Core/Security/VOMS.py
chaen: I am happy with that too. This argument is actually never used :shrug:

fstagni (approved):

PR #7653, opened 2024-06-06
[8.0] Remove dead code path in ReplicateAndRegister

chaen:
@chrisburr @andresailer I need a good review of that one :-) (not the RMS, that's just dead code) 
Oyr FTS agents had half a billion of proxy files left... not with this anymore I hope :)

BEGINRELEASENOTES

*Core
CHANGE: register the DictCache destructor as an atexit handler

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
+++ b/src/DIRAC/DataManagementSystem/Agent/RequestOperations/ReplicateAndRegister.py
@@ -605,13 +605,6 @@ class ReplicateAndRegister(DMSRequestOperationsBase):
                     # All source SEs are banned, delay execution by 1 hour
                     delayExecution = 60
                 continue
-            # # get the first one in the list
-            if sourceSE not in validReplicas:
-                if sourceSE:
-                    err = "File not at specified source"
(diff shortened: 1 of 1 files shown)


chrisburr:
Isn't the issue here that the FTSAgent isn't passing a `deleteFunction` so there is nothing to delete the files?

chaen:
> Isn't the issue here that the FTSAgent isn't passing a `deleteFunction` so there is nothing to delete the files?

I don't think it is, as the `ProxyManagerClient` which the `FTS3Agent` uses has the proper delete function https://github.com/DIRACGrid/DIRAC/blob/877f334cf10eff17f046e757e2df82c158de12f4/src/DIRAC/FrameworkSystem/Client/ProxyManagerClient.py#L33

chrisburr:
Delightful! A global singleton and a reference cycle between `DictCache` and `ProxyManagerClient`. It's impossible for this object to ever get garbage collected.

I think the right way to fix this is: https://github.com/DIRACGrid/DIRAC/pull/7666

chrisburr:
I guess this should be rebased and renamed to just include the dead code removal.

fstagni (approved):

PR #7666, opened 2024-06-10
[8.0] Replace __del__ with weakref.finalize in DictCache

chrisburr:
See https://github.com/DIRACGrid/DIRAC/pull/7653 and https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods

BEGINRELEASENOTES

*Core
FIX: Replace `__del__` with weakref.finalize in DictCache

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Utilities/DictCache.py
+++ b/src/DIRAC/Core/Utilities/DictCache.py
@@ -3,6 +3,7 @@
 """
 import datetime
 import threading
+import weakref
 
 # DIRAC
 from DIRAC.Core.Utilities.LockRing import LockRing
@@ -63,6 +64,8 @@ class DictCache:
 
         # Function to clean the elements
         self.__deleteFunction = deleteFunction
+
+        self.__finalizer = weakref.finalize(self, self.purgeAll, useLock=False)
 
     @property
     def lock(self):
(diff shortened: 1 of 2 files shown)


src/DIRAC/Core/Utilities/DictCache.py
```diff
@@ -64,6 +65,8 @@ def __init__(self, deleteFunction=False, threadLocal=False):
         # Function to clean the elements
         self.__deleteFunction = deleteFunction
 
+        self.__finalizer = weakref.finalize(self, self.purgeAll, useLock=False)
```
chrisburr: The use of `purgeAll` here should probably be refactored to not depend on self and go pass in the members it needs instead.

chaen:
Ah cool, I did not know about `weakref`, and apparently, it's the right thing to do :laughing: https://github.com/cupy/cupy/pull/8315

src/DIRAC/Core/Utilities/DictCache.py
chaen: Fine by me, do you want to add it in this PR ?

chrisburr: Done

chaen (approved):

PR #7683, opened 2024-06-21
[8.0] Transformation InputDataAgent with MultiVOMetadata

### LINKED ISSUE #7681: InputDataAgent with MultiVo*Metadata
Because the Transformation InputDataAgent is not using the credentials of the transformation author, but just the host information, we would need to be use author credentials when making the FileCatalog client. Otherwise the MultiVo*Metadata is not being picked up.

cc. @marianne013 

https://github.com/DIRACGrid/DIRAC/blob/4886513ac7a42f01c06cd5b2339fa666d3a04115/src/DIRAC/TransformationSystem/Agent/InputDataAgent.py#L119

andresailer:
This is needed for MultiVO*Metadata file catalog configurations

BEGINRELEASENOTES

*TS
NEW: InputDataAgent: new Option MultiVO, which makes the FileCatalog Query use the author of the DN, rather than the Host, to resolve MultiVO metadata correctly. Fixes #7681 

ENDRELEASENOTES

@marianne013 Please test? But mind the other issue with dirac-dms-find-lfns and metadata that might cause way too many files to be found?

### DIFF
--- a/src/DIRAC/TransformationSystem/Agent/InputDataAgent.py
+++ b/src/DIRAC/TransformationSystem/Agent/InputDataAgent.py
@@ -46,6 +46,7 @@ class InputDataAgent(AgentModule):
         self.transClient = TransformationClient()
         self.metadataClient = FileCatalogClient()
         self.transformationTypes = None
+        self.multiVO = False
 
     #############################################################################
     def initialize(self):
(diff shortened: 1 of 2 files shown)


marianne013:
@andresailer Can I just confirm that this is different to the fix we tried out on Friday ?
FYI: @sfayer @martynia

andresailer:
No, this _is_ what we tried on Friday, isn't it?

andresailer:
I am also sufficiently confident that there is nothing in principle wrong in this patch. The issue to solve is #7687 but the changes for that should not impact what we need to do here.

marianne013:
OK, I think I need to wait for @martynia to have a go at fixing https://github.com/DIRACGrid/DIRAC/issues/7687 first.

fstagni (approved):

PR #7702, opened 2024-06-27
[8.0] fix(HTCondorCE): UseSSLSubmission: we can just use the proxy file for everything

andresailer:
Hi @atsareg  @aldbr 

This is what you are doing, right?

BEGINRELEASENOTES

*Resources:
CHANGE:  HTCondorCE: UseSSLSubmission: use the generated proxy file for everything, no longer need to have certificate of user present on the server.

ENDRELEASENOTES


aldbr:
I think this corresponds to @atsareg's suggestion, but I will let him review the PR

atsareg (approved):

PR #7703, opened 2024-06-27
[9.0] feat(Resources): introduce fabric in SSHCE

### LINKED ISSUE #7026: Replace the `SSH` class by a Python library?
As explained in https://github.com/DIRACGrid/DIRAC/pull/7025, the ` _ssh_call()` method does not seem to work perfectly.
There exist a few popular python libraries to perform SSH operations like `Fabric` (Paramiko) (https://docs.fabfile.org/en/stable/).

We could investigate the possibility of using one of them instead of relying on our `SSH` class to perform remote operations over SSH.

aldbr:
Replace the Dirac-specific `SSH` class by `fabric`.

BEGINRELEASENOTES
*Resources
CHANGE: Replace SSH by fabric in SSHComputingElement
ENDRELEASENOTES

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,11 +18,14 @@ dependencies:
   - cwltool
   - db12
   - opensearch-py
+  - fabric
   - fts3
   - gitpython >=2.1.0
+  - invoke
   - m2crypto >=0.38.0
   - matplotlib
   - numpy
+  - paramiko
   - pexpect >=4.0.1
   - pillow
   - prompt-toolkit >=3,<4
--- a/src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
+++ b/src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
@@ -1,20 +1,19 @@
 #!/usr/bin/env python
 """
-Test the interactions with a given set of Computing Elements (CE). For each CE:
+Test the interactions with a given set of Computing Elements (CE). For each CE
 
 - Get the CE status if available
 - Submit a job to the CE
 - Get the job status
 - Get the job output/error/log if available
 
-Conditions:
+Conditions
 
 - The CEs must be configured in the DIRAC configuration
-- The script should be executed with an admin proxy: used to fetch a pilot proxy and a token
-- The script should be executed:
-
-  - in a DIRAC client environment for normal CEs, such as AREX and HTCondor
-  - in a DIRAC host environment for SSH/Local CEs (credentials would not be available otherwise)
+- For non-SSH CEs (e.g. AREX, HTCondor): requires an admin proxy (FullDelegation)
+  to fetch pilot proxy and token. Run from a DIRAC client environment.
+- For SSH/SSHBatch CEs: no proxy required. These use SSH credentials from the CS
+  configuration. Run from a DIRAC server/host environment.
 
 Usage:
   dirac-admin-debug-ce <VO> [--site <site>] [--ce <ce>] [--ce-type <type>] [--script <path>]
@@ -23,6 +22,7 @@ Example:
   $ dirac-admin-debug-ce dteam --site LCG.CERN.cern --ce-type HTCondorCE
 """
 import concurrent.futures
+import tempfile
 import time
 from pathlib import Path
 
@@ -153,10 +153,11 @@ def buildQueues(vo, sites, ces, ceTypes):
     return result["Value"]
 
 
-def interactWithCE(ce):
+def interactWithCE(ce, executableFile):
     """Interact with a given Computing Element (CE).
 
     :param ce: The Computing Element instance.
+    :param str executableFile: The path to the executable script to submit.
     :return: A dictionary with the result of each check.
     """
     checks = {
@@ -178,7 +179,7 @@ def interactWithCE(ce):
 
     # Submit a job to the CE
     gLogger.info(f"[{ce.ceName}]Submitting a job")
-    res = ce.submitJob("workloadExec.sh", None)
+    res = ce.submitJob(executableFile, None)
     if not res["OK"]:
         gLogger.error(f"[{ce.ceName}]Cannot submit job to CE: {res['Message']}")
         checks["job_submit"]["Message"] = res["Message"]
@@ -236,30 +237,18 @@ def main():
     Script.registerSwitch("", "script=", "Path to custom executable script (default: workloadExec.sh)", setScript)
     Script.registerSwitch("", "timeout=", "Timeout in seconds for job status polling (default: 1800)", setTimeout)
     Script.registerArgument("VO: Virtual Organization")
+
+    # Determine the credential context before parseCommandLine contacts the CS:
+    # if no proxy file exists, assume we are on a server with a host certificate.
+    from DIRAC.Core.Security.Locations import getProxyLocation
+
+    hasProxy = getProxyLocation() is not None
+    if not hasProxy:
+        Script.localCfg.addDefaultEntry("/DIRAC/Security/UseServerCertificate", "true")
+
     Script.parseCommandLine()
 
-    from DIRAC.Core.Security.Properties import SecurityProperty
-    from DIRAC.Core.Security.ProxyInfo import getProxyInfo
-
-    # Check credentials
-    result = getProxyInfo()
-    if not result["OK"]:
-        gLogger.error("Do you have a valid proxy?")
-        gLogger.error(result["Message"])
-        DIRAC.exit(1)
-    proxyProps = result["Value"]
-
-    if SecurityProperty.FULL_DELEGATION not in proxyProps.get("groupProperties", []):
-        gLogger.error("You need an admin proxy (with FullDelegation property) to run this script")
-        DIRAC.exit(1)
-
     vo = Script.getPositionalArgs()[0]
-
-    # Get credentials for the given VO
-    pilotDN, pilotGroup = findGenericCreds(vo)
-    if not pilotDN or not pilotGroup:
-        gLogger.error("Cannot get pilot credentials")
-        DIRAC.exit(1)
 
     # Get the queues
     queueDict = buildQueues(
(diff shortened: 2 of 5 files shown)


environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
   - elasticsearch <7.14
   - elasticsearch-dsl
   - opensearch-py
+  - fabric
```
chrisburr, line 21: invoke and paramiko should be added here as well if they're used directly. Also they should be added the the setup.cfg.

fstagni:
Just a note: we do not (yet) have a way to do proper integration test for the Computing Elements, but one may think about adding them to our integration tests setup. Something to think about it, it would be nice if it was in this PR. 
It involves creating the "site", with the "CE" (this would be yet another container) and the SiteDirector could send pilots to it.

aldbr:
> Just a note: we do not (yet) have a way to do proper integration test for the Computing Elements, but one may think about adding them to our integration tests setup. Something to think about it, it would be nice if it was in this PR. It involves creating the "site", with the "CE" (this would be yet another container) and the SiteDirector could send pilots to it.

I agree it would great to add integration tests for CEs, at least to test basic features. But it will likely become complex because:
- if we want to test things properly, we need to set up a CE and a Batch System.
- we will have to choose one configuration, but it might not reflect the configuration of the sites in production.

I will give it a try with the `SSHCE`, let's see.

aldbr:
I wonder if it really makes sense to add CEs (and Batch Systems) in the integration tests: while it would be great to have a "grid in a box" in a controller environment, it would be cumbersome to maintain on the long term and would not be representative of all the instances we can find out there (e.g. Arc v6, v6 with a hack, v7, transferring jobs to Slurm, HTCondor, SSH, SSH tunnel, HTCondor with local scheduler, with remote scheduler...).

It would probably make more sense to add some scripts to run during the hackathons. For each type of CE supported it would:
- get all the instances related to the given type of CE and for each of them:
  - submit a "hello world" job
  - get the CE status
  - get the job status until it reaches a final state
  - get the job output and logging info (if available)

Basically, it would be very similar to (i) submitting pilots with the Site Director and (ii) checking their results manually. But it would be more focused on the CE interfaces and would be more automated (though a human would need to check whether errors come from the CE instance itself or the Dirac CE interface).

Any opinion @fstagni ?

fstagni:
I think the only one that would make sense to set up here is the SSHCE. The others, "proper Grid ones", can not be tested here.

aldbr:
I don't even know if testing SSHCE in an integration test makes sense. The only easy test we can set up would be `SSHCE` + `Host`, which is not representative of what we can have in production.

fstagni:
OK OK, give up on the idea...

aldbr:
I will add a certification test focused on the CE interfaces as I explained (+ a card in the kanban board to explain how to execute it). I will execute it in the lhcb environment to make sure the changes in this PR are correct. 

And I can also try to add a container that would act as a "Site" and use `SSH` + `Host` so that we can at least test the Site Director "in a box". Would it be okay?

fstagni:
Sure, thanks

aldbr:
Tested with https://github.com/DIRACGrid/DIRAC/pull/8420 in LHCb production + now used by the LHCb Site Directors for a few hours without any issue so far.

This PR does not modify the interfaces so the transition is expected to be transparent.
The only risky point I identified is the `SSHTunnel`, which can take any command as value: I need to extract the hostname as well as the port from it.
My function is supposed to be flexible and support various cases, but I can't guarantee that this is going to work fine if the value is tricky.

aldbr:
Another risky point is the `SSHBatchCE` because I don't know how to test that properly, we don't have any instance in LHCb as far as I know.

fstagni:
Does this need a new DIRACOS release?

fstagni:
A few comments after asking the relevant admins:
- the `SSHTunnel` option is not used at the moment by any other user but LHCb. It is anyway felt that it can be useful (and looks like it is!)
- SSHBatch CE is there for testing purposes, can be some times used for short-lived "CEs". There is a SSHBatch CE defined in the certification machine.

This PR should be tested in the certification setup.

aldbr:
Alright, then I will simplify the `SSHTunnel` option.
For the `SSHBatch` CE: I still fail to understand its goal. What would you test with that exactly? Is it really needed? I see the `SSHBatch` CE in certification, it looks like it is serving only 1 host and could be just an `SSHCE`. Am I wrong?

fstagni:
The one in certification is just an example, of course. More than a single IP can be added.

SSHBatch can be used for creating a grid in a box, and in this case it's just for local testing, and/or demonstration possibilities. 
My impression is that its functionalities could be included in SSHCE (if `BatchSystem == "Host"` and an `SSHHost` list is defined).

aldbr:
Given that:
- there is no real use case in production (IIUC)
- the "grid in a box" use case is hypothetical at this point (again IIUC), and we could just use an `SSHCE` per host
  - btw, if we ever have a "grid in a box", then it would be likely better to use the `SSHCE` to test it (as we said in earlier comments)
- other CEs work the same way: 1 CE = 1 endpoint

Shouldn't we just drop that feature?

arrabito:
For me dropping SSHBatch is fine. 

Then about:

> the SSHTunnel option is not used at the moment by any other user but LHCb. It is anyway felt that it can be useful (and looks like it is!)

As I said in my answer to the admin list, we do use the SSHTunnel option combined with SSHCE.
To be honest I've just configured my SSHCE with it, as I'm not sure about which are the alternatives.

To clarify the context of our usage, we have 2 use cases:

1/ in production for a site which does not have an HTCondor, neither a ARC CE (even if very soon the site will adopt HTCondor, so that we will not need to use SSHCE anymore)

2/ in our certification instance and also in our kind deployment used in the CI -> 'grid in a box'

To avoid any misunderstanding I share here our configurations.

In production the Site using the SSHCE is configured as follows:

Capture d’écran 2026-03-11 à 08 21 46

Is my understanding correct that in this case we are using the SSHTunnel to execute slurm submission commands?

While for the certfication instance, the Site is a Host and is configured as follows:

Capture d’écran 2026-03-11 à 08 24 04

Again, my understanding is that we are also using the SSHTunnel in this case. Is that correct?

aldbr:
> Is my understanding correct that in this case we are using the SSHTunnel to execute slurm submission commands?

From what I can see, you are not using the `SSHTunnel` option here (I don't see it in the screenshot you provided).
`SSHTunnel` is used when you are passing through one (or multiple) jump node(s) to access the target node where the batch system commands are available.
Here from what I understand, you just SSH to a target node where slurm commands are available.
Am I understanding correctly?

arrabito:
> From what I can see, you are not using the SSHTunnel option here (I don't see it in the screenshot you provided).
SSHTunnel is used when you are passing through one (or multiple) jump node(s) to access the target node where the batch system commands are available.
Here from what I understand, you just SSH to a target node where slurm commands are available.
Am I understanding correctly?

Yes that's correct. So in this case we don't use the SSHTunnel option. Thank you for the clarification.

aldbr:
I have been testing the `SSHCE` in LHCb production for a few months now
I have just tested the `SSHBatchCE` in DIRAC certification, that allowed me to solve one issue, now everything looks fine.

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
```diff
@@ -236,31 +237,19 @@ def main():
     Script.registerSwitch("", "script=", "Path to custom executable script (default: workloadExec.sh)", setScript)
     Script.registerSwitch("", "timeout=", "Timeout in seconds for job status polling (default: 1800)", setTimeout)
     Script.registerArgument("VO: Virtual Organization")
-    Script.parseCommandLine()
 
-    from DIRAC.Core.Security.Properties import SecurityProperty
-    from DIRAC.Core.Security.ProxyInfo import getProxyInfo
+    # Determine the credential context before parseCommandLine contacts the CS:
+    # if no proxy file exists, assume we are on a server with a host certificate.
```
fstagni, line 242: From where do you want to run this? In general these commands are not to be run on servers.

aldbr: Indeed, but if you want to debug an SSHCE, you don't have any other choice because credentials are set on the server side

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
```diff
@@ -272,6 +261,37 @@ def main():
         gLogger.error("Cannot get queues")
         DIRAC.exit(1)
 
+    # SSH-based CEs run from the server and use SSH credentials from the CS,
+    # so they do not need a proxy or pilot credentials.
+    sshCETypes = {"SSH", "SSHBatch"}
+    hasNonSSHCEs = any(q["CE"].ceType not in sshCETypes for q in queueDict.values())
```
fstagni: I do not know if we will ever add more "SSH" CEs, but we can safely assume they will start with "SSH", so maybe the check can be


```suggestion
    hasNonSSHCEs = any(not q["CE"].ceType.startwith("SSH") for q in queueDict.values())
```
(the above is probably wrong syntax but you get my point)

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
fstagni, line 242: When would this be called by a host?

aldbr: That would never be called directly by a host. That would be called by a person needing to debug a given SSH CE.

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
```diff
@@ -10,11 +10,10 @@
 Conditions:
```
andresailer:
```suggestion
Conditions
```

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
andresailer:
```suggestion
Test the interactions with a given set of Computing Elements (CE). For each CE
```

andresailer (dismissed):
I propose to drop the colons before the bullet list, the diracdocutils add a second colon to make the following paragraph a verbatim environment, but that is not what you want. And that verbatim environments leads to an indendation issue.

src/DIRAC/WorkloadManagementSystem/scripts/dirac_admin_debug_ce.py
aldbr: Thanks!

fstagni (approved):

PR #7708, opened 2024-06-27
[8.0] FIX  Multi VO metadata double VO suffix issue fix (#7697)

### LINKED ISSUE #7687: MultiVOMetadata doubling VO identifier
It happens that the metadata key has twice appended (for example in the Transformation.InputDataAgent)

https://github.com/DIRACGrid/DIRAC/blob/4886513ac7a42f01c06cd5b2339fa666d3a04115/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/DirectoryMetadata/MultiVODirectoryMetadata.py#L23

Should
1. Prevent the creation of meta data keys with `___` inside
2. Only append `___<vo>` if it hasn't been appended yet

cc. @marianne013 @martynia

martynia:
BEGINRELEASENOTES

*DMS
FIX: Remove `def findDirIDsByMetadata(self, metaDict, dPath, credDict):` method from `MultiVODirectoryMetadata` (derived) class which caused an extra VO suffix added when searching. The method is meant to be used _internally_ only on keys which are already expanded in a MultiVO case. Add a user-level def `findDirectoriesByMetadata(self, queryDict, path, credDict)` to the derived class thus adding a VO suffix for a directory search. Fixes  #7687.

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/MultiVOFileMetadata.py
+++ b/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/MultiVOFileMetadata.py
@@ -120,7 +120,7 @@ class MultiVOFileMetadata(FileMetadata):
     def findFilesByMetadata(self, metaDict, path, credDict):
         """Find Files satisfying the given metadata
 
-        :param dict metaDict: dictionary with the metaquery parameters
+        :param dict metaDict: dictionary with the metaquery parameters                                                                                                                                                                                   :param dict metaDict: dictionary with the metaquery parameters
         :param str path: Path to search into
         :param dict credDict: Dictionary with the user credentials
 
--- a/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/MultiVOFileMetadata.py
+++ b/src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/MultiVOFileMetadata.py
@@ -120,7 +120,7 @@ class MultiVOFileMetadata(FileMetadata):
     def findFilesByMetadata(self, metaDict, path, credDict):
         """Find Files satisfying the given metadata
 
-        :param dict metaDict: dictionary with the metaquery parameters                                                                                                                                                                                   :param dict metaDict: dictionary with the metaquery parameters
+        :param dict metaDict: dictionary with the metaquery parameters
         :param str path: Path to search into
         :param dict credDict: Dictionary with the user credentials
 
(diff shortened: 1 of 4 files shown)


src/DIRAC/DataManagementSystem/DB/FileCatalogComponents/FileMetadata/MultiVOFileMetadata.py
```diff
@@ -120,7 +120,7 @@ def getFileUserMetadata(self, path, credDict):
     def findFilesByMetadata(self, metaDict, path, credDict):
         """Find Files satisfying the given metadata
 
-        :param dict metaDict: dictionary with the metaquery parameters
+        :param dict metaDict: dictionary with the metaquery parameters                                                                                                                                                                                   :param dict metaDict: dictionary with the metaquery parameters
```
andresailer:
```suggestion
        :param dict metaDict: dictionary with the metaquery parameters
```

marianne013:
I'll try to test this. If I haven't by Monday afternoon, please ping me again.

marianne013:
Simon had the following comment offline, here reproduced for completeness, so I can find it when I have time to get back to this.

[start quote]
This boils down to:

findDirIDsByMetadata is effectively an internal function and should expect
to given a pre-processed meta key name: It's only called by one other
function which has already processed meta name (which is why the double VO
sneaks in). In vanilla DIRAC at least.

I'm not sure the best way to fix that, obviously just removing the
_getMetaNameDict from it works, but if there is some notion of internal and
external functions this should be clear in the function naming or
something. Maybe it's a simple as it should be _findDirIDsByMetadata and
all _ functions shouldn't process the metadata keys?

There may be others that effectively have the same problem?
[end quote]

martynia:
The fix makes `findDirIDsByMetadata` an internal utility function (removed from the derived class) which does not add or manipulate a VO suffix in any way.  In order to maintain a correct directory search `findDirectoriesByMetadata(self, queryDict, path, credDict)` was added to the derived class which overloads the base class method and adds the suffix directly before calling the base class method. This follows a pattern  already present for files.

marianne013:
I just tested this on dirac00. It now says I am 60% done, which is progress. Now I just need to remember how many files I actually tagged and check that the files are where I think they should be.

marianne013:
Also, the changes I see in InputDataAgent.py are actually in https://github.com/DIRACGrid/DIRAC/pull/7683
Does this need re-basing ? Otherwise we can probably go ahead. One issue at a time.

andresailer (approved):

Issue #7487, opened 2024-02-28
dirac-dms-find-lfns misbehaves if given an invalid path


sfayer:
Hi,

I've noticed that dirac-dms-find-lfns returns entries even if it's given a Path= filter that points to a directory that doesn't exist. It seems to walk up the tree until it finds a path that does exist and then return all the files under that instead.

```
$ dirac-version 
8.0.38
$ dirac-dms-find-lfns Path=/gridpp/user/u/user.name/myjob/
/gridpp/user/u/user.name/myjob/job.log
# I'd expect this to say ".../badpath" doesn't exist
$ dirac-dms-find-lfns Path=/gridpp/user/u/user.name/myjob/badpath
/gridpp/user/u/user.name/myjob/job.log
$ dirac-dms-find-lfns Path=/gridpp/user/u/user.name/myjob/badpath/reallybadpath
/gridpp/user/u/user.name/myjob/job.log
```
(Server catalog backend is normal DIRAC FileCatalog)

I'd expect this to stop with an error rather than return files that don't match the filter! We use this command in a script to handle a specific type of deletion task and as you might imagine, this has resulted in files being removed from the catalogue that shouldn't have been.

Would it be possible for this to be fixed?

Regards,
Simon

sfayer:
I think this was all fixed by the linked PRs.

PR #7779, opened 2024-09-10
[v8r0] fix: dirac_dms_find_lfns: check if requested path exists and return error if not

marianne013:
BEGINRELEASENOTES

*DataManagement
FIX: dirac_dms_find_lfns: Check if requested path exists and returns error if not. Prevents users hitting #7487

ENDRELEASENOTES


fstagni (dismissed):
Just replace `'` with `"` (e.g. `["Value"]`) to make `black` happy, otherwise OK.

PR #7784, opened 2024-09-11
[8.0] IAM2CS: ForceNickname, fix duplicate accounts for users

andresailer:
BEGINRELEASENOTES

*ConfigurationSystem
CHANGE: VOMS2CSAgent: if a nickname is set, this nickname will always be used and no new accounts are going to be created if a DN changes or a user is in multiple VOs
NEW: VOMS2CSAgent: New option "ForceNickname", if this option is enabled no dirac user is created if no nickname attribute is set for a user
CHANGE: IAMService: use logger and return errors for users so that the VOMS2CSAgent can notify admins about issues

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/ConfigurationSystem/Client/VOMS2CSSynchronizer.py
+++ b/src/DIRAC/ConfigurationSystem/Client/VOMS2CSSynchronizer.py
@@ -342,23 +342,26 @@ class VOMS2CSSynchronizer:
 
                 # We have a real new user
                 if not diracName:
-                    if nickName:
-                        newDiracName = nickName.strip()
-                    else:
-                        newDiracName = self.getUserName(dn)
-
                     # Do not consider users with Suspended status in VOMS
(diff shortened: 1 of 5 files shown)


src/DIRAC/ConfigurationSystem/Client/VOMS2CSSynchronizer.py
```diff
@@ -257,7 +259,8 @@ def syncCSWithVOMS(self):
         if not result["OK"]:
             self.log.error("Could not retrieve user information", result["Message"])
             return result
-
+        if getUserErrors := result.get("Errors", []):
```
andresailer: This is currently not working, because of the way return values are done from the `getUsers` function.

Is this something people would be interested to get send via VOMS2CSAgent at all?
I.e. some missing entry like users that didn't link their certificate could come from this.

andresailer: Now the value contains both Users and Errors

fstagni (approved):

PR #7812, opened 2024-10-02
[v8r0] Improve getTransformationFiles performance

chrisburr:
This significantly improves the performance of `getTransformationFiles` by:

* Using a `JOIN` instead of manually looking up the LFNs from file IDs
* Remove the batching that was needed due to the use of `__getFileIDsForLfns`
    * This is doubly helpful as the use of `OFFSET N LIMIT 10000` made the function O(N^2) due to the database having to rebuild and scan the results for every batch.
* Giving the option to only get the columns you need

For example in LHCb: this optimises a ~1200 second call to be ~20 seconds.

BEGINRELEASENOTES

*TransformationSystem
NEW: 
CHANGE: Improve getTransformationFiles performance

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/TransformationSystem/Client/TransformationClient.py
+++ b/src/DIRAC/TransformationSystem/Client/TransformationClient.py
@@ -179,12 +179,14 @@ class TransformationClient(Client):
             res = rpcClient.getTransformationFiles(
                 condDict, older, newer, timeStamp, orderAttribute, offset, maxfiles, columns
             )
+            # TransformationDB.getTransformationFiles includes a "Records"/"ParameterNames"
+            # that we don't want to return to the client so explicitly return S_OK with the value
             if not res["OK"]:
                 return res
             return S_OK(res["Value"])
 
-        # If LFNs requested continue to do the old behavior of requesting in small batches
-        # Probably not needed?
+        # If LFNs requested, request in small batches, because...
+        # Probably not needed? Because this should always be a list
         if isinstance(condDict["LFN"], str):
             lfnList = [condDict["LFN"]]
         else:
--- a/src/DIRAC/TransformationSystem/Client/TransformationClient.py
+++ b/src/DIRAC/TransformationSystem/Client/TransformationClient.py
@@ -163,6 +163,7 @@ class TransformationClient(Client):
         timeout=1800,
         offset=0,
         maxfiles=None,
+        columns=None,
     ):
         """gets all the transformation files for a transformation, incrementally.
         "limit" here is just used to determine the offset.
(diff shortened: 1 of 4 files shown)


src/DIRAC/TransformationSystem/Client/TransformationClient.py
```diff
@@ -173,34 +174,37 @@ def getTransformationFiles(
...
+
+        if "LFN" not in condDict:
+            res = rpcClient.getTransformationFiles(
+                condDict, older, newer, timeStamp, orderAttribute, offset, maxfiles, columns
+            )
+            if not res["OK"]:
+                return res
+            return S_OK(res["Value"])
+
+        # If LFNs requested continue to do the old behavior of requesting in small batches
```
andresailer:
```suggestion
        # If LFNs requested, request in small batches, because...
```
Refering to "old behaviour" is going to be confusing soon

src/DIRAC/TransformationSystem/Client/TransformationClient.py
```diff
@@ -173,34 +174,37 @@ def getTransformationFiles(
...
+        # Probably not needed?
```
andresailer:
```suggestion
        # Probably not needed? Because this should always be a list
```
?

src/DIRAC/TransformationSystem/Client/TransformationClient.py
```diff
@@ -173,34 +174,37 @@ def getTransformationFiles(
...
-            # If a list of LFNs is given, use chunks of 1000 only
-            limit = limit if limit else 1000
+
+        if "LFN" not in condDict:
+            res = rpcClient.getTransformationFiles(
+                condDict, older, newer, timeStamp, orderAttribute, offset, maxfiles, columns
+            )
+            if not res["OK"]:
+                return res
+            return S_OK(res["Value"])
```
andresailer, line 186:
```suggestion
            return res
```
Or is there something from res we want to filter out when `OK`? (Which would then deserve a comment)

fstagni (dismissed):

src/DIRAC/TransformationSystem/Client/TransformationClient.py
```diff
@@ -179,12 +179,10 @@ def getTransformationFiles(
             res = rpcClient.getTransformationFiles(
                 condDict, older, newer, timeStamp, orderAttribute, offset, maxfiles, columns
             )
-            if not res["OK"]:
-                return res
-            return S_OK(res["Value"])
+            return res
```
chrisburr: This is wrong!

chrisburr:
I've rolled back incorrect commit and added the comment as suggested in https://github.com/DIRACGrid/DIRAC/pull/7812#discussion_r1786417196 so this should be good to go

chaen (approved):

PR #7823, opened 2024-10-10
[8.0] Drop CloudCE proxy handling

sfayer:
Hi,

Now that the pilot wrapper always includes a bundled proxy, we don't need to handle writing one specially in the cloud scripts, so this patch drops all of that functionality. This also fixes a recent v8 bug where CloudCE keeps self.proxy in a different format (string rather than proxy object) to other CEs resulting in a SiteDirector crash.

[Note: Users with a custom cloudinit.template will need to update it to match these changes].

Regards,
Simon

BEGINRELEASENOTES
*Resources
FIX: Drop CloudCE proxy handling
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Computing/CloudComputingElement.py
+++ b/src/DIRAC/Resources/Computing/CloudComputingElement.py
@@ -9,23 +9,20 @@ jobs. There are however some things that work differently:
 - File I/O: A small amount of input may be transferred through the
   instance metadata, but after that the VM is inaccessible.
 - Authentication: Most cloud endpoints use a password or API style credentials
-  rather than a grid style proxy based authentication. The pilot still requires
-  a suitable proxy, but this cannot be renewed via the cloud interface due to
-  the I/O limitations.
+  rather than a grid style proxy based authentication.
 - Pilot (VM) Tidy-up: Cloud providers will not remove stopped instances by
   default.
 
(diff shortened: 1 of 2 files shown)


src/DIRAC/Resources/Computing/CloudComputingElement.py
```diff
@@ -375,37 +360,6 @@ def _getMetadata(self, executableFile, pilotStamp=""):
...
-            return False
-        resdump = res["Value"].dumpAllToString()
-        if not resdump["OK"]:
-            self.log.error("Failed to dump proxy to string", resdump["Message"])
-            return False
-        self.proxy = resdump["Value"]
-        self.valid = datetime.datetime.utcnow() + proxyLifetime * datetime.timedelta(seconds=1)
-        return True
-
     def __init__(self, *args, **kwargs):
```
aldbr, line 363: I think you can remove `self.proxy = ""` in `__init__()`  because it is already defined similarly in `ComputingElement`, and thus, inherited to `CloudComputingElement`.

sfayer: Thanks, yes... I'm not sure why we're initialising those there in the first place.

aldbr (approved):

PR #7825, opened 2024-10-11
fix(WMS): JobWrapper checks existence of executable in jobIDPath

aldbr:
Should solve an issue observed during the latest hackathon.

BEGINRELEASENOTES
*WorkloadManagement
FIX: JobWrapper checks existence of executable in jobIDPath
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
+++ b/src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
@@ -307,7 +307,10 @@ class JobWrapper:
             executable = shutil.which(executable) or executable
 
         # Make the full path since . is not always in the PATH
-        executable = os.path.abspath(executable)
+        # self.jobIDPath is an absolute path, so we can use it directly
+        # if executable is an absolute path, self.jobIDPath is ignored
+        # Example: "/bin/ls" will be used as is
+        executable = str(self.jobIDPath / executable)
         if not os.path.exists(executable):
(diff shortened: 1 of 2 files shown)


src/DIRAC/WorkloadManagementSystem/JobWrapper/JobWrapper.py
```diff
@@ -307,7 +307,7 @@ def __prepareCommand(self):
             executable = shutil.which(executable) or executable
 
         # Make the full path since . is not always in the PATH
-        executable = os.path.abspath(executable)
+        executable = str(self.jobIDPath / executable)
```
andresailer, line 313: But if executable is `/bin/ls` this is going to fail, or is that treated in some other branch of the code?

aldbr: This is clearly not straightforward, so I am going to add a comment  to emphasize that behavior but:

```python
In [1]: from pathlib import Path

In [2]: Path.cwd() / "/usr/bin/ls"
Out[2]: PosixPath('/usr/bin/ls')
```

(there are already some tests handling this case, e.g. `test_preProcess_no_arguments`) 

Explained here in the documentation: https://docs.python.org/3/library/pathlib.html#pure-paths

> If a segment is an absolute path, all previous segments are ignored

andresailer: Ahh, OK, I see the Path object here works the same as os.path.join.
Thanks for adding the comment!

chrisburr (approved):

Issue #7851, opened 2024-10-23
Components stuck due to failure to send logs


chrisburr:
In LHCb we're seeing many services stuck. If I attach py-spy the spit out a bunch of errors and then start working again:

```python
2024-10-23 07:35:36 UTC WorkloadManagement/JobMonitoring NOTICE: Executing action ([2001:1458:d03:3::100:319]:37772)[lhcb_user:menghao] RPC/getJobsStates(<masked>)
2024-10-23 07:35:36 UTC WorkloadManagement/JobMonitoring NOTICE: Returning response ([2001:1458:d03:3::100:319]:37772)[lhcb_user:menghao] (0.04 secs) OK
2024-10-23 07:35:40 UTC WorkloadManagement/JobMonitoring NOTICE: Executing action ([2001:1458:d00:2f::100:1fb]:35594)[hosts:lbvobox301.cern.ch] RPC/getJobsStatus(<masked>)
2024-10-23 07:35:40 UTC WorkloadManagement/JobMonitoring NOTICE: Returning response ([2001:1458:d00:2f::100:1fb]:35594)[hosts:lbvobox301.cern.ch] (0.01 secs) OK
2024-10-23 11:00:33 UTC WorkloadManagement/JobMonitoring WARN: Sleeping as service requested throttling WorkloadManagement/JobMonitoring
2024-10-23 11:00:33 UTC WorkloadManagement/JobMonitoring ERROR: Error processing proposal Error in _write: (110, 'Connection timed out') SSLError(110, 'Connection timed out')
2024-10-23 11:00:33 UTC WorkloadManagement/JobMonitoring ERROR: Exception in handleConnection thread
Traceback (most recent call last):
  File "/opt/dirac/versions/v11.0.49-1729166531/Linux-x86_64/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/dirac/versions/v11.0.49-1729166531/Linux-x86_64/lib/python3.11/site-packages/DIRAC/Core/DISET/private/Service.py", line 351, in _processInThread
    if result["closeTransport"] or not result["OK"]:
       ~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'closeTransport'
[...]

PR #7854, opened 2024-10-23
[v8.0] fix (StompMQConnector): add a timeout for the StompConnector

ryuwd:
cc @chrisburr 

for #7851 

BEGINRELEASENOTES

*StompMQConnector
FIX: add a timeout for the StompConnector to handle nonresponsive sockets which can cause logging to be blocked

ENDRELEASENOTES


src/DIRAC/Resources/MessageQueue/StompMQConnector.py
```diff
@@ -83,6 +86,7 @@ def setupConnection(self, parameters=None):
         connectionArgs = {
             "vhost": vhost,
             "keepalive": True,
+            "timeout": stompTimeout,
```
chrisburr, line 95: Having tried it we also need something like `"heartbeats": (outgoingHeartbeatMs, incomingHeartbeatMs),` here as well. `15_000` seems like a sensible default

chrisburr:
I think the logic also needs fixing to close the connection if it's still open.

ryuwd:
> I think the logic also needs fixing to close the connection if it's still open.

It doesn't close by itself on timeout?

chrisburr:
> It doesn't close by itself on timeout?

I wouldn't expect it to.

chrisburr (approved):

chrisburr:
After more consideration I don't think it's needed.

PR #7867, opened 2024-11-01
[integration] fix: modify echo in function installDIRACX() where error occurs

martynia:
Function` installDIRACX()` in `utilities.sh` had a malformed `echo` statement which referred to undefined variables.

*Tests
 FIX: print a correct package name and a source directory.

The error would normally not pop up in CI since the GitHub workflow created necessary wheels:
```
pip install build
for pkg_dir in $PWD/diracx-* .; do
echo "Building $pkg_dir"
python -m build --outdir "${GITHUB_WORKSPACE}/dist" $pkg_dir
```

And when trying to run tests locally I naively issued a command:

`./integration_tests.py prepare-environment "TEST_DIRACX=Yes"  --diracx-dist-dir  <my_diracX_source_tree>"`

without building the wheels first.


aldbr:
Basic tests are failing because of a trailing space in `release-notes`. https://github.com/DIRACGrid/DIRAC/pull/7868 should fix it.

fstagni (approved):

PR #7886, opened 2024-11-12
[8.0] Use plain proxy for pilot bundle

sfayer:
Hi,

As discussed at the last BiLD meeting, I've been looking at the problem of the bundled proxy expiring in the middle of the pilot job, causing various failures (as this is no longer renewed like the proxy delegated via the CE).

Here is an implementation of my proposed fix: Instead of using the proxy that's used to submit the job to the CE (which is capped by the VO's maximum VOMS lifetime), we generate a new vanilla proxy (with the lifetime set to a multiple of the expected maximum runtime) to send in the bundle. The pilot then uses this new proxy to report back to DIRAC which doesn't need VOMS. This is the same approach we were previously using in the CloudCE (where pilot runtime could easily be 2+ weeks for efficiency reasons).

I've done basic testing and it seems to fix the problems I was initially seeing, seemingly without creating any new issues.

Regards,
Simon

BEGINRELEASENOTES
*WorkloadManagement
CHANGE: Use plain proxy for the pilot bundle
ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
+++ b/src/DIRAC/WorkloadManagementSystem/Agent/SiteDirector.py
@@ -109,6 +109,9 @@ class SiteDirector(AgentModule):
         self.maxQueueLength = 86400 * 3
         # Maximum number of times the Site Director is going to try to get a pilot output before stopping
         self.maxRetryGetPilotOutput = 3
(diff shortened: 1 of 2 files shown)


sfayer:
I think that's the sitedirector test fixed for this patch now... The one remaining failure appears to be a random one.

aldbr:
Yes I reran the test, it looks good!

aldbr (approved):

fstagni (approved):

PR #7936, opened 2024-12-05
[8.0] fix(Resources): AREX should renew the delegation before submitting a payload

aldbr:
As discussed with @marianne013 a few weeks ago

Otherwise, 
- if there is no pilot submission, there is no pilot monitoring, and
- if there is no pilot monitoring, there is no delegation renewal, and
- if there is no delegation renewal, then the delegation can expire (which happens in the certification instance)

I am not even sure it solves https://github.com/DIRACGrid/DIRAC/issues/7931 though

BEGINRELEASENOTES
*WorkloadManagement
FIX: renew delegation prior to submitting pilots
ENDRELEASENOTES


fstagni (approved):
I do not see anything wrong

PR #8053, opened 2025-02-20
[8.0] fix doc readme

andresailer:
BEGINRELEASENOTES

* docs
FIX: update and correct the README to build the DIRAC documenation

ENDRELEASENOTES

Addresses: https://github.com/DIRACGrid/DIRAC/discussions/8051

### DIFF
--- a/docs/README
+++ b/docs/README
@@ -3,16 +3,23 @@ How to build DIRAC documentation
 
 1. Create DIRAC client environment by an appropriate
    source bashrc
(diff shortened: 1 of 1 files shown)


martynia (approved):
I verified that the instructions are now correct. I installed a locally cloned version and I needed sphinx extensions in this case, so "might need" means "need" here.

fstagni (approved):

PR #8070, opened 2025-03-04
[8.0] feat (Storage): disable bearer plugins unless copying files

chaen:
Closes https://github.com/DIRACGrid/DIRAC/issues/8052
Thanks @cserf  ! 

Test script:
```python
from DIRAC import initialize; initialize()
from DIRAC.Resources.Storage.StorageElement import StorageElement
se = StorageElement('MyStorageWithHighLatency')
for i in range(100):
    se.removeFile(f"/lhcb/user/c/chaen/IDoNotExist_{i}.txt")
```

Before:

```bash
$ time python /tmp/remove.py

real    2m45.621s
user    0m8.523s
sys     0m0.389s
```

After

```bash
$ time python /tmp/remove.py

real    0m49.405s
user    0m6.151s
sys     0m0.285s
```
BEGINRELEASENOTES

*Resources
CHANGE: Disable Bearer token for HTTPs unless upload/TPC

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
+++ b/src/DIRAC/Resources/Storage/GFAL2_StorageBase.py
@@ -164,6 +164,10 @@ class GFAL2_StorageBase(StorageBase):
             os.environ.get("DIRAC_GFAL_GRIDFTP_ENABLE_IPV6", "true").lower() not in ["false", "no"],
         )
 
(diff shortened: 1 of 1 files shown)


fstagni (approved):
From @cserf : Maybe I forgot to push the button to finish the review, but cannot check right now. I didn't have any comment. The PR is good to go.

PR #8100, opened 2025-03-26
[9.0] Move JobWrapperUtilities after Script.parseCommandLine()

chrisburr:
I've seen jobs getting stuck in Matched at some sites. I'm 90% sure this is caused by `JobWrapperUtilities` being imported in `JobWrapperTemplate` before the config is initialised. This makes the CS refresh latency dependent as it's a race for if the CS is loaded before it is used.

BEGINRELEASENOTES

*WorkloadManagement
FIX: Move JobWrapperUtilities after import Script.parseCommandLine()

ENDRELEASENOTES


fstagni (approved):
I can't be fully sure you are right but this was anyway a bug.

PR #8117, opened 2025-04-04
[9.0] Follow DiracX client changes

chrisburr:
Required for https://github.com/DIRACGrid/diracx/pull/441

### DIFF
--- a/src/DIRAC/Core/Security/DiracX.py
+++ b/src/DIRAC/Core/Security/DiracX.py
@@ -14,12 +14,17 @@ import importlib
 import json
 import re
 import textwrap
+from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


src/DIRAC/Core/Security/DiracX.py
```diff
@@ -82,7 +87,7 @@ class FutureClient:
 
 
 @contextmanager
-def DiracXClient() -> _DiracClient:
+def DiracXClient() -> Iterator[SyncDiracClient]:
```
chaen, line 90: Why `Iterator`?

chrisburr: As this function `yield`s an instance of `SyncDiracClient`. The previous type hint is wrong. The `contextmanager` decorator then transforms `Generator[T, None, None]` -> `ContextManager[T]`.

(`Iterator[T]` is an alias for `Generator[T, None, None]`)

chaen (approved):

PR #8121, opened 2025-04-07
[9.0] Drop the direct use of the JobStateUpdateClient from servers

### LINKED ISSUE #8122: `JobStateUpdateClient().setJobParameter` no longer works with a host certificate
To draw attention to this, I think `JobStateUpdateClient().setJobParameter` no longer works with a host certificate as the VO isn't knonw.

cc @fstagni

_Originally posted by @chrisburr in https://github.com/DIRACGrid/DIRAC/pull/8121#discussion_r2030756564_

chrisburr:
BEGINRELEASENOTES

*RequestManagement
CHANGE: No longer directly use of the JobStateUpdateClient when processing requests
FIX: Calling setJobParameter when processing requests

*Transformation
CHANGE: No longer directly use of the JobStateUpdateClient from the DataRecoveryAgent

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/RequestManagementSystem/Client/ReqClient.py
+++ b/src/DIRAC/RequestManagementSystem/Client/ReqClient.py
@@ -11,20 +11,24 @@ import time
 import random
 import json
 import datetime
+from functools import cached_property
 
 # # from DIRAC
 from DIRAC import gLogger, S_OK, S_ERROR
 from DIRAC.Core.Utilities.List import randomize, fromChar
 from DIRAC.Core.Utilities.JEncode import strToIntDict
 from DIRAC.Core.Utilities.DEncode import ignoreEncodeWarning
+from DIRAC.Core.Utilities.ObjectLoader import ObjectLoader
+from DIRAC.Core.Utilities.ReturnValues import returnValueOrRaise
 from DIRAC.ConfigurationSystem.Client import PathFinder
 from DIRAC.Core.Base.Client import Client, createClient
 from DIRAC.RequestManagementSystem.Client.Request import Request
 from DIRAC.RequestManagementSystem.private.RequestValidator import RequestValidator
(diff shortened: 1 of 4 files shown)


src/DIRAC/RequestManagementSystem/Client/ReqClient.py
```diff
@@ -309,8 +313,7 @@ def finalizeRequest(self, requestID, jobID, useCertificates=True):
         :param str requestID: request id
         :param int jobID: job id
         """
-
-        stateServer = JobStateUpdateClient(useCertificates=useCertificates)
+        stateServer = _JobDBInteraction(useCertificates)
```
chaen, line 316: you're going to run into the same issue with the JobMonitoringClient below.

chrisburr: I know but that is used in multiple places so I thought it was better to focus on one client at a time.

src/DIRAC/RequestManagementSystem/Client/ReqClient.py
```diff
@@ -688,3 +691,52 @@ def recoverableRequest(request):
...
+
+    def __init__(self, useCertificates: bool):
+        self._useCertificates = useCertificates
+
+    def setJobParameter(self, jobID: int, key: str, value: str):
+        if self._useCertificates:
+            vo = returnValueOrRaise(self._jobStatusUtility.jobDB.getJobAttribute(jobID, "VO"))
+            return self._elasticJobParametersDB.setJobParameter(int(jobID), key, value, vo=vo)
+        else:
+            return self._client.setJobParameter(jobID, key, value)
```
chrisburr, line 713: To draw attention to this, I think `JobStateUpdateClient().setJobParameter` no longer works with a host certificate as the VO isn't knonw.

cc @fstagni

PR #8136, opened 2025-04-08
fix (proxy): pilot proxy are shipped with a token

chaen:
BEGINRELEASENOTES

*Core
FIX: Proxy Pilots are sent with a token

ENDRELEASENOTES


src/DIRAC/WorkloadManagementSystem/Agent/JobAgent.py
```diff
@@ -479,7 +479,6 @@ def _setupProxy(self, owner, ownerGroup):
         proxyChain = ret["Value"]["chain"]
         if "groupProperties" not in ret["Value"]:
             print(ret["Value"])
```
aldbr:
```suggestion
```

aldbr (dismissed):

fstagni (approved):

PR #8155, opened 2025-04-15
[9.0] Add caches to asn1_utils for better performance

chrisburr:
BEGINRELEASENOTES

*Core
NEW: Add caches to asn1_utils for better performance

ENDRELEASENOTES

### DIFF
--- a/src/DIRAC/Core/Security/m2crypto/asn1_utils.py
+++ b/src/DIRAC/Core/Security/m2crypto/asn1_utils.py
@@ -14,6 +14,8 @@ This is now pure python, but it might be interesting to wrap the existing
 C library (https://github.com/italiangrid/voms) instead...
 
 """
+from functools import lru_cache
+
 from pyasn1.codec.der.decoder import decode as der_decode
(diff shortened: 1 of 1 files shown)


chrisburr:
I added the following to get some instrumentation on the cache:

```python
import threading
import time

def cache_stats_monitor():
    while True:
        print("Cache stats for _extensionsFromCertDER:", _extensionsFromCertDER.cache_info())
        print("Cache stats for _decodeDIRACGroup  :", _decodeDIRACGroup.cache_info())
        # Wait for 5 seconds before printing the stats again
        time.sleep(5)

# Create and start a daemon thread to run the cache statistics monitor
monitor_thread = threading.Thread(target=cache_stats_monitor, daemon=True)
monitor_thread.start()
```

```
Cache stats for _extensionsFromCertDER: CacheInfo(hits=27212, misses=55, maxsize=1024, currsize=55)
Cache stats for _decodeDIRACGroup  : CacheInfo(hits=9106, misses=6, maxsize=128, currsize=6)
```

fstagni (approved):

PR #8180, opened 2025-04-25
[9.0] fix: disable watchdog wallclock check for remote executions

aldbr:
For some reason, payloads executed through the `PushJobAgent` are quickly killed by the `Watchdog`s because the CPU consumption is way below the wallclock time, which is expected as the real work is done on a remote worker node.

I don't really understand why we didn't have such issues in `v8.0` but the following patch should solve the problem, I am going to test it.
 
BEGINRELEASENOTES
*WorkloadManagement
FIX: disable watchdgo wallclock check for remote executions
ENDRELEASENOTES


aldbr:
It seems to work, I can see the following logs in the hotfixed `PushJobAgent`:

```
N.B. job would be declared as stalled but CPU / WallClock check is disabled by payload
```

fstagni (approved):

PR #8208, opened 2025-05-28
[8.0] fix: AREXCE not detecting pilots in an accepted/staged states

aldbr:
If the CE does not manage to pick up pilots from the session directory, then they are not detected as waiting by `AREX.getCEStatus`, and the Site Director just tries to submit pilots endlessly thinking there is no waiting pilot.

BEGINRELEASENOTES
*WorkloadManagement
FIX: AREXCE not detecting pilots in an accepted/staged states
ENDRELEASENOTES


aldbr:
CI should pass once https://github.com/DIRACGrid/DIRAC/pull/8210 is merged

fstagni (approved):

PR #8247, opened 2025-07-09
feat(resources): spool SSH+Condor job inputs/outputs

aldbr:
- would reduce the space taken by the logs in the user space
- we need to clean the user space from time to time using the `SSHCE` anyway, but if the submission pace is too fast, it's better to only get the logs of interests and not all of them.

BEGINRELEASENOTES
*Resources
CHANGE: add spool option to SSHCE+Condor
ENDRELEASENOTES


aldbr:
Tested in LHCb, it works fine!

fstagni (approved):

PR #8260, opened 2025-07-15
[9.0] Update job monitoring to use DiracX

Robin-Van-de-Merghel:
BEGINRELEASENOTES

*JobMonitoring
NEW: Add compatibility with DiracX and the JobMonitoring

ENDRELEASENOTES


Robin-Van-de-Merghel:
Based on #8248

Robin-Van-de-Merghel:
Based on #8248

PR #8268, opened 2025-07-17
[9.0] [POC] Pilot legacy logging

Robin-Van-de-Merghel:
BEGINRELEASENOTES

*PilotManagerHandler
NEW: POC base on #8233  to manage pilots, and allow legacy pilots to send logs.

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -62,7 +62,7 @@ jobs:
     - name: Building wheels
       run: |
         # Clone diracx
-        git clone --single-branch --branch robin-pilot-management https://github.com/Robin-Van-de-Merghel/diracx.git $GITHUB_WORKSPACE/diracx
+        git clone --single-branch --branch robin-pilot-legacy-logging https://github.com/Robin-Van-de-Merghel/diracx.git $GITHUB_WORKSPACE/diracx
 
         # Create dist dir
         mkdir -p $GITHUB_WORKSPACE/diracx-dist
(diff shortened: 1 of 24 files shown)


Robin-Van-de-Merghel:
# Context

> DIRAC pilot: Pilot sent by DIRAC with a proxy without a DiracX token embedded in it

> Legacy pilot: Pilot sent by DIRAC with a proxy with a DiracX token embedded in it

> DiracX pilot: Pilot sent by DiracX with a secret to exchange for a DiracX token

# Feature

Connected to:

1. https://github.com/DIRACGrid/diracx/pull/596 for DiracX backend supporting pilot management, and legacy logging
2. https://github.com/DIRACGrid/Pilot/pull/260 to contact DiracX directly to send logs from a proxy (+ later with a secret)

We can have a legacy pilot:

1. Extracting a DiracX token from its proxy  (showed by every integration test)
2. Sending logs to DiracX
3. Refreshing its tokens because it's an user token with a short lifespan

Robin-Van-de-Merghel:
> [!IMPORTANT]
> This PR is not usable in prod as of now. 

Pilots **DO NOT** have tokens in their proxies, so we cannot extract their token from it. https://github.com/DIRACGrid/Pilot/pull/260 is only working as in the CI, pilot proxies are generate via a `dirac-proxy-init` which embed a DiracX token. 

In the SiteDirector, there's no such thing.

PR #8270, opened 2025-07-18
[8.0] IdP clients per user group 

atsareg:
This PR adds the possibility to define IdP client credentials per user group to allow using the same IdProvider for several VOs without the need to define separate IdProvider per VO. If no setting is added for a group, generic client credentials will be taken. So, existing IdP configurations will be still valid. Example IdP configuration:

      Resources
      {
        IdProviders
        {
          <IdProvider name>
          {
            client_id = <client_id>
            client_secret = <client_secret>
            dirac_pilot
            {
              client_id = <client_id_for_dirac_VO>
              client_secret = <client_secret_for_dirac_VO>
            }
            biomed_pilot
            {
              client_id = <client_id_for_biomed_VO>
              client_secret = <client_secret_for_biomed_VO>
            }
          }
        }
      }

BEGINRELEASENOTES

*Framework
NEW: Possibility to define several OAuth2 clients for a given IdP for different DIRAC groups

ENDRELEASENOTES

### DIFF
--- a/docs/source/AdministratorGuide/HowTo/pilotsWithTokens.rst
+++ b/docs/source/AdministratorGuide/HowTo/pilotsWithTokens.rst
@@ -28,9 +28,24 @@ Setting up an ``IdProvider``
           {
             client_id = <client_id>
             client_secret = <client_secret>
+            dirac_pilot
+            {
+              client_id = <client_id_for_dirac_VO>
+              client_secret = <client_secret_for_dirac_VO>
+            }
+            biomed_pilot
+            {
+              client_id = <client_id_for_biomed_VO>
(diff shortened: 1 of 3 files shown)


aldbr (approved):
LGTM

chaen:
For my education, what is the gain of doing it like this instead of using one IdP per vo ? 
Also, given the examples you write, it is aimed at the pilot submission. But how does it work when you need tkens for other purposes which use a different group (VOMS2CS, FTS3, etc) ?

chaen:
After discussion with @atsareg we agreed that VO specific configuration shouldn't go as low in the CS, and rather stay at a higher level, just like done in `Operations`

